PR #69, opened 2022-10-07
Bugfix/ssn dups

albrja:
### Fixed bug that led to some simulants having the same social security numbers.
- *Category*: Bugfix
- *JIRA issue*: [MIC-3419](https://jira.ihme.washington.edu/browse/MIC-3419)
- *Research reference*: N/A

-update to calls to random_integers function in SSN Generator to include additional key to properly use randomness stream
-update to random_integers function to fix bug where simulants SSN group was 100.
-adds check when SSNs are generated to make sure no duplicates are created.

### Verification and Testing
Successfully ran simulation and saw expected update in SSNs for simulants.
initial_births
initial_ssns
six_month_births
six_month_ssns
ssn_group_fix

### DIFF
--- a/src/vivarium_census_prl_synth_pop/components/household.py
+++ b/src/vivarium_census_prl_synth_pop/components/household.py
@@ -155,7 +155,9 @@ class HouseholdMigration:
             self.proportion_households_leaving_country,
             "abroad_households",
         )
-        domestic_household_ids = ~all_household_ids_that_move.isin(abroad_household_ids)
+        domestic_household_ids = all_household_ids_that_move.loc[
+            ~all_household_ids_that_move.isin(abroad_household_ids)
+        ]
 
         # Get index of all simulants in households moving abroad and domestic
         abroad_households_idx = households.loc[
--- a/src/vivarium_census_prl_synth_pop/components/household.py
+++ b/src/vivarium_census_prl_synth_pop/components/household.py
@@ -138,34 +138,28 @@ class HouseholdMigration:
         move those households to a new address
         """
         households = self.population_view.get(event.index)
-        household_heads = households.loc[
+        household_ids = households.loc[
             households["relation_to_household_head"] == "Reference person"
         ]["household_id"]
 
-        households_that_move_idx = self.randomness.filter_for_rate(
-            household_heads.index,
-            self.household_move_rate(household_heads.index),
+        all_household_ids_that_move = self.randomness.filter_for_rate(
+            household_ids,
+            self.household_move_rate(household_ids.index),
             "all_moving_households",
         )
 
         # Determine which households move abroad
-        households_heads_that_move_abroad_idx = filter_by_rate(
-            households_that_move_idx,
+        abroad_household_ids = filter_by_rate(
+            all_household_ids_that_move,
             self.randomness,
             self.proportion_households_leaving_country,
             "abroad_households",
         )
-        # Find households_ids that move abroad and domestic
-        abroad_households_ids = household_heads.loc[households_heads_that_move_abroad_idx]
-        domestic_household_ids = household_heads.loc[
-            household_heads.index.isin(
-                households_that_move_idx.difference(households_heads_that_move_abroad_idx)
-            )
-        ]
+        domestic_household_ids = ~all_household_ids_that_move.isin(abroad_household_ids)
 
         # Get index of all simulants in households moving abroad and domestic
         abroad_households_idx = households.loc[
-            households["household_id"].isin(abroad_households_ids)
+            households["household_id"].isin(abroad_household_ids)
         ].index
         domestic_households_idx = households.loc[
             households["household_id"].isin(domestic_household_ids)
--- a/src/vivarium_census_prl_synth_pop/components/household.py
+++ b/src/vivarium_census_prl_synth_pop/components/household.py
@@ -139,21 +139,28 @@ class HouseholdMigration:
         """
         households = self.population_view.get(event.index)
         household_heads = households.loc[
-            households["relation_to_household_head"] == "Reference person"]['household_id']
+            households["relation_to_household_head"] == "Reference person"
+        ]["household_id"]
 
         households_that_move_idx = self.randomness.filter_for_rate(
-            household_heads.index, self.household_move_rate(household_heads.index)
+            household_heads.index,
+            self.household_move_rate(household_heads.index),
+            "all_moving_households",
         )
 
         # Determine which households move abroad
         households_heads_that_move_abroad_idx = filter_by_rate(
-            households_that_move_idx, self.randomness, self.proportion_households_leaving_country, "abroad_households"
+            households_that_move_idx,
+            self.randomness,
+            self.proportion_households_leaving_country,
+            "abroad_households",
         )
         # Find households_ids that move abroad and domestic
         abroad_households_ids = household_heads.loc[households_heads_that_move_abroad_idx]
         domestic_household_ids = household_heads.loc[
-            (household_heads.index.isin(households_that_move_idx)) &
-            (~household_heads.index.isin(households_heads_that_move_abroad_idx))
+            household_heads.index.isin(
+                households_that_move_idx.difference(households_heads_that_move_abroad_idx)
+            )
         ]
 
         # Get index of all simulants in households moving abroad and domestic
@@ -180,7 +187,7 @@ class HouseholdMigration:
             households = update_address_and_zipcode(
                 df=households,
                 rows_to_update=domestic_households_idx,
-                id_key=households['household_id'],
+                id_key=households["household_id"],
                 address_map=address_map,
                 zipcode_map=zipcode_map,
             )
--- a/src/vivarium_census_prl_synth_pop/components/person.py
+++ b/src/vivarium_census_prl_synth_pop/components/person.py
@@ -94,21 +94,24 @@ class PersonMigration:
         persons = self.population_view.get(event.index)
         non_household_heads = persons.loc[
             persons.relation_to_household_head != "Reference person"
-            ]
+        ]
 
         # Get subsets of possible simulants that can move
         persons_who_move_idx = self.randomness.filter_for_rate(
-            non_household_heads.index, self.person_move_rate(non_household_heads.index), "all_movers"
+            non_household_heads.index,
+            self.person_move_rate(non_household_heads.index),
+            "all_movers",
         )
         # Find simulants that move out of the country and those that move domestically
         abroad_movers_idx = filter_by_rate(
-            persons_who_move_idx, self.randomness, self.proportion_simulants_leaving_country, "abroad_movers"
+            persons_who_move_idx,
+            self.randomness,
+            self.proportion_simulants_leaving_country,
+            "abroad_movers",
         )
         domestic_movers_idx = persons.loc[
-            (persons.index.isin(persons_who_move_idx)) &
-            (~persons.index.isin(abroad_movers_idx))
-        ].index
-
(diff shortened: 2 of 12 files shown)


src/vivarium_census_prl_synth_pop/components/household.py
```diff
@@ -138,27 +138,30 @@ def on_time_step(self, event: Event):
...
         """
         households = self.population_view.get(event.index)
-        household_heads = households.loc[
-            households["relation_to_household_head"] == "Reference person"]['household_id']
-
-        households_that_move_idx = self.randomness.filter_for_rate(
-            household_heads.index, self.household_move_rate(household_heads.index)
+        household_ids = households.loc[
+            households["relation_to_household_head"] == "Reference person"
+        ]["household_id"]
```
mattkappel, line 143: You may want to consider shortening the string constants you're passing around. Doing string comparisons on long strings is needless expense, never mind the additional memory it'll take in the population table.

Even better, you may want to put strings like "Reference person" into a constant in data_values.py, so you redefine to a different string without fuss.

src/vivarium_census_prl_synth_pop/constants/results.py
```diff
@@ -95,5 +95,7 @@ def RESULTS_MAP(kind):
     df["key"] = columns
     df[
         "measure"
-    ] = kind  # per researcher feedback, this column is useful, even when it's identical for all rows
+    ] = (
+        kind
+    )  # per researcher feedback, this column is useful, even when it's identical for all rows
```
mattkappel, line 100: Yuck. This formatting change (black?) hits code in the template and is ugly. Nothing for you to change here, you shouldn't need to change this block.

src/vivarium_census_prl_synth_pop/utilities.py
```diff
@@ -232,32 +232,42 @@ def random_integers(
...
+    rate_producer: rate_producer for move rates
     randomness: RandomnessStream for component this is being run in
     additional_key: descriptive key to make sure randomness stream produces unique results
 
     Returns
     -------
     a pd.Index, subset from simulants, with those selected to be filtered.
     """
-    simulants = simulants.drop_duplicates()
+    sims_to_filter = sims_to_filter.drop_duplicates()
```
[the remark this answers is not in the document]
beatrixkh: sometimes it's not a list of sims, but "household_ids" (or "business_ids") that get passed to this function. in those cases, because there are multiple people per household or business, there can be duplicates.

i think it would make sense to change the signature of this function @albrja so that that is clearer, but maybe not in this PR

src/vivarium_census_prl_synth_pop/components/population.py
```diff
@@ -274,12 +278,16 @@ def initialize_newborns(self, pop_data: SimulantData) -> None:
             additional_key="sex_of_child",
         )
         new_births["alive"] = "alive"
+        new_births["entrance_time"] = pop_data.creation_time
+        new_births["exit_time"] = pd.NaT
+
+        # Generate SSNs for newborns and prevent dups
         new_births["ssn"] = self.ssn_generator.generate(new_births).ssn
+        while len(new_births.ssn) != len(new_births.ssn.unique()) or sum(new_births.ssn.isin(ssns)) > 0:
+            new_births["ssn"] = self.ssn_generator.generate(new_births).ssn
```
beatrixkh, line 287: why does this need to happen for this line, but not for the original ssns created on population initialization?

src/vivarium_census_prl_synth_pop/components/population.py
```diff
@@ -124,17 +123,20 @@ def generate_initial_population(self, pop_data: SimulantData) -> None:
...
         )
-
         # format
-        n_chosen = pop.shape[0]
+        n_chosen = pop.shape[0]  # what is this?
+
+       # Add Social Security Numbers
         pop["ssn"] = self.ssn_generator.generate(pop).ssn
+        # Check for duplicate ssn
+        while len(pop.ssn) != len(pop.ssn.unique()):
```
beatrixkh, line 132: if we have a really huge population, what are the chances that this starts to have to loop many many times?

^ if that's a possibility, we could switch to looping over regenerating just the duplicates. im not sure if that ends up saving a lot of time or no, but possibly(?) if we have 350 million sims

we do this a couple places in the code. for example in `persons.py`:

```
    def _get_new_household_ids(
        self, persons_who_move: pd.DataFrame, event: Event
    ) -> pd.Series:
        households = self.population_view.subview(["household_id"]).get(event.index)
        all_household_ids = list(households.squeeze().drop_duplicates())

        new_household_ids = persons_who_move["household_id"].copy()
        additional_seed = 0
        while (new_household_ids == persons_who_move.household_id).any():
            unchanged_households = new_household_ids == persons_who_move.household_id
            new_household_ids[unchanged_households] = self.randomness.choice(
                new_household_ids.loc[unchanged_households].index,
                all_household_ids,
                additional_key=additional_seed,
            )
            additional_seed += 1

        return pd.Series(new_household_ids)
```

src/vivarium_census_prl_synth_pop/components/synthetic_pii.py
```diff
@@ -54,19 +54,19 @@ def generate(self, df_in: pd.DataFrame) -> pd.DataFrame:
...
-        )
+            min_val=1, max_val=99, index=df.index, randomness=self.randomness, additional_key="ssn_group"
+        )  # this is making some sims have a group of 100 - wrong
         df["ssn_group"] = group
 
         serial = random_integers(
-            min_val=1, max_val=9999, index=df.index, randomness=self.randomness
-        )
+            min_val=1, max_val=9999, index=df.index, randomness=self.randomness, additional_key="ssn_serial"
+        )  # this does not make serial == 10,000 but could (sample size?)
```
beatrixkh, line 69: what does this comment mean?

src/vivarium_census_prl_synth_pop/components/population.py
beatrixkh, line 287: same comment as on line 132.

also, might need to check that none of the new babies are duplicate with existing adult's ssns?

beatrixkh (changes requested):
I think it's currently possible for newborns to have sims that are duplicate with existing adults in the sim (unless im misreading). + a few other comments

src/vivarium_census_prl_synth_pop/utilities.py
```diff
@@ -232,32 +232,42 @@ def random_integers(
...
+    additional_key:
+        A identifying key for the randomness stream (usually a string)
 
     Returns
     -------
     pandas.Series
         An indexed set of integers in the interval [a,b)
     """
-    return (randomness.get_draw(index=index) * max_val + min_val).round().astype(int)
+    return np.floor(randomness.get_draw(index=index, additional_key=additional_key) * max_val + min_val).astype(int)
```
rmudambi, line 243: This still seems wrong. If you call `random_integers(min_val=5, max_val=10)`, you could draw a random number 0.99 and get `np.floor(0.99 * 10 + 5) == np.floor(14.9) == 14`, which is greater than the supposed max value of 10.

src/vivarium_census_prl_synth_pop/utilities.py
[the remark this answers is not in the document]
rmudambi: I think it is in scope to fix the docstring in this PR, since you're updating the functionality.

src/vivarium_census_prl_synth_pop/components/population.py
```diff
@@ -124,17 +123,20 @@ def generate_initial_population(self, pop_data: SimulantData) -> None:
         pop["date_of_birth"] = self.start_time - pd.to_timedelta(
             np.round(pop["age"] * 365.25), unit="days"
         )
-
         # format
-        n_chosen = pop.shape[0]
+        n_chosen = pop.shape[0]  # what is this?
```
rmudambi, line 127: Is this comment intended to be here?

src/vivarium_census_prl_synth_pop/components/person.py
```diff
@@ -174,20 +184,23 @@ def on_time_step(self, event: Event) -> None:
...
+        households = pop["household_id"]
+        all_household_ids = list(
+            households.squeeze().drop_duplicates()
+        )  # all household_ids in simulation
 
-        new_household_ids = persons_who_move["household_id"].copy()  # People who move household_ids
+        new_household_ids = pop.loc[
+            sims_who_move, "household_id"
+        ].copy()  # People who move household_ids
+        old_household_ids = new_household_ids.copy()
```
rmudambi, line 196: No need for this copy - you can just compare to `pop.loc[sims_who_move, "household_id"`

rmudambi (changes requested):

src/vivarium_census_prl_synth_pop/components/population.py
albrja: That's what the right side of the or statement is checking but I agree with the other part.

src/vivarium_census_prl_synth_pop/components/synthetic_pii.py
```diff
@@ -31,7 +31,7 @@ class SSNGenerator(GenericGenerator):
     def name(self):
         return "SSNGenerator"
 
-    def generate(self, df_in: pd.DataFrame) -> pd.DataFrame:
+    def generate(self, df_in: pd.DataFrame, additional_key: Any = None) -> pd.DataFrame:
         """Generate synthetic Social Security Numbers
 
         Parameters
```
beatrixkh, line 37: might be nice to add to the "results" here the guarantee that they are unique for the inputted df_in

beatrixkh (approved):

src/vivarium_census_prl_synth_pop/components/population.py
```diff
@@ -280,16 +280,14 @@ def initialize_newborns(self, pop_data: SimulantData) -> None:
         new_births["ssn"] = self.ssn_generator.generate(new_births).ssn
         # Check for SSN duplicates with existing SSNs
         ssn_values = ssns.loc[ssns != ""]
-        all_ssns = pd.concat([ssn_values, new_births.ssn])
-        duplicate_mask = all_ssns.duplicated()
+        duplicate_mask = new_births["ssn"].isin(ssn_values) | new_births.duplicated(subset=["ssn"])
```
rmudambi, line 283: `new_births` can't have duplicates due to the way the ssn generator works unless I misunderstood that method (in which case the initialization method is not correct).

src/vivarium_census_prl_synth_pop/components/population.py
```diff
@@ -280,16 +280,14 @@ def initialize_newborns(self, pop_data: SimulantData) -> None:
...
         additional_key = 1
         while sum(duplicate_mask) > 0:
             additional_key += 1
             new_births.loc[duplicate_mask, "ssn"] = self.ssn_generator.generate(
                 new_births.loc[duplicate_mask], additional_key).ssn
-            all_ssns = pd.concat([ssn_values, new_births.ssn])
-            duplicate_mask = all_ssns.duplicated()
-            # This checks for duplicates against existing SSNs that are not missing SSNs
+            duplicate_mask = new_births["ssn"].isin(ssn_values) | new_births.duplicated(subset=["ssn"])
```
rmudambi, line 289: Instead of doing this, why not add the new ssns to `ssn_values`?

rmudambi:
```
duplicated_ssns = new_births.duplicated(subset=["ssn"])
ssns = pd.concat([ssns, new_births.loc[duplicated_mask & ~duplicated_ssns]])
duplicate_mask = new_births["ssn"].isin(ssns)
```

rmudambi (approved):

PR #70, opened 2022-10-07
Update/pipeline removal

albrja:
### Removed unnecessary pipelines and changed them to lookup tables to correctly use desired rates.
- *Category*: Other
- *JIRA issue*: [MIC-3487](https://jira.ihme.washington.edu/browse/MIC-3487)
- *Research reference*: N/A

-Changed pipelines to lookup table in places where pipelines were being used incorrectly.

### Verification and Testing
Successfully ran simulation and saw expected behavior after changing pipelines to lookup tables.
abroad_movers
domestic_household_movers
domestic_movers
missing_ssns

### DIFF
--- a/src/vivarium_census_prl_synth_pop/components/businesses.py
+++ b/src/vivarium_census_prl_synth_pop/components/businesses.py
@@ -71,15 +71,9 @@ class Businesses:
         self.population_view = builder.population.get_view(self.columns_used)
         self.businesses = None
 
-        job_change_rate_data = builder.lookup.build_table(data_values.YEARLY_JOB_CHANGE_RATE)
-        self.job_change_rate = builder.value.register_rate_producer(
-            f"{self.name}.job_change_rate", source=job_change_rate_data
-        )
+        self.job_change_rate = builder.lookup.build_table(data_values.YEARLY_JOB_CHANGE_RATE)
+        self.businesses_move_rate = builder.lookup.build_table(data_values.BUSINESS_MOVE_RATE_YEARLY)
 
-        move_rate_data = builder.lookup.build_table(data_values.BUSINESS_MOVE_RATE_YEARLY)
-        self.businesses_move_rate = builder.value.register_rate_producer(
-            f"{self.name}.move_rate", source=move_rate_data
-        )
         self.addresses = builder.components.get_component("Address")
         builder.population.initializes_simulants(
             self.on_initialize_simulants,
--- a/src/vivarium_census_prl_synth_pop/components/businesses.py
+++ b/src/vivarium_census_prl_synth_pop/components/businesses.py
@@ -71,8 +71,15 @@ class Businesses:
         self.population_view = builder.population.get_view(self.columns_used)
         self.businesses = None
 
-        self.job_change_rate = builder.lookup.build_table(data_values.YEARLY_JOB_CHANGE_RATE)
-        self.businesses_move_rate = builder.lookup.build_table(data_values.BUSINESS_MOVE_RATE_YEARLY)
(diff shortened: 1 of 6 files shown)


src/vivarium_census_prl_synth_pop/components/businesses.py
```diff
@@ -71,15 +71,9 @@ def setup(self, builder: Builder):
         self.population_view = builder.population.get_view(self.columns_used)
         self.businesses = None
 
-        job_change_rate_data = builder.lookup.build_table(data_values.YEARLY_JOB_CHANGE_RATE)
-        self.job_change_rate = builder.value.register_rate_producer(
-            f"{self.name}.job_change_rate", source=job_change_rate_data
-        )
+        self.job_change_rate = builder.lookup.build_table(data_values.YEARLY_JOB_CHANGE_RATE)
```
hussain-jafari, line 74: Was something breaking or being processed incorrectly when using pipelines instead of lookup tables?

beatrixkh: no; they accomplish the same thing. however, pipelines allow you to register pipeline modifiers -- which we don't have or expect to have, so pipelines were overkill for our use-case

src/vivarium_census_prl_synth_pop/utilities.py
```diff
@@ -246,14 +246,14 @@ def random_integers(
 def filter_by_rate(
     sims_to_filter: Union[pd.Index, pd.Series],
```
beatrixkh, line 247: per comment in other PR (this is not the correct place, but since you're touching the function in this PR anyway, maybe this is an okay place to tweak this without opening a whole new PR) instead of `sims_to_filter`, maybe something like `entities_to_filter`? given that it could be household_ids or business_ids (i think, right?)

beatrixkh (approved):
haha oh no we had so many of these. thank you for updating.

src/vivarium_census_prl_synth_pop/components/businesses.py
```diff
@@ -71,15 +71,9 @@ def setup(self, builder: Builder):
...
+        self.businesses_move_rate = builder.lookup.build_table(data_values.BUSINESS_MOVE_RATE_YEARLY)
 
-        move_rate_data = builder.lookup.build_table(data_values.BUSINESS_MOVE_RATE_YEARLY)
-        self.businesses_move_rate = builder.value.register_rate_producer(
```
rmudambi, line 80: When we talked about this last week we discussed how `register_rate_producer` is actually doing something important (namely scaling the rate for a time-step). We only want to change this in the locations where we are only using `register_value_producer`. 

Both the job change rate and business move rates should remain as they were.

src/vivarium_census_prl_synth_pop/components/person.py
```diff
@@ -70,13 +70,9 @@ def setup(self, builder: Builder):
         self.person_move_rate = builder.value.register_rate_producer(
             f"{self.name}.move_rate", source=move_rate_data
         )
-        proportion_simulants_leaving_country_data = builder.lookup.build_table(
+        self.proportion_simulants_leaving_country = builder.lookup.build_table(
             data=data_values.PROPORTION_PERSONS_LEAVING_COUNTRY
         )
-        self.proportion_simulants_leaving_country = builder.value.register_rate_producer(
```
rmudambi, line 76: This is a correct change, since it isn't a rate.

rmudambi (changes requested):

mattkappel (approved):

rmudambi (approved):

PR #71, opened 2022-10-08
Fixed mapping bug for businesses

albrja:
### Updated business mapper to fix bug that was using index instead of employer id.
- *Category*: Bugfix
- *JIRA issue*: <][MIC-3489](https://jira.ihme.washington.edu/browse/MIC-3489)
- *Research reference*: N/A

-Fixed call to business mapper to use series instead of index.
-Fixed call that was incorrectly getting subset of state table.

### Verification and Testing
Ran simulation and saw correct changes to business metadata columns in both state tables.
updated_state_table
updating_business_metadata

### DIFF
--- a/src/vivarium_census_prl_synth_pop/components/businesses.py
+++ b/src/vivarium_census_prl_synth_pop/components/businesses.py
@@ -152,7 +152,7 @@ class Businesses:
         if len(businesses_that_move_idx) > 0:
             # update the employer address and zipcode in self.businesses
             address_map, zipcode_map = self.addresses.get_new_addresses_and_zipcodes(
-                businesses_that_move_idx,
+                all_businesses.loc[businesses_that_move_idx],
                 state=metadata.US_STATE_ABBRV_MAP[self.location].lower(),
             )
             self.businesses = update_address_and_zipcode(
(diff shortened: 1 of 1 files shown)


src/vivarium_census_prl_synth_pop/components/businesses.py
```diff
@@ -152,7 +152,7 @@ def on_time_step(self, event: Event):
         if len(businesses_that_move_idx) > 0:
             # update the employer address and zipcode in self.businesses
             address_map, zipcode_map = self.addresses.get_new_addresses_and_zipcodes(
-                businesses_that_move_idx,
+                all_businesses.loc[businesses_that_move_idx],
```
beatrixkh, line 155: `get_new_addresses_and_zipcodes` says it requires:
`those_that_move: a pd.Series of ids (e.g. employer_id or household_id) who will move`

^ aren't you passing in a DataFrame there, with columns "age" and "household_id"?

so i think(?) it should instead be `all_businesses.loc[businesses_that_move_idx,'household_id']`?

beatrixkh (approved):

rmudambi (approved):

PR #75, opened 2022-10-26
Data/business names

albrja:
### Adds scripts that can randomly generate business names.
- *Category*: Data artifact
- *JIRA issue*: [MIC-3437](https://jira.ihme.washington.edu/browse/MIC-3437)
- *Research reference*: https://ihme.slack.com/archives/C02KUQ9LX32/p1663282448475969

-Adds additional filepath to save new business names to project
-Creates series of business names that will be added to data artifact at a later date for large scale runes.

### Verification and Testing
-Successfully ran script with smaller scale (10,000 names).

### DIFF
--- a/src/vivarium_census_prl_synth_pop/constants/data_keys.py
+++ b/src/vivarium_census_prl_synth_pop/constants/data_keys.py
@@ -28,6 +28,7 @@ class __SyntheticData(NamedTuple):
     FIRST_NAMES: str = "synthetic_data.first_names"
     LAST_NAMES: str = "synthetic_data.last_names"
     ADDRESSES: str = "synthetic_data.addresses"
+    BUSINESS_NAMES: str = "synthetic_data.business_names"
 
     @property
     def name(self):
--- a/src/vivarium_census_prl_synth_pop/constants/data_values.py
+++ b/src/vivarium_census_prl_synth_pop/constants/data_values.py
@@ -57,6 +57,7 @@ PROBABILITY_OF_SPACE_IN_NAME = {
 #####################
 
 # data from https://www.jec.senate.gov/public/index.cfm/republicans/fl/
+BUSINESS_NAMES_MAX_TOKENS_LENGTH = 15
 PROPORTION_WORKFORCE_EMPLOYED = {"Florida": 0.576}
 WORKING_AGE = 18
 EXPECTED_EMPLOYEES_PER_BUSINESS = 90.105203
--- a/src/vivarium_census_prl_synth_pop/constants/paths.py
+++ b/src/vivarium_census_prl_synth_pop/constants/paths.py
@@ -16,7 +16,6 @@ PERSONS_DATA_DIR = PROJECT_ROOT / "data/raw_data/current/United_States/"
 
 HOUSEHOLD_MOVE_RATE_PATH = BASE_DIR / "data/raw_data/move_rates_including_gq.csv"
 BUSINESS_NAMES_DATA = PROJECT_ROOT / "data/raw_data/business_names.csv.bz2"
-# todo: where do we want to save this: share or repo?
 BUSINESS_NAMES_DATA_ARTIFACT_INPUT_PATH = (
     PROJECT_ROOT / "data/raw_data/randomly_generated_business_names.csv.bz2"
 )
--- a/src/vivarium_census_prl_synth_pop/constants/paths.py
+++ b/src/vivarium_census_prl_synth_pop/constants/paths.py
@@ -17,7 +17,9 @@ PERSONS_DATA_DIR = PROJECT_ROOT / "data/raw_data/current/United_States/"
 HOUSEHOLD_MOVE_RATE_PATH = BASE_DIR / "data/raw_data/move_rates_including_gq.csv"
 BUSINESS_NAMES_DATA = BASE_DIR / "data/raw_data/business_names.csv.bz2"
 # todo: where do we want to save this: share or repo?
-BUSINESS_NAMES_DATA_ARTIFACT_INPUT_PATH = PROJECT_ROOT / "data/raw_data/randomly_generated_business_names.hdf"
+BUSINESS_NAMES_DATA_ARTIFACT_INPUT_PATH = (
+    PROJECT_ROOT / "data/raw_data/randomly_generated_business_names.csv.bz2"
+)
 
 HOUSEHOLDS_FILENAMES = [f"psam_hus{x}.csv" for x in ["a", "b", "c", "d"]]
 PERSONS_FILENAMES = [f"psam_pus{x}.csv" for x in ["a", "b", "c", "d"]]
--- a/src/vivarium_census_prl_synth_pop/constants/paths.py
+++ b/src/vivarium_census_prl_synth_pop/constants/paths.py
@@ -15,6 +15,7 @@ HOUSEHOLDS_DATA_DIR = PROJECT_ROOT / "data/raw_data/current/United_States/"
 PERSONS_DATA_DIR = PROJECT_ROOT / "data/raw_data/current/United_States/"
 
 HOUSEHOLD_MOVE_RATE_PATH = BASE_DIR / "data/raw_data/move_rates_including_gq.csv"
+BUSINESS_NAMES_DATA = BASE_DIR / "data/raw_data/business_names.csv.bz2"
 
 HOUSEHOLDS_FILENAMES = [f"psam_hus{x}.csv" for x in ["a", "b", "c", "d"]]
 PERSONS_FILENAMES = [f"psam_pus{x}.csv" for x in ["a", "b", "c", "d"]]
--- a/src/vivarium_census_prl_synth_pop/constants/paths.py
+++ b/src/vivarium_census_prl_synth_pop/constants/paths.py
@@ -16,6 +16,8 @@ PERSONS_DATA_DIR = PROJECT_ROOT / "data/raw_data/current/United_States/"
 
 HOUSEHOLD_MOVE_RATE_PATH = BASE_DIR / "data/raw_data/move_rates_including_gq.csv"
 BUSINESS_NAMES_DATA = BASE_DIR / "data/raw_data/business_names.csv.bz2"
+# todo: where do we want to save this: share or repo?
+BUSINESS_NAMES_DATA_ARTIFACT_INPUT_PATH = PROJECT_ROOT / "data/raw_data/randomly_generated_business_names.hdf"
 
 HOUSEHOLDS_FILENAMES = [f"psam_hus{x}.csv" for x in ["a", "b", "c", "d"]]
 PERSONS_FILENAMES = [f"psam_pus{x}.csv" for x in ["a", "b", "c", "d"]]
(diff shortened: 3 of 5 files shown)


src/vivarium_census_prl_synth_pop/constants/data_keys.py
```diff
@@ -28,6 +28,7 @@ class __SyntheticData(NamedTuple):
     FIRST_NAMES: str = "synthetic_data.first_names"
     LAST_NAMES: str = "synthetic_data.last_names"
     ADDRESSES: str = "synthetic_data.addresses"
+    BUSINESS_NAMES: str = "synthetic_data.business_names"
```
albrja, line 31: This will be used later when we want to add the names we generate to the artifact.

src/vivarium_census_prl_synth_pop/data/generate_business_names.py
```diff
@@ -0,0 +1,119 @@
...
+        if i == 1:
+            logger.info("Initializing random name sampling for business name generator")
+        previous_word = f"word_{i - 1}"
+        next_word = f"word_{i}"
+        current_words_count_dict = names[previous_word].value_counts().to_dict()
+        for word in current_words_count_dict.keys():
+            logger.info(
+                f"Generating {len(list(current_words_count_dict.keys()))} words "
+                f"in {i}th of {n_max_tokens} columns for business names data."
+            )
```
albrja, line 98: This logging message could be improved.  It currently prints the cycles through the column and how many words are being created by the previous tokens.  I think adding it here or below after tokens is sampled in line 102 both make the messages printed out quite messy.

src/vivarium_census_prl_synth_pop/constants/paths.py
```diff
@@ -15,6 +15,11 @@
 PERSONS_DATA_DIR = PROJECT_ROOT / "data/raw_data/current/United_States/"
 
 HOUSEHOLD_MOVE_RATE_PATH = BASE_DIR / "data/raw_data/move_rates_including_gq.csv"
+BUSINESS_NAMES_DATA = BASE_DIR / "data/raw_data/business_names.csv.bz2"
+# todo: where do we want to save this: share or repo?
+BUSINESS_NAMES_DATA_ARTIFACT_INPUT_PATH = (
```
stevebachmeier, line 20: Just be sure if you decide the repo you make it accessible with `pip install` (no -e) by adding it to the MANIFEST (or package_data if this repo doesn't have a MANIFEST...I'm not familiar with where exactly those files come from) https://stackoverflow.com/questions/1612733/including-non-python-files-with-setup-py

src/vivarium_census_prl_synth_pop/data/generate_business_names.py
```diff
@@ -0,0 +1,119 @@
...
+from collections import defaultdict
+
+import numpy as np
+import pandas as pd
+from loguru import logger
+
+from vivarium_census_prl_synth_pop.constants import data_values, paths
+
+
+def generate_business_names_data(n_total_names: str):
```
stevebachmeier, line 11: To clarify, you're adding this functionality here but it's not actually being used anywhere yet? Why are you adding it now then?

src/vivarium_census_prl_synth_pop/constants/paths.py
albrja: The will likely be useful here.  This is going to be a pretty long series (like at least 10 million names) so I have where creating it will save it but keeping it in the repo does not seem like a good idea.

src/vivarium_census_prl_synth_pop/data/generate_business_names.py
```diff
@@ -0,0 +1,119 @@
...
+        header=["business_names"],
+        index=False,
+        compression="bz2"
+    )
+
+
+def make_bigrams(df: pd.DataFrame):
+    # Makes default dict of business names for map to sample from
+    # bigrams will be a Dict[str: Dict[str: int]]
+    # Example {"<start>": {keys are all first words for businesses: values are frequence where these pairs happen}}
```
stevebachmeier, line 45: Can you provide a realistic example?

src/vivarium_census_prl_synth_pop/data/generate_business_names.py
albrja: Yes.  This is to generate a list of random names.  Its going to be a big list so we want to keep it separate from the artifact.  There is also a good chance since we do not need it during the actual simulation (we will identify employers by the employer_id column) we may add them later in some type of post-processing step.

src/vivarium_census_prl_synth_pop/constants/paths.py
mattkappel: Consider putting that in the artifact.

Why would you want a non-editable install of a research repo under development?

mattkappel (approved):

src/vivarium_census_prl_synth_pop/constants/data_values.py
```diff
@@ -57,6 +57,7 @@
 #####################
 
 # data from https://www.jec.senate.gov/public/index.cfm/republicans/fl/
+BUSINESS_NAMES_MAX_TOKENS_LENGTH = 15
```
beatrixkh, line 60: it would be worth double checking with abie that this number wasn't chosen arbitrarily. a smaller max token will let it run faster

aflaxman:
15 definitely was an arbitrary choice that I did not explore much.  But how much smaller would you need to see things go much faster?

albrja:
@aflaxman I think we would have to drop the number significantly.  Runtime is expected to be around 24 hours due to the large size of the list (10.75 million) and there will likely not be very many business that have 15 words in their name.

hussain-jafari (approved):

src/vivarium_census_prl_synth_pop/data/generate_business_names.py
```diff
@@ -0,0 +1,148 @@
...
+
+
+def generate_business_names_data(n_total_names: str):
+    # loads csv of business names and generates pandas series with random business names
+
+    business_names = pd.read_csv(paths.BUSINESS_NAMES_DATA)
+    bigrams = make_bigrams(business_names)
+
+    # Get frequency of business names and find uncommon ones
+    s_name_freq = business_names.location_name.value_counts()
```
beatrixkh, line 18: could you rename this to be a little clearer? i cant tell what `s` refers to

albrja: Changed.

src/vivarium_census_prl_synth_pop/data/generate_business_names.py
```diff
@@ -0,0 +1,148 @@
...
+
+    return bigrams
+
+
+def sample_names(bigrams: defaultdict, n_businesses: int, n_max_tokens: int) -> pd.Series:
+    """
+
+    Parameters
+    ----------
+    bigrams: Default dict produced from make_bigrams function (see formatting of default dict.
```
beatrixkh, line 103: could you update the naming or the comments so it's clearer what this `bigrams` contains?

src/vivarium_census_prl_synth_pop/data/generate_business_names.py
beatrixkh: yeah this might be more logging than we need. what is the purpose of the logging? are you wanting to track this as it runs?

src/vivarium_census_prl_synth_pop/data/generate_business_names.py
```diff
@@ -0,0 +1,148 @@
...
+        next_word = f"word_{i}"
+        current_words_count_dict = names[previous_word].value_counts().to_dict()
+        for word in current_words_count_dict.keys():
+            logger.info(
+                f"Generating {len(list(current_words_count_dict.keys()))} words "
+                f"in {i}th of {n_max_tokens} columns for business names data."
+            )
+            if word != "<end>":
+                vals = list(bigrams[word].keys())
+                pr = np.array(list(bigrams[word].values()))
```
beatrixkh, line 130: maybe name this something else? because right now `np.array(list(bigrams[word].values()))` is a frequency, and then it becomes a probability on line 132 when you do `pr / pr.sum()`

src/vivarium_census_prl_synth_pop/data/generate_business_names.py
```diff
@@ -0,0 +1,148 @@
...
+
+    Create bigrams to map a word and the frequence at which the next choice happens.
+
+        {'<start>': {'Cottage': 1,
+                     'Union': 1,
+                     'Larkfield': 1,
+                     'Randall': 1,
+                     'Sport': 1
+                     }
+        }
```
beatrixkh, line 68: can you include a second-level of this example? like {'Cottage':['Inn', 'St']} type of thing?

src/vivarium_census_prl_synth_pop/data/generate_business_names.py
```diff
@@ -0,0 +1,148 @@
...
+
+    def dict_factory():
+        return lambda: defaultdict(int)
+
+    bigrams = defaultdict(dict_factory())
+    n_rows = len(df)  # expect a few minutes
+
+    for i in range(n_rows):
+        if i % 10_000 == 0:
+            print(".", end=" ", flush=True)
```
beatrixkh, line 84: i think this print isn't necessary if we're not running in an interactive context -- right?

src/vivarium_census_prl_synth_pop/data/generate_business_names.py
```diff
@@ -0,0 +1,148 @@
...
+    while len(new_names) < n_total_names:
+        n_needed = n_total_names - len(new_names)
+        more_names = sample_names(
+            bigrams, n_needed, data_values.BUSINESS_NAMES_MAX_TOKENS_LENGTH
+        )
+        new_names = pd.concat([new_names, more_names]).drop_duplicates()
+        new_names = new_names.loc[~new_names.isin(real_but_uncommon_names)]
+
+    new_names.to_csv(
+        paths.BUSINESS_NAMES_DATA_ARTIFACT_INPUT_PATH,
```
beatrixkh, line 35: I think getting @rmudambi 's input here would be helpful -- depending on how much processing you'll need to do on this, i think it makes sense to put it in share, not the project repo?

beatrixkh (approved):
a few naming/documentation nits, but overall the structure looks good. as for where it gets saved --- share vs artifact vs the repo -- i think that depends on what else we expect to need to do to it, which im not currently clear on. do we know? if no further processing needs to happen to it, and names might only get assigned after the sim has run, i would think `/share` as opposed to the project repo or the artifact.

src/vivarium_census_prl_synth_pop/data/generate_business_names.py
albrja: Yes.  Basically tracking progress as it generates names just since it will be running for a long time.

src/vivarium_census_prl_synth_pop/data/generate_business_names.py
albrja: I've updated the docstring.  Im trying to keep the names for the different data objects throughout the module so it is (hopefully) easier to follow.

src/vivarium_census_prl_synth_pop/data/generate_business_names.py
albrja: It is just another way of tracking that the dictionary is still being created.

PR #80, opened 2022-11-17
Feature/guardian initialization

albrja:
### Initializes guardians for simulants under 18 not living in group quarters.
- *Category*: Feature
- *JIRA issue*: [MIC-3540](https://jira.ihme.washington.edu/browse/MIC-3540)
- *Research reference*:  https://vivarium-research.readthedocs.io/en/latest/concept_models/vivarium_census_synthdata/concept_model.html#initializing-guardian-s-for-all-simulants

-Creates household structure dataframe to make household lookups easier (see docstring)
-Creates guardian_1 and guardian_2 columns in the state table with value -1 (N/A). 
-If a simulant is determined to have a guardian(s) (see research documentation for diagram), the guardian columns will be filled with values of the index for that simulants guardians.

### Verification and Testing
-Successfully ran simulation
-Saw updated guardian_1 and guardian_2 columns for simulants who have guardians.
orange_box
red_box
yellow_box
blurple_box
under_18_general_pop

### DIFF
--- a/src/vivarium_census_prl_synth_pop/components/population.py
+++ b/src/vivarium_census_prl_synth_pop/components/population.py
@@ -1,6 +1,5 @@
 import numpy as np
 import pandas as pd
-
 from vivarium.framework.engine import Builder
 from vivarium.framework.event import Event
 from vivarium.framework.population import SimulantData
@@ -237,7 +236,7 @@ class Population:
         parent_ids_idx = pop_data.user_data["parent_ids"]
         pop_index = pop_data.user_data["current_population_index"]
         mothers = self.population_view.get(parent_ids_idx.unique())
-        ssns = self.population_view.subview(['ssn']).get(pop_index).squeeze()
+        ssns = self.population_view.subview(["ssn"]).get(pop_index).squeeze()
         new_births = pd.DataFrame(data={"parent_id": parent_ids_idx}, index=pop_data.index)
 
         inherited_traits = [
@@ -262,16 +261,24 @@ class Population:
         # birthday map between parent_ids and DOB (so twins get same bday)
         # note we use np.floor to guarantee birth at midnight
         dob_map = {
-            parent: dob for (parent, dob) in zip(
+            parent: dob
+            for (parent, dob) in zip(
                 parent_ids_idx.unique(),
-                pop_data.creation_time + pd.to_timedelta(
-                    np.floor(self.randomness.get_draw(parent_ids_idx.unique(), "dob") * self.step_size_days), unit="days"
-                )
+                pop_data.creation_time
+                + pd.to_timedelta(
+                    np.floor(
+                        self.randomness.get_draw(parent_ids_idx.unique(), "dob")
+                        * self.step_size_days
+                    ),
+                    unit="days",
+                ),
             )
         }
         new_births["date_of_birth"] = new_births["parent_id"].map(dob_map)
 
-        new_births["age"] = (pop_data.creation_time - new_births["date_of_birth"]).dt.days / DAYS_PER_YEAR
+        new_births["age"] = (
+            pop_data.creation_time - new_births["date_of_birth"]
+        ).dt.days / DAYS_PER_YEAR
 
         # add some noise because our randomness keys on entrance time and age,
         # so don't want people born same day to have same exact age
@@ -328,7 +335,9 @@ class Population:
         event : vivarium.framework.event.Event
 
         """
-        population = self.population_view.subview(["age"]).get(event.index, query="alive == 'alive'")
+        population = self.population_view.subview(["age"]).get(
+            event.index, query="alive == 'alive'"
+        )
         population["age"] += to_years(event.step_size)
         self.population_view.update(population)
 
--- a/src/vivarium_census_prl_synth_pop/components/population.py
+++ b/src/vivarium_census_prl_synth_pop/components/population.py
@@ -61,8 +61,7 @@ class Population:
             "entrance_time",
             "exit_time",
             "housing_type",
-            "dependents",
-            "guardians",
+            "guardian",
         ]
         self.register_simulants = builder.randomness.register_simulants
         self.population_view = builder.population.get_view(self.columns_created)
@@ -134,7 +133,7 @@ class Population:
         pop = pop.set_index(pop_data.index)
 
         # todo: initialize guardians for simulants: Should this be the last thing we do in initialize simulants?
-        pop = self.assign_guardians(pop)
+        pop = self.assign_gen_pop_guardians(pop)
 
         self.population_view.update(pop)
 
@@ -316,7 +315,7 @@ class Population:
         population["age"] += to_years(event.step_size)
         self.population_view.update(population)
 
-    def assign_guardians(self, pop: pd.DataFrame) -> pd.DataFrame:
+    def assign_gen_pop_guardians(self, pop: pd.DataFrame) -> pd.DataFrame:
         """
 
         Parameters
@@ -325,43 +324,78 @@ class Population:
 
         Returns
         -------
-        # todo: return a seaparate teable for assigned guardians
-        pd.Dataframe that will be pop with an additional guardians column which will be a list of one or two indexes or
-            "N/A" for that simulant's guaridans.
+        pd.Dataframe that will be pop with an additional guardians column containing the index for that simulant's
+          guardian..
         """
+        # Initialize column
+        pop["guardian"] = pd.NaN
+        # Helper lists
+        non_relatives = ["Roommate", "Other nonrelative"]
+        children = ["Biological child", "Adopted child", "Foster child", "Stepchild"]
+        parents = ["Parent", "Parent-in-law"]
+        partners = ["Opp-sex spouse", "Opp-sex partner", "Same-sex spouse", "Same-sex partner"]
 
-        pop["dependents"] = [[] for _ in range(len(pop))]
-        pop["guardians"] = [[] for _ in range(len(pop))]
+        # Get household structure for population to vectorize choices
+        # Non-GQ population
+        gen_population = pop.loc[
+            ~pop["household_id"].isin(data_values.GQ_HOUSING_TYPE_MAP)
+        ]
+        household_structure = self.get_household_structure(gen_population)
 
-        # Get index for 4 main groups in general population < 18 years old not in GQ
-        gen_pop_child_idx = pop.loc[
-            (pop["age"] < 18) & (pop["relation_to_household_head"].isin(
-                ["Biological child", "Adopted child", "Foster child", "Stepchild"])
-            )
-            ].index
-        gen_pop_relative_idx = pop.loc[
-            (pop["age"] < 18) & (pop["relation_to_household_head"].isin(
-                ["Other relative", "Grandchild", "Child-in-law", "Sibling"])
-            )
-            ].index
-        gen_pop_non_relative_idx = pop.loc[
-            (pop["age"] < 18) & (pop["relation_to_household_head"].isin(
-                ["Roommate", "Other nonrelative"])
-            )
-            ].index
-        gen_pop_child_reference_idx = pop.loc[
-            (pop["age"] < 18) & (pop["relation_to_household_head"] == "Reference person")
-            ].index
-        # todo: Find way to apply get_household_structure and determine guardian functions to each sim who needs it.
+        # Get indexes of subgroups to work through decision tree
+        # todo: Make subsets to get series
+        under_18_idx = household_structure.loc[
+            household_structure["age_x"] < 18
+        ].index
+
+        relatives_of_rp_idx = household_structure.loc[
+            ~household_structure["relation_to_household_head_y"].isin(non_relatives)
+        ].index
+
+        # Children of reference person - red box
+        children_of_rp = household_structure.loc[
+            under_18_idx.intersection(
+                household_structure.loc[
+                    (household_structure["relation_to_household_head_x"].isin(children)) &
+                    (household_structure["relation_to_household_head_y"] == "Reference person")
+                    ].index)
+        ].reset_index().groupby(["ch_id"])["id"].apply(np.random.choice)
+        pop.loc[children_of_rp.index, "guardian"] = children_of_rp
+
+
+
+        # Children are reference person and have a parent in household - green box
+        rp_children_with_parent = household_structure.loc[
+            under_18_idx.intersection(
+                household_structure.loc[
+                    (household_structure["relation_to_household_head_x"] == "Reference person") &
+                    (household_structure["relation_to_household_head_y"].isin(parents))
+                    ].index)].reset_index().groupby(["ch_id"])["id"].apply(np.random.choice)
+        pop.loc[rp_children_with_parent.index, "guardian"] = rp_children_with_parent
+
 
         # todo: Work through decision tree for < 24 year olds in GQ
 
-    def get_household_structure(self, pop: pd.DataFrame, sim_idx: pd.Index) -> pd.DataFrame:
-        # Get household structure from a given index in the state table
-        household_id = pop.loc[sim_idx, "household_id"]
-        household_structure = pop.loc[
-            pop["household_id"] == household_id, ["age", "household_id", "relation_to_household_head"]
-        ]
+    def get_household_structure(self, pop: pd.DataFrame) -> pd.DataFrame:
+        # Returns a 3 level multi-index with levels ["household_id", "child_id", "person_id"]
+        # Columns will contain data for child alongside each household member
+        # This will allow us to do lookups related to both a child and other household members
+        under_18 = (
+                    pop.loc[pop["age"] < 18, ["household_id", "relation_to_household_head", "age"]]
+                    .reset_index()
+                    .rename(columns={"index": "child_id"})
+                    .set_index(["household_id", "child_id"])
+                )
+        household_info = (
+            pop[["household_id", "relation_to_household_head", "age"]]
+            .reset_index()
+            .rename(columns={"index": "person_id"})
+            .set_index(["household_id", "person_id"])
+        )
+        # Merge dataframes to cast on household_id and child_id
+        household_structure = under_18.merge(household_info, left_index=True, right_index=True)
+        household_structure["age_difference"] = household_structure["age_y"] - household_structure["age_x"]
+
         return household_structure
 
     def determine_general_pop_guardians_and_dependents(
--- a/src/vivarium_census_prl_synth_pop/components/population.py
+++ b/src/vivarium_census_prl_synth_pop/components/population.py
@@ -1,6 +1,9 @@
+from loguru import logger
+import time
+
 import numpy as np
 import pandas as pd
-from typing import (Tuple, Union)
+
 from vivarium.framework.engine import Builder
 from vivarium.framework.event import Event
 from vivarium.framework.population import SimulantData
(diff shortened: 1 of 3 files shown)


src/vivarium_census_prl_synth_pop/components/population.py
```diff
@@ -60,6 +60,8 @@ def setup(self, builder: Builder):
             "entrance_time",
             "exit_time",
             "housing_type",
+            "guardian_1",
+            "guardian_2",
```
beatrixkh, line 64: i want to confirm: have the research team documented that a sim has a _max_ of 2 guardians?

src/vivarium_census_prl_synth_pop/components/population.py
```diff
@@ -317,6 +335,254 @@ def on_time_step__prepare(self, event: Event):
...
 
         """
-        population = self.population_view.subview(["age"]).get(event.index, query="alive == 'alive'")
+        population = self.population_view.subview(["age"]).get(
+            event.index, query="alive == 'alive'"
+        )
         population["age"] += to_years(event.step_size)
         self.population_view.update(population)
+
+    def assign_gen_pop_guardians(self, pop: pd.DataFrame) -> pd.DataFrame:
```
beatrixkh, line 344: i think this is worth naming more explicitly (if you havent thought about this in a while gen might be generation, generate, etc.)

src/vivarium_census_prl_synth_pop/components/population.py
```diff
@@ -317,6 +335,254 @@ def on_time_step__prepare(self, event: Event):
...
+        """
+
+        Parameters
+        ----------
+        pop: State stable of simulants
+
+        Returns
+        -------
+        pd.Dataframe that will be pop with an additional guardians column containing the index for that simulant's
```
beatrixkh, line 353: additional two columns?

src/vivarium_census_prl_synth_pop/components/population.py
```diff
@@ -317,6 +335,254 @@ def on_time_step__prepare(self, event: Event):
...
+        # Initialize column
+        pop["guardian_1"] = data_values.UNKNOWN_GUARDIAN_IDX
+        pop["guardian_2"] = data_values.UNKNOWN_GUARDIAN_IDX
+        # Helper lists
+        non_relatives = ["Roommate", "Other nonrelative"]
+        children = ["Biological child", "Adopted child", "Foster child", "Stepchild"]
+        parents = ["Parent", "Parent-in-law"]
+        partners = [
+            "Opp-sex spouse",
+            "Opp-sex partner",
```
beatrixkh, line 366: did the research team confirm and document that `partner`s in addition to `spouse`s should become additional guardians?

src/vivarium_census_prl_synth_pop/components/population.py
```diff
@@ -317,6 +335,254 @@ def on_time_step__prepare(self, event: Event):
...
+
+    @staticmethod
+    def get_household_structure(pop: pd.DataFrame) -> pd.DataFrame:
+        """
+
+        Parameters
+        ----------
+        pop: population state table
+
+        Returns
```
beatrixkh, line 548: can you add slightly more explanation to this? i think i understand it after looking at it for a while, but it took a bit. my understanding is:

for every household X containing a simulant under 18 years old, there will be N rows in the returned table corresponding to household X

N = (number of sub-18 year olds in household X) * (number of people in household X - 1)

then, each row of the table will denote a comparison between one such sub-18 year old and someone else in their household. the included columns are the two sims' ages, relationships to head of household, and their age difference

is that correct?

beatrixkh: ^ something to this effect would have helped me comprehend

beatrixkh: and / or, perhaps renaming "age_x" and "age_y" to "child_age" and "comparison_age" or something like that would have made it more readable without writing the kind of thing i just wrote

src/vivarium_census_prl_synth_pop/components/population.py
```diff
@@ -317,6 +335,254 @@ def on_time_step__prepare(self, event: Event):
...
+        ]
+
+        # Get household structure for population to vectorize choices
+        # Non-GQ population
+        gen_population = pop.loc[~pop["household_id"].isin(data_values.GQ_HOUSING_TYPE_MAP)]
+        household_structure = self.get_household_structure(gen_population)
+
+        # Children helper index groups
+        children_of_rp_idx = household_structure.loc[
+            household_structure["relation_to_household_head_x"].isin(children)
```
beatrixkh, line 378: i think im going to double down here on my comment within `self.get_household_structure` (sorry these will read out of order) -- it would be helpful if you rename "relation_to_household_head_x" to something like "childs_relation_to_household_head_x" (same for age_x and the two _y cols)

src/vivarium_census_prl_synth_pop/components/population.py
```diff
@@ -317,6 +335,254 @@ def on_time_step__prepare(self, event: Event):
...
+            "Same-sex partner",
+        ]
+
+        # Get household structure for population to vectorize choices
+        # Non-GQ population
+        gen_population = pop.loc[~pop["household_id"].isin(data_values.GQ_HOUSING_TYPE_MAP)]
+        household_structure = self.get_household_structure(gen_population)
+
+        # Children helper index groups
+        children_of_rp_idx = household_structure.loc[
```
beatrixkh, line 377: instead of `rp` could you do something like `ref_person`?

beatrixkh: or include a comment here like #children of reference person index groups so that reading thru the rest i know `rp` is reference person

src/vivarium_census_prl_synth_pop/components/population.py
```diff
@@ -317,6 +335,254 @@ def on_time_step__prepare(self, event: Event):
...
+        ].index
+
+        # Potential guardian/household member index groups
+        relatives_of_rp_idx = household_structure.loc[
+            ~household_structure["relation_to_household_head_y"].isin(non_relatives)
+        ].index
+        non_relatives_of_rp_idx = household_structure.loc[
+            household_structure["relation_to_household_head_y"].isin(non_relatives)
+        ].index
+        age_bound_idx = household_structure.loc[
```
beatrixkh, line 400: oh wow this was clever

src/vivarium_census_prl_synth_pop/components/population.py
```diff
@@ -317,6 +335,254 @@ def on_time_step__prepare(self, event: Event):
...
+            (household_structure["age_difference"] >= 20)
+            & (household_structure["age_difference"] < 46)
+        ].index
+        parents_of_rp_idx = household_structure.loc[
+            household_structure["relation_to_household_head_y"].isin(parents)
+        ].index
+        over_18_idx = household_structure.loc[household_structure["age_y"] >= 18].index
```
beatrixkh, line 407: should these people be excluded from righthand (_y) column from the start? isn't it the case that someone who is 18 or under cannot be a guardian? i might be misremembering how that turned out

src/vivarium_census_prl_synth_pop/components/population.py
```diff
@@ -317,6 +335,254 @@ def on_time_step__prepare(self, event: Event):
...
+        pop.loc[rp_parent_ids.index, "guardian_1"] = rp_parent_ids
+        # Assign partners of reference person as partners
+        partners_of_rp_parent_ids = household_structure.loc[
+            children_of_rp_idx.intersection(partners_idx)
+        ]
+        if len(partners_of_rp_parent_ids) > 0:
+            partners_of_rp_parent_ids = (
+                partners_of_rp_parent_ids.reset_index()
+                .groupby(["child_id"])["person_id"]
+                .apply(np.random.choice)
```
beatrixkh, line 431: it would be helpful to leave a comment here to note what you're doing re: selecting one at random if they have multiple partners

src/vivarium_census_prl_synth_pop/components/population.py
```diff
@@ -317,6 +335,254 @@ def on_time_step__prepare(self, event: Event):
...
+            "Opp-sex spouse",
+            "Opp-sex partner",
+            "Same-sex spouse",
+            "Same-sex partner",
+        ]
+
+        # Get household structure for population to vectorize choices
+        # Non-GQ population
+        gen_population = pop.loc[~pop["household_id"].isin(data_values.GQ_HOUSING_TYPE_MAP)]
+        household_structure = self.get_household_structure(gen_population)
```
rmudambi, line 374: Maybe rename to `child_households`?

src/vivarium_census_prl_synth_pop/components/population.py
```diff
@@ -317,6 +335,254 @@ def on_time_step__prepare(self, event: Event):
...
+            relatives_with_guardian_ids = (
+                relatives_with_guardian.reset_index()
+                .groupby(["child_id"])["person_id"]
+                .apply(np.random.choice)
+            )
+            pop.loc[
+                relatives_with_guardian_ids.index, "guardian_1"
+            ] = relatives_with_guardian_ids
+
+        # Children are relative of reference person with no age bound relative(s) - yellow box
```
beatrixkh, line 451: in addition to the "yellow box" comment here it would be helpful to include what the assignment is. like: 
`Children are relative of reference person with no age bound relative(s) - yellow box, assign reference person`

so that if youre going through this you dont have to also have the diagram up to know what this snippet of code is assigning

beatrixkh: --- fwiw the others were more readable to me, this one is hard to understand what you're doing

src/vivarium_census_prl_synth_pop/components/population.py
```diff
@@ -317,6 +335,254 @@ def on_time_step__prepare(self, event: Event):
...
+        gen_population = pop.loc[~pop["household_id"].isin(data_values.GQ_HOUSING_TYPE_MAP)]
+        household_structure = self.get_household_structure(gen_population)
+
+        # Children helper index groups
+        children_of_rp_idx = household_structure.loc[
+            household_structure["relation_to_household_head_x"].isin(children)
+        ].index
+        child_relative_of_rp_idx = household_structure.loc[
+            ~household_structure["relation_to_household_head_x"].isin(
+                non_relatives + children
```
rmudambi, line 382: Need to exclude reference person and partners as well here. Check if it's easier just to enumerate the non-child relations

src/vivarium_census_prl_synth_pop/components/population.py
```diff
@@ -317,6 +335,254 @@ def on_time_step__prepare(self, event: Event):
...
+            child_rp_with_relative_ids = (
+                child_rp_with_relative_ids.reset_index()
+                .groupby(["child_id"])["person_id"]
+                .apply(np.random.choice)
+            )
+            pop.loc[
+                child_rp_with_relative_ids.index, "guardian_1"
+            ] = child_rp_with_relative_ids
+
+        # Child is not related to and is not reference person, assign to age bound non-relative - blurple box
```
beatrixkh, line 506: is this blue-purple? that box appears to say assign any non-relative over 18. should it be over 18, or age-bound? if i was understanding correctly the "age-bound" was between 20-45 years older, right?

beatrixkh: oh wait i was misreading; thats the next one

beatrixkh (changes requested):
wow this was a lot; nice job. i just asked for some additional comments throughout code to make it more readable

src/vivarium_census_prl_synth_pop/components/population.py
```diff
@@ -317,6 +335,254 @@ def on_time_step__prepare(self, event: Event):
...
+            ~household_structure["relation_to_household_head_x"].isin(
+                non_relatives + children
+            )
+        ].index
+        child_non_relative_of_rp_idx = household_structure.loc[
+            household_structure["relation_to_household_head_x"].isin(non_relatives)
+        ].index
+        child_rp_idx = household_structure.loc[
+            (household_structure["relation_to_household_head_x"] == "Reference person")
+            & (household_structure["age_x"] < 18)
```
rmudambi, line 390: The age check isn't needed because person x is guaranteed to be under 18

src/vivarium_census_prl_synth_pop/components/population.py
rmudambi, line 377: For consistency call this `child_of_rp_idx`

beatrixkh: ^ after seeing how many times you used `rp` im maybe leaning more towards just defining it here / near the top before using it a bunch -- i guess depending on the degree to which actually renaming it would make the lines long/unwieldy

src/vivarium_census_prl_synth_pop/components/population.py
```diff
@@ -317,6 +335,254 @@ def on_time_step__prepare(self, event: Event):
...
+        ].index
+
+        # Potential guardian/household member index groups
+        relatives_of_rp_idx = household_structure.loc[
+            ~household_structure["relation_to_household_head_y"].isin(non_relatives)
```
rmudambi, line 395: I think partners and reference person should be excluded in addition to non-relatives

src/vivarium_census_prl_synth_pop/components/population.py
```diff
@@ -317,6 +335,254 @@ def on_time_step__prepare(self, event: Event):
...
+        ].index
+        non_relatives_of_rp_idx = household_structure.loc[
+            household_structure["relation_to_household_head_y"].isin(non_relatives)
+        ].index
+        age_bound_idx = household_structure.loc[
+            (household_structure["age_difference"] >= 20)
```
rmudambi, line 401: Confirm the lower bound should be 20 and not something like 15, since this essentially denies the existence of teen pregnancy

src/vivarium_census_prl_synth_pop/components/population.py
```diff
@@ -317,6 +335,254 @@ def on_time_step__prepare(self, event: Event):
...
+            .set_index("child_id")["person_id"]
+        )
+        pop.loc[relative_ids.index, "guardian_1"] = relative_ids
+        # Assign guardian to spouse/partner of reference person
+        relative_with_rp_partner_guardian_ids = household_structure.loc[
+            child_relative_of_rp_idx.drop(
+                relatives_with_guardian.index.unique("child_id"), level="child_id"
+            ).intersection(partners_idx)
+        ]
+        if len(relative_with_rp_partner_guardian_ids) > 0:
```
rmudambi, line 468: This part inside the if block can be refactored to avoid code duplication with the previous time you assigned partners to guardian 2

src/vivarium_census_prl_synth_pop/components/population.py
```diff
@@ -317,6 +335,254 @@ def on_time_step__prepare(self, event: Event):
...
+
+        # Children are reference person and have a parent in household - green box
+        child_rp_with_parent = household_structure.loc[
+            child_rp_idx.intersection(parents_of_rp_idx)
+        ]
+        if len(child_rp_with_parent) > 0:
+            child_rp_with_parent_ids = (
+                child_rp_with_parent.reset_index()
+                .groupby(["child_id"])["person_id"]
+                .apply(np.random.choice)
```
rmudambi, line 486: I thought we were supposed to assign a random two parents if there are multiple?

src/vivarium_census_prl_synth_pop/components/population.py
```diff
@@ -317,6 +335,254 @@ def on_time_step__prepare(self, event: Event):
...
+            )
+            pop.loc[child_rp_with_parent_ids.index, "guardian_1"] = child_rp_with_parent_ids
+
+        # Child is reference person with no parent, assign to age bound relative - blue box
+        child_rp_with_relative_ids = household_structure.loc[
+            child_rp_idx.drop(child_rp_with_parent.index.unique("child_id"), level="child_id")
```
rmudambi, line 492: You should be able to avoid code duplication as i think this logic is identical to that for box yellow?

rmudambi (changes requested):
Looks great, but I think there are few gaps that need to get RT answers. I think your parent of child RP scenario is missing guardian 2.

src/vivarium_census_prl_synth_pop/components/population.py
albrja: Yes.

src/vivarium_census_prl_synth_pop/components/population.py
albrja: Yes.

src/vivarium_census_prl_synth_pop/components/population.py
albrja: TRhis is correct but I can update the doc string and the column names.

src/vivarium_census_prl_synth_pop/components/population.py
albrja: Changed to 18 per research team.

src/vivarium_census_prl_synth_pop/components/population.py
albrja: Syl confirmed to keep it at 1 for now.  There was confusion between the documentation and what was stated in a slack thread.

src/vivarium_census_prl_synth_pop/components/population.py
```diff
@@ -317,6 +335,254 @@ def on_time_step__prepare(self, event: Event):
...
+        """
+
+        # Initialize column
+        pop["guardian_1"] = data_values.UNKNOWN_GUARDIAN_IDX
+        pop["guardian_2"] = data_values.UNKNOWN_GUARDIAN_IDX
+        # Helper lists
+        non_relatives = ["Roommate", "Other nonrelative"]
+        children = ["Biological child", "Adopted child", "Foster child", "Stepchild"]
+        parents = ["Parent", "Parent-in-law"]
+        partners = [
```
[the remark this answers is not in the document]
albrja: After talking to Syl we are going to leave it as N/A (-1).

src/vivarium_census_prl_synth_pop/components/population.py
albrja: This is the only comment I did not address.  The logic is the same but the lookups are slightly different.

src/vivarium_census_prl_synth_pop/components/population.py
```diff
@@ -317,6 +335,267 @@ def on_time_step__prepare(self, event: Event):
...
+        child_households = self.get_household_structure(gen_population)
+
+        # Children helper index groups
+        # Ref_person = "Reference person"
+        child_of_ref_person_idx = child_households.loc[
+            child_households["child_relation_to_household_head"].isin(children)
+        ].index
+        child_relative_of_ref_person_idx = child_households.loc[
+            child_households["child_relation_to_household_head"].isin(
+                ["Sibling", "Other relative", "Grandchild", "Child-in-law"]
```
rmudambi, line 383: Store this other relatives list up with your other relative lists

src/vivarium_census_prl_synth_pop/components/population.py
```diff
@@ -371,165 +371,171 @@ def assign_gen_pop_guardians(self, pop: pd.DataFrame) -> pd.DataFrame:
...
         ]
-        if len(partners_of_rp_parent_ids) > 0:
-            partners_of_rp_parent_ids = (
-                partners_of_rp_parent_ids.reset_index()
-                .groupby(["child_id"])["person_id"]
-                .apply(np.random.choice)
+        if len(partners_of_ref_person_parent_ids) > 0:
+            # Select random partner of reference person and assign as second guardian
+            partners_of_ref_person_parent_ids = (
+                choose_random_guardian(partners_of_ref_person_parent_ids)
```
rmudambi: Nit: don't put this expression in parentheses

rmudambi (approved):

beatrixkh (approved):

PR #82, opened 2022-11-22
Feature/newborn guardians

albrja:
### Assigned guardians to newborns each time step.
- *Category*: Feature
- *JIRA issue*: [MIC-3595](https://jira.ihme.washington.edu/browse/MIC-3595)
- *Research reference*: https://vivarium-research.readthedocs.io/en/latest/concept_models/vivarium_census_synthdata/concept_model.html#initializing-guardian-s-for-all-simulants

-Updated helper function to choose random guardian if multiple by adding additional argument for groupby call.
-Added function to create mothers household structure
-Assign guardian_1 and guardian_2 to newborns.

### Verification and Testing
Successfully ran simulation and saw updated guardian columns for newborns.
new_births_1
new_births_2
new_births_3

### DIFF
--- a/src/vivarium_census_prl_synth_pop/components/population.py
+++ b/src/vivarium_census_prl_synth_pop/components/population.py
@@ -236,7 +236,9 @@ class Population:
         parent_ids_idx = pop_data.user_data["parent_ids"]
         pop_index = pop_data.user_data["current_population_index"]
         mothers = self.population_view.get(parent_ids_idx.unique())
-        households = self.population_view.subview(["household_id", "relation_to_household_head"]).get(pop_index)
+        households = self.population_view.subview(
+            ["household_id", "relation_to_household_head"]
+        ).get(pop_index)
         # Making separate subviews because SSNS will be moved to post-processing
         ssns = self.population_view.subview(["ssn"]).get(pop_index).squeeze()
         new_births = pd.DataFrame(data={"parent_id": parent_ids_idx}, index=pop_data.index)
@@ -447,7 +449,9 @@ class Population:
         ]
         if len(relatives_with_guardian) > 0:
             # Select random relative if multiple and assign as guardian
-            relatives_with_guardian_ids = self.choose_random_guardian(relatives_with_guardian, "child_id")
+            relatives_with_guardian_ids = self.choose_random_guardian(
+                relatives_with_guardian, "child_id"
+            )
             pop.loc[
                 relatives_with_guardian_ids.index, "guardian_1"
             ] = relatives_with_guardian_ids
@@ -515,7 +519,9 @@ class Population:
         ]
         if len(non_relative_guardian) > 0:
             # Select random non-relative if multiple and assign to guardian
-            non_relative_guardian_ids = self.choose_random_guardian(non_relative_guardian, "child_id")
+            non_relative_guardian_ids = self.choose_random_guardian(
+                non_relative_guardian, "child_id"
+            )
             pop.loc[non_relative_guardian_ids.index, "guardian_1"] = non_relative_guardian_ids
 
         # Child is not reference person with no age bound non-relative, assign to adult non-relative - purple box
@@ -607,7 +613,9 @@ class Population:
         )
         return member_ids
 
-    def assign_second_guardian_to_newborns(self, new_births: pd.DataFrame, households: pd.DataFrame) -> pd.DataFrame:
+    def assign_second_guardian_to_newborns(
+        self, new_births: pd.DataFrame, households: pd.DataFrame
+    ) -> pd.DataFrame:
         """
         Parameters
         ----------
@@ -621,7 +629,9 @@ class Population:
         """
         # Setup
         new_births["guardian_2"] = data_values.UNKNOWN_GUARDIAN_IDX
-        mothers_households = self.get_mothers_household_structure(new_births["parent_id"], households)
+        mothers_households = self.get_mothers_household_structure(
+            new_births["parent_id"], households
+        )
         partners = [
             "Opp-sex spouse",
             "Opp-sex partner",
@@ -633,7 +643,7 @@ class Population:
         # Mother index groups
         mother_rp_idx = mothers_households.loc[
             mothers_households["mother_relation_to_household_head"] == "Reference person"
-            ].index
+        ].index
         mother_partner_idx = mothers_households.loc[
             mothers_households["mother_relation_to_household_head"].isin(partners)
         ].index
@@ -643,13 +653,10 @@ class Population:
         ].index
         reference_person_idx = mothers_households.loc[
             mothers_households["member_relation_to_household_head"] == "Reference person"
-            ].index
+        ].index
 
         # Assign second guardian to random partner of mothers
-        partner_ids = mothers_households.loc[
-            mother_rp_idx
-            .intersection(partners_idx)
-        ]
+        partner_ids = mothers_households.loc[mother_rp_idx.intersection(partners_idx)]
         if len(partner_ids) > 0:
             partner_ids = self.choose_random_guardian(partner_ids, "mother_id")
             new_births.loc[
@@ -657,11 +664,12 @@ class Population:
             ] = new_births["parent_id"].map(partner_ids)
 
         reference_person_ids = mothers_households.loc[
-            mother_partner_idx
-            .intersection(reference_person_idx)
+            mother_partner_idx.intersection(reference_person_idx)
         ]
         if len(reference_person_ids) > 0:
-            reference_person_ids = self.choose_random_guardian(reference_person_ids, "mother_id")
+            reference_person_ids = self.choose_random_guardian(
+                reference_person_ids, "mother_id"
+            )
             new_births.loc[
                 new_births["parent_id"].isin(reference_person_ids.index), "guardian_2"
             ] = new_births["parent_id"].map(reference_person_ids)
@@ -669,7 +677,9 @@ class Population:
         return new_births
 
     @staticmethod
-    def get_mothers_household_structure(mothers_idx: pd.Series, households: pd.DataFrame) -> pd.DataFrame:
+    def get_mothers_household_structure(
+        mothers_idx: pd.Series, households: pd.DataFrame
+    ) -> pd.DataFrame:
         """
 
         Parameters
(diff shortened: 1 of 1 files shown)


src/vivarium_census_prl_synth_pop/components/population.py
```diff
@@ -594,11 +603,119 @@ def get_household_structure(pop: pd.DataFrame) -> pd.DataFrame:
...
         # member_ids is a subset of child_households dataframe
+        # groupby_level will be index level to group by (the first level index of member_ids (child_id or mother_id).
         member_ids = (
             member_ids.reset_index()
-            .groupby(["child_id"])["person_id"]
+            .groupby([groupby_level])["person_id"]
             .apply(np.random.choice)
         )
         return member_ids
+
```
albrja, line 615: Everything above here is just a small refactor and black updates.  Everything below here is newly implemented code.

src/vivarium_census_prl_synth_pop/components/population.py
```diff
@@ -594,11 +603,119 @@ def get_household_structure(pop: pd.DataFrame) -> pd.DataFrame:
...
     @staticmethod
-    def choose_random_guardian(member_ids: pd.DataFrame) -> pd.Series:
+    def choose_random_guardian(member_ids: pd.DataFrame, groupby_level: str) -> pd.Series:
         # member_ids is a subset of child_households dataframe
+        # groupby_level will be index level to group by (the first level index of member_ids (child_id or mother_id).
         member_ids = (
             member_ids.reset_index()
-            .groupby(["child_id"])["person_id"]
+            .groupby([groupby_level])["person_id"]
             .apply(np.random.choice)
```
stevebachmeier, line 612: Not part of this PR, but is `np.random.choice` appropriate or does it need to be a predictable randomness stream like we usually do?

src/vivarium_census_prl_synth_pop/components/population.py
```diff
@@ -321,9 +325,10 @@ def initialize_newborns(self, pop_data: SimulantData) -> None:
 
         # typing
         new_births["household_id"] = new_births["household_id"].astype(int)
-        # todo: Update guardian columns MIC-3595
-        new_births["guardian_1"] = data_values.UNKNOWN_GUARDIAN_IDX
-        new_births["guardian_2"] = data_values.UNKNOWN_GUARDIAN_IDX
+        # Assign guardian_1 to birth mother
+        new_births["guardian_1"] = new_births["parent_id"]
```
stevebachmeier, line 329: `parent_id` is definitively the mother?

src/vivarium_census_prl_synth_pop/components/population.py
```diff
@@ -594,11 +603,119 @@ def get_household_structure(pop: pd.DataFrame) -> pd.DataFrame:
...
+            .set_index(["household_id", "person_id"])
+        )
+
+        mother_households = mothers.merge(household_info, left_index=True, right_index=True)
+        mother_households.rename(
+            columns={
+                "relation_to_household_head_x": "mother_relation_to_household_head",
+                "relation_to_household_head_y": "member_relation_to_household_head",
+            },
+            inplace=True,
```
stevebachmeier, line 717: nit: you `inplace=True` here but then overwrite in the next `.droplevel` step. Consistency is nice

rmudambi: +1

src/vivarium_census_prl_synth_pop/components/population.py
```diff
@@ -594,11 +603,119 @@ def get_household_structure(pop: pd.DataFrame) -> pd.DataFrame:
...
+            reference_person_ids = self.choose_random_guardian(
+                reference_person_ids, "mother_id"
+            )
+            new_births.loc[
+                new_births["parent_id"].isin(reference_person_ids.index), "guardian_2"
+            ] = new_births["parent_id"].map(reference_person_ids)
+
+        return new_births
+
+    @staticmethod
```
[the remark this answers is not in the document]
rmudambi: Pycharm encourages you to make methods static if they can be. This function doesn't need access to the `self` object so it can function as a static method. See here for some insights into the benefits - https://stackoverflow.com/questions/2438473/what-is-the-advantage-of-using-static-methods. In this case, it makes very little difference.

src/vivarium_census_prl_synth_pop/components/population.py
rmudambi: Since we're not running multiple scenarios the common random number (CRN) framework is less valuable. Also, since the point is to generate simulated data, the specifics don't really matter as long as they're internally consistent. 

The most important thing is to be able to harness numpy for performance reasons.

src/vivarium_census_prl_synth_pop/components/population.py
```diff
@@ -594,11 +603,119 @@ def get_household_structure(pop: pd.DataFrame) -> pd.DataFrame:
...
+        Returns
+        -------
+        new_births with the additional guardian_2 column updated.
+        """
+        # Setup
+        new_births["guardian_2"] = data_values.UNKNOWN_GUARDIAN_IDX
+        mothers_households = self.get_mothers_household_structure(
+            new_births["parent_id"], households
+        )
+        partners = [
```
rmudambi, line 635: I'd suggest defining all these relationship lists as constants outside the functions, since you're using the same list of relationships for partners in both functions now.

src/vivarium_census_prl_synth_pop/components/population.py
```diff
@@ -594,11 +603,119 @@ def get_household_structure(pop: pd.DataFrame) -> pd.DataFrame:
...
+            "Opp-sex spouse",
+            "Opp-sex partner",
+            "Same-sex spouse",
+            "Same-sex partner",
+        ]
+
+        # Index helpers
+        # Mother index groups
+        mother_rp_idx = mothers_households.loc[
```
rmudambi, line 644: rp => ref_person

src/vivarium_census_prl_synth_pop/components/population.py
```diff
@@ -594,11 +603,119 @@ def get_household_structure(pop: pd.DataFrame) -> pd.DataFrame:
...
+            mothers_households["member_relation_to_household_head"] == "Reference person"
+        ].index
+
+        # Assign second guardian to random partner of mothers
+        partner_ids = mothers_households.loc[mother_rp_idx.intersection(partners_idx)]
+        if len(partner_ids) > 0:
+            partner_ids = self.choose_random_guardian(partner_ids, "mother_id")
+            new_births.loc[
+                new_births["parent_id"].isin(partner_ids.index), "guardian_2"
+            ] = new_births["parent_id"].map(partner_ids)
```
rmudambi, line 664: Why isn't this just
```
new_births.loc[partner_ids.index, "guardian_2"] = partner_ids
```

src/vivarium_census_prl_synth_pop/components/population.py
```diff
@@ -594,11 +603,119 @@ def get_household_structure(pop: pd.DataFrame) -> pd.DataFrame:
...
+        reference_person_ids = mothers_households.loc[
+            mother_partner_idx.intersection(reference_person_idx)
+        ]
+        if len(reference_person_ids) > 0:
+            reference_person_ids = self.choose_random_guardian(
+                reference_person_ids, "mother_id"
+            )
+            new_births.loc[
+                new_births["parent_id"].isin(reference_person_ids.index), "guardian_2"
+            ] = new_births["parent_id"].map(reference_person_ids)
```
rmudambi, line 675: Same question as above

rmudambi (approved):
Looks good, I just have a few questions that I'd like answers to at stand-up tomorrow

src/vivarium_census_prl_synth_pop/components/population.py
```diff
@@ -594,11 +603,119 @@ def get_household_structure(pop: pd.DataFrame) -> pd.DataFrame:
...
+        if len(partner_ids) > 0:
+            partner_ids = self.choose_random_guardian(partner_ids, "mother_id")
+            new_births.loc[
+                new_births["parent_id"].isin(partner_ids.index), "guardian_2"
+            ] = new_births["parent_id"].map(partner_ids)
+
+        reference_person_ids = mothers_households.loc[
+            mother_partner_idx.intersection(reference_person_idx)
+        ]
+        if len(reference_person_ids) > 0:
```
stevebachmeier, line 669: Consider making a helper function since these two `if` statements are identical.

src/vivarium_census_prl_synth_pop/components/population.py
```diff
@@ -594,11 +603,119 @@ def get_household_structure(pop: pd.DataFrame) -> pd.DataFrame:
...
+            reference_person_ids = self.choose_random_guardian(
+                reference_person_ids, "mother_id"
+            )
+            new_births.loc[
+                new_births["parent_id"].isin(reference_person_ids.index), "guardian_2"
```
stevebachmeier, line 674: So if the parent id is not in the curated list of possible guardians then it just stays with data_values.UNKNOWN_GUARDIAN_IDX? I would have thought it would go to the next family member on some hierarchy of something like that (but I haven't read the docs in depth either)

src/vivarium_census_prl_synth_pop/components/population.py
albrja: Yes.  That how it was defined in earlier versions before the guardian functionality was known to us.

src/vivarium_census_prl_synth_pop/components/population.py
```diff
@@ -594,11 +603,119 @@ def get_household_structure(pop: pd.DataFrame) -> pd.DataFrame:
...
+            mothers_households["member_relation_to_household_head"].isin(partners)
+        ].index
+        reference_person_idx = mothers_households.loc[
+            mothers_households["member_relation_to_household_head"] == "Reference person"
+        ].index
+
+        # Assign second guardian to random partner of mothers
+        partner_ids = mothers_households.loc[mother_rp_idx.intersection(partners_idx)]
+        if len(partner_ids) > 0:
+            partner_ids = self.choose_random_guardian(partner_ids, "mother_id")
```
stevebachmeier, line 661: This needs a method name change if it's now used for something other than choosing random guardians

src/vivarium_census_prl_synth_pop/components/population.py
stevebachmeier, line 664: So if the partner of a a mother does not live in the household then the child gets assigned a guardian of `data_values.UNKNOWN_GUARDIAN_IDX`? It seems like there would be some functionality to have non-partner guardians but I admittedly haven't read through the docs

src/vivarium_census_prl_synth_pop/components/population.py
```diff
@@ -577,28 +575,139 @@ def get_household_structure(pop: pd.DataFrame) -> pd.DataFrame:
...
+non_relatives = ["Roommate", "Other nonrelative"]
+children = ["Biological child", "Adopted child", "Foster child", "Stepchild"]
+children_relatives = ["Sibling", "Other relative", "Grandchild", "Child-in-law"]
+parents = ["Parent", "Parent-in-law"]
+partners = [
+    "Opp-sex spouse",
+    "Opp-sex partner",
+    "Same-sex spouse",
+    "Same-sex partner",
+]
```
albrja, line 713: @beatrixkh Seems likely these will be useful later on for observers or other components so if you think they should live in data values I can move them there.

src/vivarium_census_prl_synth_pop/components/population.py
stevebachmeier: Oh I'm confused then. I thought this was choosing a random partner for the female which then may or may not be assigned as guardian of the child.

albrja: choose_random_guardian is picking a random guardian from a subset of mother_households here (or in the other use cases child_households).  This is why there is always a preceding check for len since you can't groupby on an object of length 0.

PR #93, opened 2022-12-14
Refactor census observer to inherit from BaseObserver

aflaxman:
### Description

- *Category*: refactor 
- *JIRA issue*: MIC-3642
- *Research reference*: https://vivarium-research.readthedocs.io/en/latest/models/concept_models/vivarium_census_synthdata/concept_model.html?highlight=census%20observer#census

### Changes and notes
 Refactor census observer to inherit from BaseObserver

### Verification and Testing
 pytest ran successfully, 
```
time simulate run --pdb -v src/vivarium_census_prl_synth_pop/model_specifications/model_spec.yaml
```
also completed successfully, examined files created and they looked as expected.

image


src/vivarium_census_prl_synth_pop/components/observers.py
```diff
@@ -89,78 +89,47 @@ def on_simulation_end(self, event: Event) -> None:
...
-                f"ymd_{self.start_date.year}_{self.start_date.month}_{self.start_date.day}"
-            )
-            state_table = self.population_view.get(event.index)
-            state_table.to_hdf(self.output_path, start_date_str)
-        self.counter += 1
-        if (self.clock().year % 10 == 0) & (self.clock().month == 4):
-            if self.clock().day < 29:  # because we only want one observation in April
-                self.decennial_census(event, hdf_key=f"year_{self.clock().year}")
+    def get_responses(self) -> pd.DataFrame:
+        return pd.DataFrame()
```
stevebachmeier, line 119: This will work but won't lend itself to enforcing schema/columns. Not sure how important that is at this point.

src/vivarium_census_prl_synth_pop/components/observers.py
```diff
@@ -89,78 +89,47 @@ def on_simulation_end(self, event: Event) -> None:
...
+        if (self.clock().year % 10 == 0) & (self.clock().month == 4):
+            if self.clock().day < 29:  # because we only want one observation in April  FIXME: cooler to do this with the timestep
+                return True
 
-    def decennial_census(self, event: Event, hdf_key) -> None:
-        pop = self.population_view.subview(metadata.DECENNIAL_CENSUS_COLUMNS_USED).get(
-            event.index
+    def do_observation(self, event) -> None:
+        pop = self.population_view.get(
+            event.index, # query="alive == 'alive'",  # TODO: uncomment this to include only living simulants in census
```
stevebachmeier, line 128: Why have it commented out at all?

src/vivarium_census_prl_synth_pop/components/observers.py
```diff
@@ -89,78 +89,47 @@ def on_simulation_end(self, event: Event) -> None:
...
-                self.decennial_census(event, hdf_key=f"year_{self.clock().year}")
+    def get_responses(self) -> pd.DataFrame:
+        return pd.DataFrame()
 
-    def on_simulation_end(self, event: Event) -> None:
-        end_date_str = f"ymd_{self.end_date.year}_{self.end_date.month}_{self.end_date.day}"
-        state_table = self.population_view.get(event.index)
-        state_table.to_hdf(self.output_path, end_date_str)
+    def to_observe(self, event: Event) -> bool:
+        if (self.clock().year % 10 == 0) & (self.clock().month == 4):
```
stevebachmeier, line 122: Unless I'm missing something, as implemented, the two datasets will be identical. I think what you want is to check that `self.clock().year == self.census_date.year`

README.rst
```diff
@@ -109,7 +109,7 @@ Running Simulations
 
 You can then run simulations by, e.g.::
 
-   (vivarium_census_prl_synth_pop) :~$ simulate run -v src/vivarium_census_prl_synth_pop/model_specifications/model_spec.yaml
+   (vivarium_census_prl_synth_pop) :~$ simulate run --pdb -v src/vivarium_census_prl_synth_pop/model_specifications/model_spec.yaml
```
rmudambi, line 112: This is atypical for our readmes. Is there a reason you're adding this here? Even if we want to make this change, PRs should be targeted and not include unrelated changes

src/vivarium_census_prl_synth_pop/components/observers.py
rmudambi: The purpose of this method is to define the columns we expect explicitly.

src/vivarium_census_prl_synth_pop/components/observers.py
```diff
@@ -89,78 +89,47 @@ def on_simulation_end(self, event: Event) -> None:
         self.responses.to_hdf(self.output_dir / self.output_filename, key="responses")
 
 
-# FIXME: give this a more descriptive name (eg CensusObserver)
-class Observers:
-    # FIXME: the docstring is no longer correct; update it
+class CensusObserver(BaseObserver):
```
rmudambi, line 92: The intention is to have a single Observer for the census, not a separate component for each time we want to take the census.

src/vivarium_census_prl_synth_pop/components/observers.py
rmudambi: What we actually want here is 
```
return event.time.year % 10 == 0 and event.time.month == 4 and event.time.day <= self.time_step
```
so that we observe on the first April timestep of each decade. Note: `self.time_step` should be retrieved during `setup` so that it refers to the actual time-step length used in the simulation rather than being hard-coded as below.

src/vivarium_census_prl_synth_pop/components/observers.py
rmudambi: And to be clear, I'd expect the schema to have a column for the census sample date (could be an index column if you prefer), and then a column for each of the columns you are recording in the census

rmudambi (changes requested):

src/vivarium_census_prl_synth_pop/components/observers.py
stevebachmeier: But he is instantiating the obsever twice, each time with a different date string, eg
- CensusObserver('2020-04-01')
- CensusObserver('2030-04-01')

src/vivarium_census_prl_synth_pop/components/observers.py
aflaxman: I'll follow your lead, what were you thinking for this?

src/vivarium_census_prl_synth_pop/components/observers.py
aflaxman: I figured I'd make a separate PR with the changes beyond refactoring the existing code to work with your BaseObserver class.  My intention is to keep things easy to review.

src/vivarium_census_prl_synth_pop/components/observers.py
aflaxman: You are correct, I plan to add improved logic for this in a future PR, by testing if the `census_date` is between `event.time` and `self.clock.time()`.  As above, I thought leaving it for a separate PR would make it simpler for code review... sorry if I misjudged that!

README.rst
aflaxman: oops, unintentional inclusion.

```suggestion
   (vivarium_census_prl_synth_pop) :~$ simulate run -v src/vivarium_census_prl_synth_pop/model_specifications/model_spec.yaml
```

src/vivarium_census_prl_synth_pop/components/observers.py
aflaxman: I think it will be simpler to include a separate observer for each decade.  Let me see if I can convince you.  The key points are (1) it will simplify the logic of testing if the timestep is right for observation; and (2) it will make it easier to have multiple censuses for testing purposes without running for a full decade.  Let's discuss in real time if necessary.

src/vivarium_census_prl_synth_pop/components/observers.py
aflaxman: Ah, got it.  How about if I add this in a future PR, where I update the code to include the required columns?

src/vivarium_census_prl_synth_pop/components/observers.py
aflaxman: But before I forget (3) if we have a single observer we have to (a) put all years in a single table or (b) extend the BaseObserver to allow (b1) saving multiple keys or (b2) saving multiple files, and I think these tables are big enough that having them in separate files is better than options (a) or (b1).  This approach accomplishes something equivalent to (b2) without requiring changes to BaseObserver.

src/vivarium_census_prl_synth_pop/components/observers.py
aflaxman: Rajan strongly prefers an approach with a single observer, so that we can change the duration of the sim and get all the decennial censuses observed without having to add additional `CensusObserver` lines to the model spec, which I have implemented in e6c49f7f4e0a888f329bf362481af84b110032a0

src/vivarium_census_prl_synth_pop/components/observers.py
aflaxman: Rajan agreed that I should add this in a future PR

src/vivarium_census_prl_synth_pop/components/observers.py
aflaxman: Rajan strongly prefers an approach with a single observer, so that we can change the duration of the sim and get all the decennial censuses observed without having to add additional CensusObserver lines to the model spec, which I have implemented in https://github.com/ihmeuw/vivarium_census_prl_synth_pop/commit/e6c49f7f4e0a888f329bf362481af84b110032a0

src/vivarium_census_prl_synth_pop/components/observers.py
```diff
@@ -89,78 +89,43 @@ def on_simulation_end(self, event: Event) -> None:
...
-                self.decennial_census(event, hdf_key=f"year_{self.clock().year}")
+    def get_responses(self) -> pd.DataFrame:
+        return pd.DataFrame()  # TODO: include schema here, including column for census year
 
-    def on_simulation_end(self, event: Event) -> None:
-        end_date_str = f"ymd_{self.end_date.year}_{self.end_date.month}_{self.end_date.day}"
-        state_table = self.population_view.get(event.index)
-        state_table.to_hdf(self.output_path, end_date_str)
+    def to_observe(self, event: Event) -> bool:
+        if (self.clock().year % 10 == 0) & (self.clock().month == 4):
```
rmudambi, line 118: You should be using `event.time` not `self.clock()`. The previous code was incorrect.

Is there a reason you are using `&` rather than `and`?

This observer should return `False` when we're not supposed to observe. Currently it returns `None` which means it will work, since `None` is falsey, but this is bad practice. It's much better in a boolean function like this just to return the conditional you currently have in an `if` statement like I noted in my above comment.

src/vivarium_census_prl_synth_pop/components/observers.py
```diff
@@ -89,78 +89,43 @@ def on_simulation_end(self, event: Event) -> None:
...
-                f"ymd_{self.start_date.year}_{self.start_date.month}_{self.start_date.day}"
-            )
-            state_table = self.population_view.get(event.index)
-            state_table.to_hdf(self.output_path, start_date_str)
-        self.counter += 1
-        if (self.clock().year % 10 == 0) & (self.clock().month == 4):
-            if self.clock().day < 29:  # because we only want one observation in April
-                self.decennial_census(event, hdf_key=f"year_{self.clock().year}")
+    def get_responses(self) -> pd.DataFrame:
+        return pd.DataFrame()  # TODO: include schema here, including column for census year
```
rmudambi, line 115: Ideally this would be set in this PR

rmudambi (approved):
Fine to merge once you address the small notes

src/vivarium_census_prl_synth_pop/components/observers.py
```diff
@@ -89,78 +89,43 @@ def on_simulation_end(self, event: Event) -> None:
...
         pop = pop.drop(columns="middle_name")
 
-        # we don't have a 100% census response rate:
-        respondents = self.randomness.filter_for_probability(
-            pop, self.response_probability_decennial(pop.index), "decennial_respondents"
-        )
+        # TODO: include additional columns specified in MIC-3642
 
-        respondents.to_hdf(self.decennial_path, hdf_key)
+        self.responses = pd.concat([self.responses, pop])
```
stevebachmeier, line 131: As is, this is going to save out a single hdf with a single default key. The only way to differentiate which census is by looking at the date column. Maybe this is fine?

Another option is to get clever with `on_simulation_end` and wrap the saving to hdf part in a method that can be overwritten. Then in this class you can subset to however many cencus' you have in the object and save out w/ different keys. Thoughts?

@rmudambi

src/vivarium_census_prl_synth_pop/components/observers.py
aflaxman: I am trying to make this PR simply a refactor of the existing observer, and I would prefer to improve this logic in a future PR --- The logic here is identical to what was present before. I can improve it here if you want, but I think it will be clearer to keep the shift to BaseObserver separate from the other changes.

src/vivarium_census_prl_synth_pop/components/observers.py
aflaxman: I will add

src/vivarium_census_prl_synth_pop/components/observers.py
aflaxman: I'm going to save all censuses in one table for now, but Rajan suggested that I could override on_simulation_end if we decided that we wanted them to be in separate files.

PR #94, opened 2022-12-14
Select individual movers of each type using rates

zmbc:
### Description

- *Category*: implementation
- *JIRA issue*: [MIC-3672](https://jira.ihme.washington.edu/browse/MIC-3672)
- *Research reference*: https://vivarium-research.readthedocs.io/en/latest/models/concept_models/vivarium_census_synthdata/concept_model.html#individual-moves

### Changes and notes

Unlike what I originally wrote in the ticket, I have made it so that _all_ move types are passed onto the old moving logic (for now), not only non-reference person movers. Seemed to make more sense because the old move logic encapsulated both non-reference person and GQ person moves.

I believe when we get to immigration/emigration, we will want to rename this component to make it clear that it is only about domestic migration.

### Verification and Testing

Ran simulation without error; stepped through code to check that reasonable numbers of people were moving.


src/vivarium_census_prl_synth_pop/components/person.py
```diff
@@ -88,30 +82,23 @@ def on_time_step(self, event: Event) -> None:
...
-            persons.relation_to_household_head != "Reference person"
-        ]
 
-        # Get subsets of possible simulants that can move
-        persons_who_move_idx = self.randomness.filter_for_rate(
-            non_household_heads.index,
-            self.person_move_rate(non_household_heads.index),
-            "all_movers",
+        # Get subsets of simulants that do each move type on this timestep
+        move_type_probabilities = (
```
rmudambi, line 87: Don't rename columns here. There's no reason why `self.person_move_rates()` shouldn't return a DataFrame which already has the correct column names. Either fix the backing csv or set the column names properly in `setup` before setting it as the backing data of the `LookupTable`. 

The residual choice column should also be created in setup

rmudambi (changes requested):
Mostly looks good. One comment about doing some things on each time-step that should be done only once in the setup phase.

src/vivarium_census_prl_synth_pop/components/person.py
```diff
@@ -88,30 +82,23 @@ def on_time_step(self, event: Event) -> None:
         """
 
         persons = self.population_view.get(event.index)
```
albrja, line 84: Can you change this to just pop or population?  The naming is old and that keeps it clear and consistent across components.  I'm updating this in the other components.

albrja (approved):
Looks good with a nit.

rmudambi (approved):
Approved pending the changes we talked about.

PR #97, opened 2022-12-16
Feature/sbachmei/mic 3669 household survey observer implementation

stevebachmeier:
### Description
- *Category*: implementation
- *JIRA issue*: [MIC-3669](https://jira.ihme.washington.edu/browse/MIC-3669)
- *Research reference*: https://vivarium-research.readthedocs.io/en/latest/models/concept_models/vivarium_census_synthdata/concept_model.html#household-surveys

### Changes and notes
Adds house hold survey observer using the base observer class. Note that this
does NOT implement non-responses yet; that will be done when the RT
has finalized the approach (and might be implemented in pseudo-people
anyway)

I also added a few pytests, though we should probably discuss coverage requirements
and create a separate ticket for that.

### Verification and Testing
* ran a short sim and ensured datasets are being generated
* pytests are passing

### DIFF
--- a/src/vivarium_census_prl_synth_pop/components/observers.py
+++ b/src/vivarium_census_prl_synth_pop/components/observers.py
@@ -5,8 +5,8 @@ from vivarium.framework.engine import Builder
 from vivarium.framework.event import Event
 from vivarium.framework.population import PopulationView
 
-from vivarium_census_prl_synth_pop.constants import metadata
 from vivarium_census_prl_synth_pop import utilities
+from vivarium_census_prl_synth_pop.constants import metadata
 
 
 class BaseObserver(ABC):
@@ -41,7 +41,7 @@ class BaseObserver(ABC):
         # FIXME: settle on output dirs
         self.output_dir = utilities.build_output_dir(builder, subdir="results")
         self.population_view = self.get_population_view(builder)
-        self.responses = self.get_responses()
+        self.responses = self.get_response_schema()
 
         # Register the listener to update the responses
         builder.event.register_listener(
@@ -61,7 +61,7 @@ class BaseObserver(ABC):
         pass
 
     @abstractmethod
-    def get_responses(self) -> pd.DataFrame:
+    def get_response_schema(self) -> pd.DataFrame:
         """Initializes the observation/results data structure and schema"""
         pass
 
@@ -127,8 +127,8 @@ class HouseholdSurveyObserver(BaseObserver):
         "puma",
         "guardian_1",  # For noise functions
         "guardian_2",  # For noise functions
-        "guardian_1_address_id",  # For noise functions 
-        "guardian_2_address_id",  # For noise functions 
+        "guardian_1_address_id",  # For noise functions
+        "guardian_2_address_id",  # For noise functions
     ]
     SAMPLING_RATE = {
         "ACS": 12,  # FIXME: 12000
@@ -175,7 +175,7 @@ class HouseholdSurveyObserver(BaseObserver):
         population_view = builder.population.get_view(columns=cols)
         return population_view
 
-    def get_responses(self) -> pd.DataFrame:
+    def get_response_schema(self) -> pd.DataFrame:
         """Returns the response schema"""
         cols = HouseholdSurveyObserver.OUTPUT_COLS
         return pd.DataFrame(columns=cols)
@@ -186,25 +186,24 @@ class HouseholdSurveyObserver(BaseObserver):
 
     def do_observation(self, event) -> None:
         """Records the survey responses on this time step."""
-        new_responses = self.population_view.get(
-            event.index, query='alive == "alive"'
-        )
+        new_responses = self.population_view.get(event.index, query='alive == "alive"')
         respondent_households = utilities.vectorized_choice(
             options=list(set(new_responses["household_id"])),
             n_to_choose=self.samples_per_timestep,
             randomness_stream=self.randomness,
             additional_key="sampling_households",
         )
-
-        breakpoint()
-        new_responses = new_responses[new_responses["household_id"].isin(respondent_households)]
+        new_responses = new_responses[
+            new_responses["household_id"].isin(respondent_households)
+        ]
         new_responses["survey"] = self.survey
         new_responses["survey_date"] = event.time.date()
-        new_responses["middle_initial"] = new_responses["middle_name"].str[0]
+        new_responses = utilities.convert_middle_name_to_initial(new_responses)
         new_responses = utilities.add_guardian_address_ids(new_responses)
         # Apply column schema and concatenate
         new_responses = new_responses[self.responses.columns]
         self.responses = pd.concat([self.responses, new_responses])
+
 
 class DecennialCensusObserver(BaseObserver):
     """Class for observing columns relevant to a decennial census on April
@@ -219,7 +218,7 @@ class DecennialCensusObserver(BaseObserver):
     @property
     def name(self):
         return f"decennial_census_observer"
-    
+
     @property
     def output_filename(self):
         return f"decennial_census.hdf"
@@ -228,30 +227,34 @@ class DecennialCensusObserver(BaseObserver):
         super().setup(builder)
         self.clock = builder.time.clock()
         self.time_step = builder.configuration.time.step_size  # in days
-        assert self.time_step <= 30, 'DecennialCensusObserver requires model specification configuration with time.step_size <= 30'
-        
+        assert (
+            self.time_step <= 30
+        ), "DecennialCensusObserver requires model specification configuration with time.step_size <= 30"
+
     def get_population_view(self, builder) -> PopulationView:
         """Get the population view to be used for observations"""
         return builder.population.get_view(columns=metadata.DECENNIAL_CENSUS_COLUMNS_USED)
 
-    def get_responses(self) -> pd.DataFrame:
-        return pd.DataFrame(columns=[
-            "first_name",
-            "middle_initial",
-            "last_name",
-            "age",
-            "date_of_birth",
-            "address_id",
-            "relation_to_household_head",
-            "sex",
-            "race_ethnicity",
-            "census_year",
-            "guardian_1",
-            "guardian_1_address_id",
-            "guardian_2",
-            "guardian_2_address_id",
-            "housing_type",
-        ])
+    def get_response_schema(self) -> pd.DataFrame:
+        return pd.DataFrame(
+            columns=[
+                "first_name",
+                "middle_initial",
+                "last_name",
+                "age",
+                "date_of_birth",
+                "address_id",
+                "relation_to_household_head",
+                "sex",
+                "race_ethnicity",
+                "census_year",
+                "guardian_1",
+                "guardian_1_address_id",
+                "guardian_2",
+                "guardian_2_address_id",
+                "housing_type",
+            ]
+        )
 
     def to_observe(self, event: Event) -> bool:
         """Note: this method uses self.clock instead of event.time to handle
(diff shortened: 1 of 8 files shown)


tests/components/observers/test_household_survey_observers.py
```diff
@@ -0,0 +1,63 @@
...
+
+
+@pytest.fixture
+def mocked_pop_view():
+    """Generate a state table view"""
+    cols = HouseholdSurveyObserver.INPUT_COLS
+    df = pd.DataFrame(np.random.randint(0, 100, size=(2, len(cols))), columns=cols)
+    df["alive"] = "alive"
+    df[["first_name", "middle_name", "last_name"]] = (
+        ["Alex", "J", "Honnold"],
```
rmudambi, line 33: Haha love the names

tests/components/observers/test_household_survey_observers.py
```diff
@@ -0,0 +1,63 @@
...
+    event = mocker.MagicMock()
+    event.time = pd.to_datetime("2021-01-01")
+    event.index = pd.Index([0, 1])
+    observer.population_view.get.return_value = mocked_pop_view  # FIXME: This returns mocked_pop_view regardless of event.index or alive status
+    mocker.patch(
+        "vivarium_census_prl_synth_pop.components.observers.utilities.vectorized_choice",
+        return_value=list(mocked_pop_view["household_id"]),
+    )
+    observer.do_observation(event)
+    assert set(observer.responses.columns) == set(HouseholdSurveyObserver.OUTPUT_COLS)
```
rmudambi, line 63: Could add a test that the results actually get written to the desired location at simulation end.

src/vivarium_census_prl_synth_pop/components/observers.py
```diff
@@ -85,8 +85,124 @@ def do_observation(self, event: Event) -> None:
...
+        "address_id",
+        "state",
+        "puma",
+        "guardian_1",  # For noise functions
+        "guardian_2",  # For noise functions
+        "guardian_1_address_id",  # For noise functions
+        "guardian_2_address_id",  # For noise functions
+    ]
+    SAMPLING_RATE = {
+        "ACS": 12,  # FIXME: 12000
```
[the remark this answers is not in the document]
stevebachmeier: I'll ask the RT about
1. can we scale based on population (eg 12000 * 2000 / 330 million)
2. What exactly _is_ the US population?

src/vivarium_census_prl_synth_pop/utilities.py
```diff
@@ -376,3 +376,23 @@ def update_address_id_for_unit_and_sims(
...
+        ]  # is it faster to remove the negative values?
+        pop[f"guardian_{i}_address_id"] = s_guardian_id.map(pop["address_id"])
+    return pop
+
+
+def convert_middle_name_to_initial(pop: pd.DataFrame) -> pd.DataFrame:
+    """Converts middle names to middle initials. Note that this drops
+    the 'middle_name' column altogether and replaces it with 'middle_initial'
+    """
+    pop["middle_initial"] = pop["middle_name"].str[0]
```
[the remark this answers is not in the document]
stevebachmeier: I confirmed that everyone does, but this approach also works w/ nans (which is why I used this instead of `.astype(str)[0]`)

```
>>> df = pd.DataFrame({'x':["David", np.nan]})
>>> df.x.str[0]
0      D
1    NaN
Name: x, dtype: object
>>> 
```

tests/components/observers/test_household_survey_observers.py
stevebachmeier: Interestingly, Honnold actually does not have a middle name. But in some interview he claimed his middle name is "J".

src/vivarium_census_prl_synth_pop/components/observers.py
```diff
@@ -85,8 +85,124 @@ def do_observation(self, event: Event) -> None:
...
+        "CPS": 60,  # FIXME: 60000
+    }
```
albrja, line 136: We have a lot of components that require different constant list, maps, etc and we also talked about having a list for all the state table columns.  I think we should think about adding something to data values and/or metadata similar to ata keys but for constants to help organize this.  Obviously not for this PR.

src/vivarium_census_prl_synth_pop/components/observers.py
```diff
@@ -85,8 +85,124 @@ def do_observation(self, event: Event) -> None:
...
+        return f"{self.survey}.hdf"
+
+    #################
+    # Setup methods #
+    #################
+
+    def setup(self, builder: Builder):
+        super().setup(builder)
+        self.randomness = builder.randomness.get_stream(self.name)
+        # FIXME: Should we adjust to get exactly the right amount after 1 year?
```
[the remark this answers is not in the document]
stevebachmeier: I've asked the RT

src/vivarium_census_prl_synth_pop/components/observers.py
stevebachmeier: Well this is unique b/c it's very specific to this particular observer and these values will likely not be used outside of this spot. If a value is used in more than one place, then I def agree there should be a centralized location of constants (column names are the big obvious here). It's less useful and perhaps even worse if you only use it once.

src/vivarium_census_prl_synth_pop/components/observers.py
```diff
@@ -72,7 +72,7 @@ def get_responses(self) -> pd.DataFrame:
     def on_collect_metrics(self, event: Event) -> None:
         if self.to_observe(event):
             self.do_observation(event)
```
rmudambi, line 74: I just thought of something. We can enforce that we maintain the output schema here and throw an exception if `self.responses` doesn't match the schema after `do_observation` is called.

src/vivarium_census_prl_synth_pop/utilities.py
```diff
@@ -376,3 +376,23 @@ def update_address_id_for_unit_and_sims(
         pop.loc[rows_changing_address_id_idx, address_id_col_name] = updated_address_ids
 
     return pop, moving_units, total_address_id_count
+
+
+def add_guardian_address_ids(pop: pd.DataFrame) -> pd.DataFrame:
+    """Map the address ids of guardians to each simulant's guardian address columns"""
+    for i in [1, 2]:
+        s_guardian_id = pop[f"guardian_{i}"].dropna()
```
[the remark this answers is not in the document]
albrja: Are we seeing NaNs here?  This still seems unnecessary to me.

src/vivarium_census_prl_synth_pop/components/observers.py
stevebachmeier: Hmm, good idea. I've started thinking a little bit about how we might want to enforce schema throughout this project but this is a good intermediate

albrja (approved):
Nice pytests.

albrja (approved):
This looks good.  For the tests, it seems like you will have a lot of duplicated code as you add more observers.  I know you can't really adequately test the abstract base class but could you provide additional args to some of these tests to test different observers?  For example, verifying the correct columns are there.  Curious as a pytest noob.

src/vivarium_census_prl_synth_pop/components/observers.py
```diff
@@ -143,25 +277,20 @@ def to_observe(self, event: Event) -> bool:
         this function is 2020-04-29 (because the time.step_size is 28
         days)
         """
-        return ((self.clock().year % 10 == 0)  # decennial year
-                and (self.clock().month == 4)  # month of April
-                and (self.clock().day <= self.time_step)  # time step containing first day of month
-               )
+        return (
+            (self.clock().year % 10 == 0)  # decennial year
```
rmudambi, line 281: Note: we could clean up this logic (and support time-steps greater than 30 days with the following:
```
census_year = 10 * (self.event.year // 10)   # integer division is floor
census_date = datetime(census__year, 4, 1)
return self.clock() <= census_date < event.time
```

tests/components/observers/test_household_survey_observers.py
```diff
@@ -0,0 +1,80 @@
...
+
+def test_instantiate(observer):
+    assert str(observer) == "HouseholdSurveyObserver(ACS)"
+
+
+def test_responses_schema(observer):
+    """Is the initial self.responses (after setup) the expected schema
+    (pd.DataFrame with correct columns)?
+    """
+    pd.testing.assert_frame_equal(
```
rmudambi, line 42: I didn't know about this function

rmudambi (approved):

src/vivarium_census_prl_synth_pop/utilities.py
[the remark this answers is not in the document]
stevebachmeier: Will remove. You know that there will never be nans at this point?

src/vivarium_census_prl_synth_pop/constants/data_values.py
```diff
@@ -1,9 +1,15 @@
 from typing import NamedTuple
 
+
+DAYS_PER_YEAR = 365.25
+
```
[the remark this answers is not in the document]
stevebachmeier: I couldn't find one in vivarium, though it does exist in vph. @rmudambi is it better practice to import from vpy just for a constant like this or define here?

stevebachmeier:
> This looks good. For the tests, it seems like you will have a lot of duplicated code as you add more observers. I know you can't really adequately test the abstract base class but could you provide additional args to some of these tests to test different observers? For example, verifying the correct columns are there. Curious as a pytest noob.

Yes, we'll need to come up with a pytest strategy in general I think. I'm also not sure offhand the best way to reduce copy/paste jobs as we add more and more observers that should be tested in similar ways. Perhaps there is a way to test the abc I'm just not aware of. Or if you understand what the abc exactly does, then you don't need to test the functionality of it beyond a single arbitrary observer.

tests/components/observers/test_household_survey_observers.py
stevebachmeier: it's a good one and I use it all the time

src/vivarium_census_prl_synth_pop/model_specifications/model_spec.yaml
```diff
@@ -29,8 +32,8 @@ configuration:
             month: 4
             day: 1
         end:
-            year: 2030
-            month: 5
+            year: 2020
+            month: 6
```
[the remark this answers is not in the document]
stevebachmeier: Oh weird, idk - my dates had always been 6/2020. They must have been merged into main on accident.

src/vivarium_census_prl_synth_pop/components/observers.py
```diff
@@ -85,8 +104,120 @@ def do_observation(self, event: Event) -> None:
...
+    @property
+    def name(self):
+        return f"household_survey_observer.{self.survey}"
+
+    @property
+    def output_filename(self):
+        return f"{self.survey}.hdf"
+
+    @property
+    def input_columns(self):
```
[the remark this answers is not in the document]
stevebachmeier: @rmudambi Don't we want to throw an error? Just to kind of force the developer to think about what columns are needed?

src/vivarium_census_prl_synth_pop/components/observers.py
stevebachmeier: Clever

src/vivarium_census_prl_synth_pop/components/observers.py
```diff
@@ -266,24 +280,14 @@ def setup(self, builder: Builder):
         super().setup(builder)
         self.clock = builder.time.clock()
         self.time_step = builder.configuration.time.step_size  # in days
-        assert (
-            self.time_step <= 30
-        ), "DecennialCensusObserver requires model specification configuration with time.step_size <= 30"
+        if self.time_step > 30:
```
rmudambi, line 283: We can remove this error now since `to_observe` now works with longer timesteps.

PR #99, opened 2022-12-16
Bugfix/guardians missing

albrja:
### Handles edge case in guardian assignment where the desired guardian type for assignment has no choices for the randomness stream.
- *Category*: Bufix
- *JIRA issue*: [MIC-3689](https://jira.ihme.washington.edu/browse/MIC-3689)
- *Research reference*: N/A

### Changes and notes
-Adds check to see if we are trying to assign college sims to an empty series of guardian types for a particular race/ethnicity.

### Verification and Testing
Ran simulation.

### DIFF
--- a/src/vivarium_census_prl_synth_pop/components/population.py
+++ b/src/vivarium_census_prl_synth_pop/components/population.py
@@ -752,11 +752,27 @@ class Population:
                     households_guardian_types[guardian_type]["race_ethnicity"] == race,
                     "reference_person_id",
                 ]
-                race_guardian_ids = self.randomness.choice(
-                    race_college_sims_with_guardian_idx,
-                    choices=race_reference_person_ids,
-                    additional_key=f"{race}_{guardian_type}_guardian_ids",
-                )
+                if race_reference_person_ids.empty & len(race_college_sims_with_guardian_idx) > 0:
+                    # Get ids of for all other race reference persons if guardian type has no reference persons
+                    # This handles an unlikely edge case
+                    race_reference_person_ids = pd.concat(
+                        [df for gt, df in households_guardian_types.items() if gt != guardian_type]
+                    )
+                    race_reference_person_ids = race_reference_person_ids.loc[
+                        race_reference_person_ids["race_ethnicity"] == race,
+                        "reference_person_id",
+                    ]
+                    race_guardian_ids = self.randomness.choice(
+                        race_college_sims_with_guardian_idx,
+                        choices=race_reference_person_ids,
+                        additional_key=f"{race}_combined_guardian_ids",
+                    )
+                else:
+                    race_guardian_ids = self.randomness.choice(
+                        race_college_sims_with_guardian_idx,
+                        choices=race_reference_person_ids,
+                        additional_key=f"{race}_{guardian_type}_guardian_ids",
+                    )
 
                 pop.loc[race_guardian_ids.index, "guardian_1"] = race_guardian_ids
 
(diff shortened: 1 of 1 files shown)


src/vivarium_census_prl_synth_pop/components/population.py
```diff
@@ -752,11 +752,34 @@ def assign_college_simulants_guaridans(self, pop: pd.DataFrame) -> pd.DataFrame:
                     households_guardian_types[guardian_type]["race_ethnicity"] == race,
                     "reference_person_id",
                 ]
-                race_guardian_ids = self.randomness.choice(
-                    race_college_sims_with_guardian_idx,
-                    choices=race_reference_person_ids,
-                    additional_key=f"{race}_{guardian_type}_guardian_ids",
-                )
+                if (
+                    race_reference_person_ids.empty & len(race_college_sims_with_guardian_idx)
```
stevebachmeier, line 756: use `and` here for boolean operator

src/vivarium_census_prl_synth_pop/components/population.py
```diff
@@ -752,11 +752,34 @@ def assign_college_simulants_guaridans(self, pop: pd.DataFrame) -> pd.DataFrame:
...
+                            df
+                            for gt, df in households_guardian_types.items()
+                            if gt != guardian_type
+                        ]
+                    )
+                    race_reference_person_ids = race_reference_person_ids.loc[
+                        race_reference_person_ids["race_ethnicity"] == race,
+                        "reference_person_id",
+                    ]
+                    race_guardian_ids = self.randomness.choice(
```
stevebachmeier, line 772: Consider removing this chunk from the if statement and instead generating the additional_key. Then in the else block below just assign the different additional_key.

Then you need this chunk only once outside of the if/else

src/vivarium_census_prl_synth_pop/components/population.py
```diff
@@ -752,11 +752,34 @@ def assign_college_simulants_guaridans(self, pop: pd.DataFrame) -> pd.DataFrame:
...
-                race_guardian_ids = self.randomness.choice(
-                    race_college_sims_with_guardian_idx,
-                    choices=race_reference_person_ids,
-                    additional_key=f"{race}_{guardian_type}_guardian_ids",
-                )
+                if (
+                    race_reference_person_ids.empty & len(race_college_sims_with_guardian_idx)
+                    > 0
+                ):
+                    # Get ids of for all other race reference persons if guardian type has no reference persons
```
stevebachmeier, line 759: This comment is confusing; I was interpreting "for all other race reference persons" as somehow meaning you were trying to assign guardians of a different race ("all other race") which is exactly incorrect.

stevebachmeier (changes requested):
Mostly small things except for the operator

src/vivarium_census_prl_synth_pop/components/population.py
albrja: I've changed it but does this make a difference functionally?

src/vivarium_census_prl_synth_pop/components/population.py
albrja: Ok this makes it a lot cleaner and removes the duplicated chunk of code and I just left the additional key as guardian_type since that was more clear to me anyways.

src/vivarium_census_prl_synth_pop/components/population.py
stevebachmeier: Full disclosure - I still struggle with when to use bitwise operations (&). For this particular case, no, the "answer" is the same. But it's just not really the correct approach since you're dealing with boolean logic here.

stevebachmeier (approved):

PR #101, opened 2022-12-17
Implement logic for new-household moves

zmbc:
### Description

- *Category*: implementation
- *JIRA issue*: [MIC-3673](https://jira.ihme.washington.edu/browse/MIC-3673)
- *Research reference*: https://vivarium-research.readthedocs.io/en/latest/models/concept_models/vivarium_census_synthdata/concept_model.html#new-household-moves

### Verification and Testing
Ran simulation without error; stepped through and inspected newly generated household IDs and address IDs; checked that household ID -> address ID was still a 1-to-1 mapping.

### DIFF
--- a/src/vivarium_census_prl_synth_pop/components/person.py
+++ b/src/vivarium_census_prl_synth_pop/components/person.py
@@ -1,3 +1,4 @@
+import numpy as np
 import pandas as pd
 from vivarium.framework.engine import Builder
 from vivarium.framework.event import Event
@@ -6,7 +7,7 @@ from vivarium.framework.utilities import from_yearly
 from vivarium.framework.values import Pipeline
 
 from vivarium_census_prl_synth_pop.constants import data_values, paths
-from vivarium_census_prl_synth_pop.utilities import filter_by_rate
+from vivarium_census_prl_synth_pop.utilities import update_address_id
 
 
 class PersonMigration:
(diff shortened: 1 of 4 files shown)


src/vivarium_census_prl_synth_pop/components/person.py
```diff
@@ -98,7 +97,12 @@ def on_time_step(self, event: Event) -> None:
             p=move_type_probabilities.values,
         )
 
-        domestic_movers_idx = pop.index[move_types_chosen != "no_move"]
+        pop = self._perform_new_household_moves(
```
rmudambi, line 100: Conventionally `pop` means the "whole" pop table. I'd suggest storing the result of this call in a new variable.for clarity.

rmudambi (approved):

src/vivarium_census_prl_synth_pop/components/person.py
zmbc: This is the whole pop table, unless I am misunderstanding what you mean by "whole". Would you prefer that this method returned only the movers?

albrja (approved):

src/vivarium_census_prl_synth_pop/components/person.py
rmudambi: Sorry, disregard my comment. I must have misread what is happening in the `perform_hew_household_Moves` method.

PR #102, opened 2022-12-17
Fix CI

zmbc:
### Description
- *Category*: chore
- *JIRA issue*: [MIC-3660](https://jira.ihme.washington.edu/browse/MIC-3660)
- *Research reference*: none

### Changes and notes

Probably everyone just turns off their email notifications, but this was slightly bothering me. Also, the fact that black and isort hadn't run in a while meant I kept getting extraneous changes in my other branches when I ran them.

The only slightly controversial change here is a very small amount of code duplication [from vivarium_cluster_tools](https://github.com/ihmeuw/vivarium_cluster_tools/blob/main/src/vivarium_cluster_tools/utilities.py#L35-L39) to avoid needing that at runtime of the sim. I think this would need to be done anyway if we want the archival version to be runnable by an external person without installing vivarium_cluster_tools.

### Verification and Testing

Made sure simulation still runs to completion, including saving results. Tests are passing, obviously 😄

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -7,11 +7,10 @@ on: [push, pull_request, workflow_dispatch]
 
 jobs:
   build:
-    # NOTE: If we want to test with 3.6, we cannot run on ubuntu-latest anymore
-    runs-on: ubuntu-20.04
+    runs-on: ubuntu-latest
     strategy:
       matrix:
-        python-version: [3.6, 3.7, 3.8]
+        python-version: [3.7, 3.8]
     defaults:
       run:
         shell: bash -l {0}
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -81,6 +81,6 @@ jobs:
 
       - name: Lint
         run: |
-          pip install black==22.1.0 isort
+          pip install black==22.3.0 isort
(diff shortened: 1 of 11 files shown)


.github/workflows/build.yml
```diff
@@ -7,7 +7,8 @@ on: [push, pull_request, workflow_dispatch]
 
 jobs:
   build:
-    runs-on: ubuntu-latest
+    # NOTE: If we want to test with 3.6, we cannot run on ubuntu-latest anymore
+    runs-on: ubuntu-20.04
```
rmudambi, line 11: Good to know. I think we probably want to stay on `ubuntu-latest`, so instead I'd suggest removing 3.6 from the strategy matrix

rmudambi (approved):

rmudambi:
There is actually already a jira ticket for this which was planned for the upcoming sprint: https://jira.ihme.washington.edu/browse/MIC-3660. I've assigned it to you and moved into ready for review

.github/workflows/build.yml
zmbc: Based on our conversation earlier, I've reversed this: sticking with `ubuntu-latest`, dropping Python 3.6.

PR #105, opened 2022-12-19
Bugfix/perturb age mask

albrja:
### Fixes bug caused by boolean mask not having same index as population table.
- *Category*: Bugfix
- *JIRA issue*: [MIC-3691](https://jira.ihme.washington.edu/browse/MIC-3691)
- *Research reference*: N/A

### Changes and notes
-Changes implementation in function to use indexes instead of masks

### Verification and Testing
Successfully ran simulation for population sizes of 2,000, 20,000, and 200,000.

### DIFF
--- a/src/vivarium_census_prl_synth_pop/components/population.py
+++ b/src/vivarium_census_prl_synth_pop/components/population.py
@@ -91,7 +91,9 @@ class Population:
 
     def _load_population_data(self, builder: Builder):
         households = builder.data.load(data_keys.POPULATION.HOUSEHOLDS)
-        persons = builder.data.load(data_keys.POPULATION.PERSONS)[metadata.PERSONS_COLUMNS_TO_INITIALIZE]
+        persons = builder.data.load(data_keys.POPULATION.PERSONS)[
+            metadata.PERSONS_COLUMNS_TO_INITIALIZE
+        ]
         return {"households": households, "persons": persons}
 
     def initialize_simulants(self, pop_data: SimulantData) -> None:
--- a/src/vivarium_census_prl_synth_pop/components/population.py
+++ b/src/vivarium_census_prl_synth_pop/components/population.py
@@ -895,13 +895,13 @@ class Population:
         # todo: Consider changing this function to build a distribution for each simulan instead of resampling.
 
         # Get age shift and redraw for simulants who get negative ages
-        to_shift = pd.Series(True, index=pop.index)
+        to_shift = pd.Index(pop.index)
         max_iterations = 10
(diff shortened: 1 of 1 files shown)


src/vivarium_census_prl_synth_pop/components/population.py
```diff
@@ -895,13 +897,13 @@ def perturb_individual_age(self, pop: pd.DataFrame) -> pd.Series:
         # todo: Consider changing this function to build a distribution for each simulan instead of resampling.
 
         # Get age shift and redraw for simulants who get negative ages
-        to_shift = pd.Series(True, index=pop.index)
+        to_shift = pd.Index(pop.index)
```
stevebachmeier, line 900: No need for pd.Index(). `pop.index` is already a pandas index by definition

src/vivarium_census_prl_synth_pop/components/population.py
```diff
@@ -912,17 +914,17 @@ def perturb_individual_age(self, pop: pd.DataFrame) -> pd.Series:
 
             # Calculate shifted ages and see if any are negative and need to be resampled.
             shifted_ages = pop.loc[to_shift, "age"] + age_shift
-            negative_ages_mask = shifted_ages < 0
-            pop.loc[~negative_ages_mask & to_shift, "age"] = shifted_ages.loc[
-                ~negative_ages_mask & to_shift
-            ]
-            to_shift = negative_ages_mask
+            positive_ages_idx = shifted_ages.loc[shifted_ages >= 0].index
```
stevebachmeier, line 917: This is better!

Hah, kinda a nit that you call it `positive_ages_idx` but they include 0 (ie, not strictly positive values)...but i also can't think of a better name.

stevebachmeier (approved):

src/vivarium_census_prl_synth_pop/components/population.py
zmbc, line 917: I think this can just be `shifted_ages.index[shifted_ages >= 0]`.

src/vivarium_census_prl_synth_pop/components/population.py
```diff
@@ -912,17 +914,17 @@ def perturb_individual_age(self, pop: pd.DataFrame) -> pd.Series:
...
+            pop.loc[positive_ages_idx, "age"] = shifted_ages.loc[positive_ages_idx]
+            to_shift = shifted_ages.loc[
```
zmbc, line 919: I think this can just be `to_shift = shifted_ages.index.difference(positive_ages_idx)`

PR #109, opened 2022-12-20
Implement logic for non-reference-person moves

zmbc:
### Description

- *Category*: implementation
- *JIRA issue*: [MIC-3675](https://jira.ihme.washington.edu/browse/MIC-3675)
- *Research reference*: https://vivarium-research.readthedocs.io/en/latest/models/concept_models/vivarium_census_synthdata/concept_model.html#non-reference-person-moves

### Changes and notes

No changes from the ticket description.

### Verification and Testing

Ran simulation, stepped through process of assigning to households.

### DIFF
--- a/src/vivarium_census_prl_synth_pop/components/person.py
+++ b/src/vivarium_census_prl_synth_pop/components/person.py
@@ -13,12 +13,16 @@ from vivarium_census_prl_synth_pop.utilities import update_address_id
 
 class PersonMigration:
     """
-    - needs to be able to update a household_id (to an existing id)
-    - needs to be able to update a relationship to head of household (to other nonrelative)
-    - on_time_step, needs to be able to look up probability of moving and determine if moving
-    - needs to be able to choose a new household at random
-    ASSUMPTIONS:
-    - head of household never moves to new household_id
+    Handles domestic (within the US) migration of individuals (not in household groups).
+
+    There are three types of individual moves in domestic migration:
+    - New-household moves, where the individual establishes a new single-person household.
+    - GQ person moves, where the individual moves into a GQ type.
+    - Non-reference-person moves, where the individual joins an existing household.
+
+    Note that the names of these move types refer to the living situation *after* the move
+    is complete.
+    The population at risk for all three move types is all simulants.
     """
 
     def __repr__(self) -> str:
--- a/src/vivarium_census_prl_synth_pop/components/person.py
+++ b/src/vivarium_census_prl_synth_pop/components/person.py
@@ -1,5 +1,6 @@
 import numpy as np
 import pandas as pd
+from loguru import logger
 from vivarium.framework.engine import Builder
 from vivarium.framework.event import Event
 from vivarium.framework.randomness import RESIDUAL_CHOICE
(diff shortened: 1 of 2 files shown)


src/vivarium_census_prl_synth_pop/components/person.py
```diff
@@ -253,25 +185,47 @@ def _perform_gq_person_moves(self, pop: pd.DataFrame, movers: pd.Index) -> pd.Da
...
+
+        return pop
+
+    def _get_address_ids_by_household_id(self, df: pd.DataFrame) -> pd.Series:
+        result = (
+            df[["household_id", "address_id"]]
+            .drop_duplicates()
+            .set_index("household_id")["address_id"]
+        )
+        assert result.index.is_unique
```
rmudambi, line 230: No need for this assert. You have `drop_duplicates`. If we care enough about this to test it, we should write a test for it.

src/vivarium_census_prl_synth_pop/components/person.py
```diff
@@ -253,25 +185,47 @@ def _perform_gq_person_moves(self, pop: pd.DataFrame, movers: pd.Index) -> pd.Da
...
 
-        return pd.Series(new_household_ids)
+        pop.loc[movers, "household_id"] = household_id_values
+        pop.loc[movers, "housing_type"] = "Standard"
+        pop.loc[movers, "address_id"] = household_id_values.map(non_gq_address_ids)
+        pop.loc[movers, "relation_to_household_head"] = "Other nonrelative"
+
+        return pop
+
+    def _get_address_ids_by_household_id(self, df: pd.DataFrame) -> pd.Series:
```
rmudambi, line 224: This method name isn't quite right. You're not providing a list of household ids - you're providing a subset of the pop table. I'd suggest something like `_get_household_addresses`

src/vivarium_census_prl_synth_pop/components/person.py
```diff
@@ -253,25 +185,47 @@ def _perform_gq_person_moves(self, pop: pd.DataFrame, movers: pd.Index) -> pd.Da
...
+        household_id_values = pop.loc[movers, "household_id"]
         additional_seed = 0
-        while (new_household_ids == pop.loc[sims_who_move, "household_id"]).any():
-            unchanged_households = new_household_ids == pop.loc[sims_who_move, "household_id"]
-            new_household_ids[unchanged_households] = self.randomness.choice(
-                new_household_ids.loc[unchanged_households].index,
-                all_household_ids,
-                additional_key=f"newHousehold_ids_{additional_seed}",
+
+        while not to_move.empty:
```
rmudambi, line 208: Use a `for` loop with a maximum iterations to avoid the potential of an infinite loop.

rmudambi (approved):
Looks good once you address the comments.

src/vivarium_census_prl_synth_pop/components/person.py
zmbc: `drop_duplicates` does not ensure per-column uniqueness (in this case "household_id") because it only drops duplicate _combinations_.

It makes sense why having an `assert` here is not right, but note that this component will error anyway (when it tries to use this mapping) if the index is not unique.

src/vivarium_census_prl_synth_pop/components/person.py
zmbc: I think the result is best thought of as a mapping from household_id -> address_id. I will rename accordingly.

PR #113, opened 2022-12-21
Feature/name removal

albrja:
### Removes names from simulation and replaces them with name id columns.
- *Category*: 
- *JIRA issue*: [MIC-3658](https://jira.ihme.washington.edu/browse/MIC-3658)
- *Research reference*: N/A

### Changes and notes
-Removes name columns from state table and replaces them with name_id columns for first, middle, and last name.
-Updates column names in observers.

### Verification and Testing
Ran simulation for 250,000 and saw improvement in performance.

### DIFF
--- a/src/vivarium_census_prl_synth_pop/components/population.py
+++ b/src/vivarium_census_prl_synth_pop/components/population.py
@@ -10,10 +10,6 @@ from vivarium.framework.population import SimulantData
 from vivarium.framework.time import get_time_stamp
 from vivarium_public_health.utilities import DAYS_PER_YEAR, to_years
 
-from vivarium_census_prl_synth_pop.components.synthetic_pii import (
-    NameGenerator,
-    SSNGenerator,
-)
 from vivarium_census_prl_synth_pop.constants import data_keys, data_values, metadata
 from vivarium_census_prl_synth_pop.utilities import vectorized_choice
 
@@ -31,7 +27,6 @@ PARTNERS = [
 
 
 class Population:
-
     @property
     def name(self):
         return "population"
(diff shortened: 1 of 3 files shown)


src/vivarium_census_prl_synth_pop/components/population.py
```diff
@@ -126,10 +114,11 @@ def generate_initial_population(self, pop_data: SimulantData) -> None:
         # drop non-unique household_id
         pop = pop.drop(columns="census_household_id")
 
-        # give names
-        first_and_middle = self.name_generator.generate_first_and_middle_names(pop)
-        last_names = self.name_generator.generate_last_names(pop)
-        pop = pd.concat([pop, first_and_middle, last_names], axis=1)
+        # give name ids
+        pop["first_name_id"] = pop.index
```
stevebachmeier, line 118: So what is the plan here? At what point do these become actual id numbers?

albrja: This is just going to be the id number.  We will handle in post-processing what these actually map to.  This takes away all generation and sampling for names out of the simulation.  So different ids may map to the same name.  We just want to make sure that if we know simulants will share a name that they will have the same name_id.  That is what the next ticket will implement for last name matching.

src/vivarium_census_prl_synth_pop/components/population.py
```diff
@@ -331,8 +320,8 @@ def initialize_newborns(self, pop_data: SimulantData) -> None:
         new_births["born_in_us"] = True
 
         # add first and middle names
-        names = self.name_generator.generate_first_and_middle_names(new_births)
-        new_births = pd.concat([new_births, names], axis=1)
+        new_births["first_name_id"] = new_births.index
```
rmudambi, line 323: Does this value increment from 0 or from the last index value in the pop table? If the latter, this looks fine

rmudambi (approved):

src/vivarium_census_prl_synth_pop/components/population.py
albrja: Last index in the pop table.

PR #114, opened 2022-12-21
Remove references to unused address component

zmbc:
### Description

- *Category*: bugfix
- *JIRA issue*: none
- *Research reference*: none

### Changes and notes

I noticed that because we were still referencing the Address component, the sim was still loading the address data every time. This only eliminates a constant from the runtime but it significantly improves quality of life for debugging with small populations!

### Verification and Testing

Ran simulation to completion; searched for references to the removed component.

On my setup (results may vary), with 2k population, before:

<pre><code>
2022-12-21 14:06:39.497 | vivarium.framework.engine:report:203 - 
              Event  Count  Mean time (s)  Std. dev. time (s)  Total time (s)  % Total time
     initialization      1           0.82                0.00            0.82          0.29
              <b>setup      1         149.84                0.00          149.84         53.74</b>
         post_setup      1           0.00                0.00            0.00          0.00
population_creation      1           2.69                0.00            2.69          0.96
 time_step__prepare    132           0.00                0.00            0.09          0.03
          time_step    132           0.92                0.08          120.94         43.38
 time_step__cleanup    132           0.02                0.00            2.65          0.95
    collect_metrics    132           0.01                0.03            1.22          0.44
     simulation_end      1           0.54                0.00            0.54          0.19
              total      1         278.80                0.00          278.80        100.00
</pre></code>

After:

<pre><code>
2022-12-21 14:11:35.644 | vivarium.framework.engine:report:203 - 
              Event  Count  Mean time (s)  Std. dev. time (s)  Total time (s)  % Total time
     initialization      1           1.42                0.00            1.42          0.97
              <b>setup      1          16.15                0.00           16.15         11.05</b>
         post_setup      1           0.00                0.00            0.00          0.00
population_creation      1           2.66                0.00            2.66          1.82
 time_step__prepare    132           0.00                0.00            0.09          0.06
          time_step    132           0.92                0.10          121.58         83.18
 time_step__cleanup    132           0.02                0.00            2.65          1.82
    collect_metrics    132           0.01                0.03            1.32          0.90
     simulation_end      1           0.30                0.00            0.30          0.20
              total      1         146.18                0.00          146.18        100.00
</pre></code>

PR #115, opened 2022-12-21
Use binary search for vectorized_choice

zmbc:
### Description

- *Category*: performance
- *JIRA issue*: [MIC-3569](https://jira.ihme.washington.edu/browse/MIC-3569)
- *Research reference*: none

### Changes and notes

Note that I have made this drop-in replacement without diving too deep into the behavior of vectorized_choice. In particular, I find its handling of Pandas indexes quite confusing.

### Verification and Testing

Stepped through and verified that the old code and new code would get the same results for actual inputs.

This change has a relatively small impact on runtime with small populations, because in small populations the timestep runtime dominates the total runtime -- I am guessing because there is some constant overhead to each timestep, while population creation scales more exactly with population size.

Performance for a 250k population, before:

<pre><code>
2022-12-21 14:59:05.745 | vivarium.framework.engine:report:203 - 
              Event  Count  Mean time (s)  Std. dev. time (s)  Total time (s)  % Total time
     initialization      1           0.55                0.00            0.55          0.05
              setup      1           7.63                0.00            7.63          0.76
         post_setup      1           0.00                0.00            0.00          0.00
<b>population_creation      1         134.24                0.00          134.24         13.39</b>
 time_step__prepare    132           0.03                0.00            3.62          0.36
          time_step    132           6.21                0.21          819.96         81.78
 time_step__cleanup    132           0.21                0.02           27.36          2.73
    collect_metrics    132           0.06                0.10            7.87          0.78
     simulation_end      1           1.48                0.00            1.48          0.15
              total      1        1002.69                0.00         1002.69        100.00
</pre></code>

After:

<pre><code>
2022-12-21 15:17:57.076 | vivarium.framework.engine:report:203 - 
              Event  Count  Mean time (s)  Std. dev. time (s)  Total time (s)  % Total time
     initialization      1           0.97                0.00            0.97          0.11
              setup      1           7.29                0.00            7.29          0.80
         post_setup      1           0.00                0.00            0.00          0.00
<b>population_creation      1          18.35                0.00           18.35          2.01</b>
 time_step__prepare    132           0.03                0.01            3.77          0.41
          time_step    132           6.42                0.24          847.21         92.61
 time_step__cleanup    132           0.21                0.01           28.05          3.07
    collect_metrics    132           0.06                0.10            7.78          0.85
     simulation_end      1           1.44                0.00            1.44          0.16
              total      1         914.86                0.00          914.86        100.00
</pre></code>

Note that both of these runs included the change in #114.

I expect the speedup to become even more dramatic as we further increase population size, and as we expand from Florida to the full US for our input data.

### DIFF
--- a/src/vivarium_census_prl_synth_pop/components/population.py
+++ b/src/vivarium_census_prl_synth_pop/components/population.py
@@ -194,18 +194,17 @@ class Population:
         return chosen_persons
 
     def choose_group_quarters(self, target_number_sims: int) -> pd.Series:
-        group_quarters = self.population_data["households"]["census_household_id"]
-        group_quarters = group_quarters.loc[
-            ["GQ" in household_id for household_id in group_quarters]
-        ]
+        group_quarters = self.population_data["households"].pipe(
+            lambda df: df[df["census_household_id"].str.contains("GQ")]
+        )
 
         # group quarters each house one person per census_household_id
         # they have NA household weights, but appropriately weighted person weights.
         chosen_units = vectorized_choice(
-            options=group_quarters,
+            options=group_quarters["census_household_id"],
             n_to_choose=target_number_sims,
             randomness_stream=self.randomness,
-            weights=self.population_data["households"]["person_weight"],
+            weights=group_quarters["person_weight"],
         )
 
         # get simulants per GQ unit
--- a/src/vivarium_census_prl_synth_pop/components/population.py
+++ b/src/vivarium_census_prl_synth_pop/components/population.py
@@ -205,7 +205,7 @@ class Population:
             options=group_quarters,
             n_to_choose=target_number_sims,
             randomness_stream=self.randomness,
-            weights=self.population_data["households"][["person_weight"]],
+            weights=self.population_data["households"]["person_weight"],
         )
 
         # get simulants per GQ unit
--- a/src/vivarium_census_prl_synth_pop/utilities.py
+++ b/src/vivarium_census_prl_synth_pop/utilities.py
@@ -214,7 +214,7 @@ def vectorized_choice(
     cdf = np.cumsum(pmf)
 
     # for each p_i in probs, count how many elements of cdf for which p_i >= cdf_i
-    chosen_indices = np.searchsorted(cdf[~pd.isnull(cdf)], probs, side="right")
+    chosen_indices = np.searchsorted(cdf, probs, side="right")
     return np.take(options, chosen_indices)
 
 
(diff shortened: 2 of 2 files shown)


src/vivarium_census_prl_synth_pop/utilities.py
```diff
@@ -214,8 +214,7 @@ def vectorized_choice(
     cdf = np.cumsum(pmf)
 
     # for each p_i in probs, count how many elements of cdf for which p_i >= cdf_i
-    vect_find_index = np.vectorize(lambda p_i: (p_i >= cdf).sum())
-    chosen_indices = vect_find_index(probs)
+    chosen_indices = np.searchsorted(cdf[~pd.isnull(cdf)], probs, side="right")
```
rmudambi, line 217: When would `cdf` be null?

rmudambi (approved):

src/vivarium_census_prl_synth_pop/components/population.py
```diff
@@ -205,7 +205,7 @@ def choose_group_quarters(self, target_number_sims: int) -> pd.Series:
             options=group_quarters,
             n_to_choose=target_number_sims,
             randomness_stream=self.randomness,
-            weights=self.population_data["households"][["person_weight"]],
+            weights=self.population_data["households"]["person_weight"],
```
albrja, line 208: Very curious how this wasn't breaking before since this would be a dataframe.

albrja (approved):

src/vivarium_census_prl_synth_pop/utilities.py
zmbc: Sometimes the `weights` has NaNs in it (but non-NaN values for all the indices present in `options`). But, I am realizing this is only true of one call to `vectorized_choice` so it is probably easy to fix that one in this PR and remove this condition.

More broadly, the way this function interfaces between numpy and pandas seems suspect to me -- I think it depends on the index of `options` being sorted, when `options` is a Series -- but working that out and refactoring seemed potentially complex and out of scope for this performance improvement.

Note that none of this is new -- there were always NaNs in `cdf`, it was just hidden "better" by `(p_i >= cdf).sum()`.

src/vivarium_census_prl_synth_pop/components/population.py
zmbc: To be honest, I am not sure. Something about using `vectorize` there.

zmbc: Or possibly `np.take` can transparently use a single-column dataframe, treating it like a 1d array.

zmbc: Ah, it was the first one: each call to the lambda inside `vectorize` returns a single-element pandas Series, but the vectorize function flattens each result into a simple integer data type (using the one value in the Series, and discarding the index) to return a numpy array.

PR #121, opened 2022-12-23
update ci, codeowners, readme

stevebachmeier:
### Description
- *Category*: other
- *JIRA issue*: [MIC-3690](https://jira.ihme.washington.edu/browse/MIC-3690)

This addresses a few issues with the github actions:
* Updates the python builds 3.7-3.10 (remove 3.6 and add 3.9-10)
* Update actions to stop future warnings (The `set-output` command
    is deprecated and will be disabled soon. Please upgrade to using
    Environment Files. For more information see: 
    https://github.blog/changelog/2022-10-11-github-actions-deprecating-save-state-and-set-output-commands/)
* Update CODEOWNERS
* Update README install 3.10
* Update setup to use min 3.7

### Testing
All tests passed and warnings are gone

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -81,5 +81,5 @@ jobs:
       - name: Lint
         run: |
           pip install black==22.3.0 isort
-          isort --profile black .
-          black --check --diff .
+          black . --check -v
+          isort . --check -v --profile black
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -81,5 +81,5 @@ jobs:
       - name: Lint
         run: |
           pip install black==22.3.0 isort
-          black . --check -v
-          isort . --check -v --profile black
+          black . --check --diff
+          isort . --check --verbose --only-modified --diff
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -82,4 +82,4 @@ jobs:
         run: |
           pip install black==22.3.0 isort
           black . --check --diff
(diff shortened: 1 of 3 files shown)


albrja (approved):

.github/workflows/build.yml
```diff
@@ -81,5 +81,5 @@ jobs:
       - name: Lint
         run: |
           pip install black==22.3.0 isort
-          black . --check -v
-          isort . --check -v
+          isort --profile black .
```
stevebachmeier, line 84: Not gonna lie, I'm not sure why this fix worked but previously we were having isort fail despite having it work fine when running locally (I confirmed same versions were being used).

Ref: https://pycqa.github.io/isort/docs/configuration/black_compatibility.html

zmbc: I think we'd want to at least remove --diff and add -v back to both commands -- those are somewhat important (first ensures we can't get into a bad state even if we merge a PR with failing CI, second allows us to see the details of where the failure is in the logs).

And, if it were to depend on the order the commands were run (which you've switched here), I think we'd want to know why.

zmbc: Ah, I bet it has to do with `--diff`: the "push" actions are succeeding while the "pull request" actions are failing.

stevebachmeier: fwiw the framework code all uses 
```
black . --check --diff
isort . --check --verbose --only-modified --diff
```

But that fails as well

stevebachmeier:
Closing this and will create a separate ticket

PR #123, opened 2022-12-24
Update logic for household moves

zmbc:
### Description
- *Category*: implementation
- *JIRA issue*: [MIC-3676](https://jira.ihme.washington.edu/browse/MIC-3676)
- *Research reference*: https://vivarium-research.readthedocs.io/en/latest/models/concept_models/vivarium_census_synthdata/concept_model.html#household-moves

### Verification and Testing

Ran simulation to completion; stepped through and verified that households were moving as units and address_ids were being updated properly.


src/vivarium_census_prl_synth_pop/components/household.py
```diff
@@ -121,50 +106,38 @@ def on_time_step(self, event: Event):
         choose which households move;
         move those households to a new address
         """
+        # NOTE: Currently, it is possible for a household not to have a reference person;
+        # in this case, that household can no longer move.
         pop = self.population_view.get(event.index)
-        household_ids = pop.loc[
-            pop["relation_to_household_head"] == "Reference person", "household_id"
-        ]
+        reference_people = pop[pop["relation_to_household_head"] == "Reference person"]
```
rmudambi, line 112: Once we take care of the reference person mortality issue you could potentially have multiple reference people in a household (the dead one and the new one). You will want to filter the population to include only live simulants.

zmbc: Oh wow, I did not realize that simulants who were not alive were not included by default. I think I need to make this change in the PersonMigration component as well. And it looks like it should be made in employment too -- I don't see anywhere that we exclude nonliving simulants from changing jobs.

PR #125, opened 2022-12-27
Only perform migration events on living simulants

zmbc:
### Description

- *Category*: bugfix
- *JIRA issue*: [MIC-3723](https://jira.ihme.washington.edu/browse/MIC-3723)
- *Research reference*: none

### Verification and Testing

Verified simulation still runs, and that this call returns a slightly smaller dataframe than before.

### DIFF
--- a/integration_tests/test_domestic_migration.py
+++ b/integration_tests/test_domestic_migration.py
@@ -1,0 +1,39 @@
+import pytest
+
+# TODO: Broader test coverage
+
+
+def test_individuals_do_move(populations):
+    for before, after in zip(populations, populations[1:]):
+        common_simulants = before.index.intersection(after.index)
+        before = before.loc[common_simulants]
+        after = after.loc[common_simulants]
+
+        individual_movers = before["household_id"] != after["household_id"]
+        assert individual_movers.any()
+        assert (
+            before[individual_movers]["address_id"] != after[individual_movers]["address_id"]
+        ).all()
+
+
+def test_households_do_move(populations):
+    for before, after in zip(populations, populations[1:]):
+        common_simulants = before.index.intersection(after.index)
+        before = before.loc[common_simulants]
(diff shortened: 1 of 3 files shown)


rmudambi (approved):
Can you also include the prohibition on dead people changing jobs?

Also include tests for this if possible.

zmbc:
> Can you also include the prohibition on dead people changing jobs?

See #126, I figured that would be merged directly to main so I separated them.

zmbc:
@rmudambi I've added a test, but because this test needs to look at consecutive time steps, I have refactored the testing setup to make that easier. Do you think those changes are appropriate to include in this PR or should I extract them? I will need this capability for most migration tests, I think.

integration_tests/test_domestic_migration.py
```diff
@@ -0,0 +1,39 @@
...
+
+
+def test_only_living_people_move(populations):
+    for before, after in zip(populations, populations[1:]):
+        common_simulants = before.index.intersection(after.index)
+        before = before.loc[common_simulants]
+        after = after.loc[common_simulants]
+
+        movers = before["address_id"] != after["address_id"]
+        assert after[movers]["tracked"].all()
```
rmudambi, line 38: What if there were no dead people at all? You should probably assert that there are some people that are dead who could have moved.

rmudambi (approved):

integration_tests/test_domestic_migration.py
zmbc: Hmm... I was thinking that the test here was that the component only moves living simulants; that people die in the sim is a pretty basic assumption that IMO would go in a test of the mortality component. But, if that doesn't sound right to you I can add the assert here.

rmudambi: No I think you're right, we should test that independently

PR #127, opened 2022-12-28
Trigger employment change on migration

zmbc:
### Description

- *Category*: implementation
- *JIRA issue*: [MIC-3677](https://jira.ihme.washington.edu/browse/MIC-3677)
- *Research reference*: https://vivarium-research.readthedocs.io/en/latest/models/concept_models/vivarium_census_synthdata/concept_model.html#simulation-strategy

### Verification and Testing

Automated integration tests, also stepped through to check reasonable numbers of movers are changing employment.

### DIFF
--- a/integration_tests/test_employment.py
+++ b/integration_tests/test_employment.py
@@ -47,18 +47,13 @@ def test_employed_have_income(tracked_live_populations, time_step):
     assert (employed["income"] > 0).all()
 
 
-def test_movers_change_employment(sim):
-    pop_prev = sim.get_population()
-    sim.step()
-
-    for _ in range(10):
-        pop = sim.get_population()
-
-        common_working_age_simulants = pop_prev.index[pop_prev.age >= 18].intersection(
-            pop.index[pop.age >= 18]
+def test_movers_change_employment(tracked_live_populations):
+    for before, after in zip(tracked_live_populations, tracked_live_populations[1:]):
+        common_working_age_simulants = before.index[before.age >= 18].intersection(
+            after.index[after.age >= 18]
         )
-        before = pop_prev.loc[common_working_age_simulants]
-        after = pop.loc[common_working_age_simulants]
+        before = before.loc[common_working_age_simulants]
+        after = after.loc[common_working_age_simulants]
 
         moved = before["address_id"] != after["address_id"]
         moved_to_military_gq = moved & (after["housing_type"] == "Military")
@@ -67,5 +62,3 @@ def test_movers_change_employment(sim):
         assert (
             before[should_change]["employer_id"] != after[should_change]["employer_id"]
         ).all()
-        pop_prev = pop
-        sim.step()
--- a/integration_tests/test_employment.py
+++ b/integration_tests/test_employment.py
@@ -47,8 +47,8 @@ def test_employed_have_income(tracked_live_populations, time_step):
     assert (employed["income"] > 0).all()
 
 
-def test_movers_change_employment(tracked_live_populations):
-    for before, after in zip(tracked_live_populations, tracked_live_populations[1:]):
+def test_movers_change_employment(populations):
+    for before, after in zip(populations, populations[1:]):
         common_working_age_simulants = before.index[before.age >= 18].intersection(
             after.index[after.age >= 18]
         )
--- a/integration_tests/test_employment.py
+++ b/integration_tests/test_employment.py
@@ -45,3 +45,27 @@ def test_employed_have_income(tracked_live_populations, time_step):
     # All people in military group quarters are
     employed = pop[pop["employer_id"] != data_values.UNEMPLOYED_ID]
     assert (employed["income"] > 0).all()
+
+
+def test_movers_change_employment(sim):
+    pop_prev = sim.get_population()
+    sim.step()
+
+    for _ in range(10):
+        pop = sim.get_population()
+
+        common_working_age_simulants = pop_prev.index[pop_prev.age >= 18].intersection(
+            pop.index[pop.age >= 18]
+        )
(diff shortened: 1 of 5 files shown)


integration_tests/test_employment.py
```diff
@@ -45,3 +45,27 @@ def test_employed_have_income(tracked_live_populations, time_step):
     # All people in military group quarters are
     employed = pop[pop["employer_id"] != data_values.UNEMPLOYED_ID]
     assert (employed["income"] > 0).all()
+
+
+def test_movers_change_employment(sim):
```
rmudambi, line 50: The testing framework is designed to have all the information you need without needing to take any more time-steps of the simulation. You should be able to do all of this by just looking at subsequent snapshots of the population state table stored in the `populations` fixture.

src/vivarium_census_prl_synth_pop/components/businesses.py
```diff
@@ -168,11 +171,14 @@ def on_time_step(self, event: Event):
...
 
-        # change jobs if of working age already
+        # change jobs if of working age already and either selected by rate or moved
+        # on this timestep
         working_age_idx = pop.loc[pop["age"] >= data_values.WORKING_AGE].index
         changing_jobs_idx = filter_by_rate(
             working_age_idx, self.randomness, self.job_change_rate, "changing jobs"
         )
+        moved_idx = pop.index[pop["address_id"] != pop["previous_timestep_address_id"]]
+        changing_jobs_idx = changing_jobs_idx.union(moved_idx.intersection(working_age_idx))
```
rmudambi, line 181: So every unemployed person who moves will get a job?

rmudambi (changes requested):
Looks good, but I'd like you to change how your test works so you don't have to do any time-steps. That simulation object you are using is the same one used by every other test, so we don't want to mess with it during tests.

integration_tests/test_employment.py
zmbc: I've extracted the changes I need for this into #128.

src/vivarium_census_prl_synth_pop/components/businesses.py
zmbc: Yes -- though I have made a TODO to spell this out more as a limitation in the docs. It's not accurate (obviously) but shouldn't be super important for PRL purposes.

integration_tests/test_employment.py
zmbc: @rmudambi I've now corrected this test to use the `populations` fixture.

integration_tests/test_employment.py
```diff
@@ -45,3 +45,20 @@ def test_employed_have_income(tracked_live_populations, time_step):
     # All people in military group quarters are
     employed = pop[pop["employer_id"] != data_values.UNEMPLOYED_ID]
     assert (employed["income"] > 0).all()
+
+
+def test_movers_change_employment(tracked_live_populations):
+    for before, after in zip(tracked_live_populations, tracked_live_populations[1:]):
```
rmudambi, line 51: Add a comment explaining which populations you're comparing

src/vivarium_census_prl_synth_pop/components/businesses.py
rmudambi: I think this will give jobs to (unemployed) people under 18 who move. Even if it doesn't you should explicitly test that you aren't giving them jobs.

rmudambi (changes requested):

src/vivarium_census_prl_synth_pop/components/businesses.py
zmbc: This test checks for that: https://github.com/ihmeuw/vivarium_census_prl_synth_pop/pull/127/files#diff-5188a0bafecb0feeccbd82bdd91b9bf65eb54464a68982faaa9f5e542d1c6c11R27-R31

The code here doesn't give people under 18 jobs because of the `.intersection(working_age_idx)` but I can separate that into a named variable instead of doing it inline for clarity.

integration_tests/test_employment.py
zmbc: Do you mean a comment explaining why I am using the tracked live population instead of total population? I actually think this would be a slightly better check if I use the total population, so I'll switch to that.

zmbc: It's a bit of a tension between making tests sensitive and making them non-overlapping. I used the living population here because we already have a test that checks that only living people move. But, I think it's fine to have overlapping tests especially if it makes it harder to miss an issue.

src/vivarium_census_prl_synth_pop/components/businesses.py
```diff
@@ -178,7 +178,8 @@ def on_time_step(self, event: Event):
             working_age_idx, self.randomness, self.job_change_rate, "changing jobs"
         )
         moved_idx = pop.index[pop["address_id"] != pop["previous_timestep_address_id"]]
-        changing_jobs_idx = changing_jobs_idx.union(moved_idx.intersection(working_age_idx))
+        moved_working_age_idx = moved_idx.intersection(working_age_idx)
+        changing_jobs_idx = changing_jobs_idx.union(moved_working_age_idx)
```
rmudambi, line 182: We have a test that people under 18 remain unemployed. I'm kind of surprised that didn't fail prior to this fix. Can you see why that was? It's a sign that our test coverage had a gap.

zmbc: This diff didn't change any logic -- just moved `moved_idx.intersection(working_age_idx)` onto its own line/variable.

rmudambi (approved):

PR #128, opened 2022-12-28
Save all populations for integration tests

zmbc:
### Description

- *Category*: chore
- *JIRA issue*: none
- *Research reference*: none

### Changes and notes

I think pretty much all tests having to do with migration are going to need to compare consecutive populations, which isn't well served by saving only a few populations.

### Verification and Testing

Ran existing tests, ran simulation to completion.

### DIFF
--- a/integration_tests/conftest.py
+++ b/integration_tests/conftest.py
@@ -1,4 +1,4 @@
-from typing import Dict
+from typing import List
 
 import pandas as pd
 import pytest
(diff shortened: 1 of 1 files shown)


albrja (approved):

integration_tests/conftest.py
```diff
@@ -20,14 +20,17 @@ def sim() -> InteractiveContext:
...
-def tracked_live_populations(sim) -> Dict[int, pd.DataFrame]:
-    previous_step_number = 0
-    population_states = {}
-    for step in TIME_STEPS_TO_TEST:
-        sim.take_steps(step - previous_step_number)
-
+def populations(sim) -> List[pd.DataFrame]:
+    population_states = []
+    for _ in range(max(TIME_STEPS_TO_TEST) + 1):
         pop = sim.get_population()
```
rmudambi, line 26: Note this is not going to return untracked simulants. Consider adding the `untracked=True` argument to `get_population` and filtering out untracked in addition to dead in the `tracked_live_populations` fixture below.

rmudambi (approved):

integration_tests/conftest.py
zmbc: I've just added this -- I will need it for emigration.

PR #131, opened 2022-12-29
ENH: add social security observer

aflaxman:
### Description

- *Category*: observers
- *JIRA issue*: MIC-3732 (https://jira.ihme.washington.edu/browse/MIC-3732)
- *Research reference*: https://vivarium-research.readthedocs.io/en/latest/models/concept_models/vivarium_census_synthdata/concept_model.html?highlight=ssa#social-security-observer

### Changes and notes
 
turned out to be pretty simple, see design doc https://jira.ihme.washington.edu/browse/MIC-3651

### Verification and Testing
ran pytests successfully, ran simulate with model_spec.yaml and confirmed that social_security.hdf has expected fields

image

### DIFF
--- a/src/vivarium_census_prl_synth_pop/components/observers.py
+++ b/src/vivarium_census_prl_synth_pop/components/observers.py
@@ -416,9 +416,7 @@ class WICObserver(BaseObserver):
 
 
 class SocialSecurityObserver(BaseObserver):
-    """Class for observing columns relevant to Social Security registry.
-
-    """
+    """Class for observing columns relevant to Social Security registry."""
 
     ADDITIONAL_INPUT_COLUMNS = ["tracked", "alive", "entrance_time", "exit_time", "ssn"]
     OUTPUT_COLUMNS = [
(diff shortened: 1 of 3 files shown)


src/vivarium_census_prl_synth_pop/components/observers.py
```diff
@@ -413,3 +413,70 @@ def do_observation(self, event) -> None:
...
+        return self.OUTPUT_COLUMNS
+
+    def setup(self, builder: Builder):
+        super().setup(builder)
+        self.clock = builder.time.clock()
+        self.end_time = dt.datetime(**builder.configuration.time.end)
+
+    def to_observe(self, event: Event) -> bool:
+        """Only observe if this is the final time step of the sim"""
+        return self.clock() <= self.end_time < event.time
```
rmudambi, line 459: I'm not sure about the edge case here, but I think it's safer to do this since I don't think a new time-step would be triggered if `self.end_time == event.time` and therefore we might not record anything. 
```suggestion
        return self.clock() < self.end_time <= event.time
```
If we are confident in vivarium's behavior around that edge case, we could get rid of the `self.clock()` check entirely.

src/vivarium_census_prl_synth_pop/components/observers.py
```diff
@@ -413,3 +413,70 @@ def do_observation(self, event) -> None:
...
             [self.responses, pop_included_mothers] + list(pop_included.values())
         )[self.output_columns]
+
+
+class SocialSecurityObserver(BaseObserver):
+    """Class for observing columns relevant to Social Security registry.
+
+    """
+
+    ADDITIONAL_INPUT_COLUMNS = ["alive", "entrance_time", "exit_time", "ssn"]
```
rmudambi, line 423: You will need to include `"tracked"` as an input column, because otherwise an implicit `tracked == True` query is applied when you do `self.population_view.get()` and you need to be able to record the SSNs of people who become untracked during the sim.

src/vivarium_census_prl_synth_pop/components/observers.py
```diff
@@ -413,3 +413,70 @@ def do_observation(self, event) -> None:
...
+            event.index,
+            query="ssn == True",  # only include simulants with a SSN
+        )
+        df_creation = pop.filter([
+            "first_name_id",
+            "middle_name_id",
+            "last_name_id",
+            "date_of_birth"])
+        df_creation["event_type"] = "creation"
+        df_creation["event_date"] = pop["entrance_time"]
```
rmudambi, line 472: Note: a large number of people will have their entrance time be the beginning of the simulation, which is not when they would have gotten their SSN. Should we use their birth date instead if it is less than the entrance time? Or maybe we should only record SSN creations that happen after the simulation starts.

rmudambi (approved):
Looks good if you add the tracked column. I have a note about SSN creation events, but that's a research question.

**IMPORTANT**: The build is broken right now. Did you run `isort` and `black`?

src/vivarium_census_prl_synth_pop/components/observers.py
zmbc: This was definitely an oversight to not spell out more clearly in the docs, but I think it is heavily implied by the example in the docs that DOB should be used when it is before the simulation start.

aflaxman:
> **IMPORTANT**: The build is broken right now. It needs to be fixed before you can merge. Did you run `isort` and `black`?

I did not, and it looks like that is the issue.  Can you point me to some docs on how we are using these tools?

src/vivarium_census_prl_synth_pop/components/observers.py
aflaxman: I tested this by changing the end time to be a time step, which confirms that @rmudambi 's approach to the edge case achieves the intended result.

src/vivarium_census_prl_synth_pop/components/observers.py
aflaxman: 👍

src/vivarium_census_prl_synth_pop/components/observers.py
aflaxman: Thanks for catching this, it looks like the docs are actually pretty clear on this, and it was my oversight, "Everyone who starts in the sim with a SSN will have a creation record made with their date of birth."

PR #132, opened 2022-12-30
Test new household moves

zmbc:
### Description

- *Category*: test
- *Research reference*: none

### Changes and notes

Tests functionality added in #101

### DIFF
--- a/integration_tests/test_domestic_migration.py
+++ b/integration_tests/test_domestic_migration.py
@@ -17,22 +17,24 @@ def test_individuals_move_into_new_households(simulants_on_adjacent_timesteps):
         new_household_movers = (
             (before["household_id"] != after["household_id"])
             & (after["housing_type"] == "Standard")
-            & (after["relation_to_household_head"] == "Reference person")
+            & (~after["household_id"].isin(before["household_id"]))
         )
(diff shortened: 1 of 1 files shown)


integration_tests/test_domestic_migration.py
```diff
@@ -12,6 +12,29 @@ def test_individuals_move(simulants_on_adjacent_timesteps):
         ).all()
 
 
+def test_individuals_move_into_new_households(simulants_on_adjacent_timesteps):
+    for before, after in simulants_on_adjacent_timesteps:
+        new_household_movers = (
+            (before["household_id"] != after["household_id"])
+            & (after["housing_type"] == "Standard")
+            & (after["relation_to_household_head"] == "Reference person")
```
rmudambi, line 20: Instead of saying they're a reference person after the move, why not say the household they moved into didn't exist in the before time-step? That seems more like the definition of a new household move to me. Then test that they're a reference person later.

rmudambi (approved):

integration_tests/test_domestic_migration.py
zmbc: I was on the fence about that when writing this. I'll switch it!

PR #134, opened 2022-12-30
Test non-reference-person moves

zmbc:
### Description

- *Category*: test
- *Research reference*: none

### Changes and notes

Tests functionality added in #109

### DIFF
--- a/integration_tests/test_domestic_migration.py
+++ b/integration_tests/test_domestic_migration.py
@@ -54,6 +54,27 @@ def test_individuals_move_into_group_quarters(simulants_on_adjacent_timesteps):
         )
 
 
+def test_individuals_move_into_existing_households(simulants_on_adjacent_timesteps):
+    for before, after in simulants_on_adjacent_timesteps:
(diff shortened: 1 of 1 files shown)


integration_tests/test_domestic_migration.py
```diff
@@ -12,6 +12,27 @@ def test_individuals_move(simulants_on_adjacent_timesteps):
         ).all()
 
 
+def test_individuals_move_into_existing_households(simulants_on_adjacent_timesteps):
+    for before, after in simulants_on_adjacent_timesteps:
+        non_reference_person_movers = (
+            (before["household_id"] != after["household_id"])
+            & (after["housing_type"] == "Standard")
+            & (after["relation_to_household_head"] != "Reference person")
```
rmudambi, line 20: Instead of saying they're not a reference person after the move, why not say the household they moved into existed in the before time-step? That seems more like the definition of a non-reference person move to me.

rmudambi (approved):

PR #137, opened 2022-12-30
Test address ID

zmbc:
### Description
- *Category*: test
- *Research reference*: none

### DIFF
--- a/integration_tests/test_household_structure.py
+++ b/integration_tests/test_household_structure.py
@@ -39,3 +39,14 @@ def test_all_households_have_reference_person(tracked_live_populations):
         # Assert these two sets are identical
         assert non_gq_household_ids.size == reference_person_household_ids.size
         assert np.setxor1d(non_gq_household_ids, reference_person_household_ids).size == 0
+
+
+def test_household_id_and_address_id_correspond(tracked_live_populations):
+    for pop in tracked_live_populations:
+        # 1-to-1 at any given point in time
+        assert (pop.groupby("household_id")["address_id"].nunique() == 1).all()
+        assert (pop.groupby("address_id")["household_id"].nunique() == 1).all()
+
+    # Even over time, there is only 1 household_id for each address_id -- they are not reused
+    all_time_pop = pd.concat(tracked_live_populations, ignore_index=True)
+    assert (all_time_pop.groupby("address_id")["household_id"].nunique() == 1).all()
--- a/integration_tests/test_household_structure.py
+++ b/integration_tests/test_household_structure.py
@@ -43,6 +43,8 @@ def test_all_households_have_reference_person(tracked_live_populations):
 
 def test_household_id_and_address_id_correspond(tracked_live_populations):
     for pop in tracked_live_populations:
+        assert pop["household_id"].notnull().all()
+        assert pop["address_id"].notnull().all()
         # 1-to-1 at any given point in time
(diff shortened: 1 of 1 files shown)


integration_tests/test_household_structure.py
```diff
@@ -29,3 +30,16 @@ def test_all_households_have_reference_person(tracked_live_populations, time_ste
...
+    for pop in tracked_live_populations:
+        assert pop["household_id"].notnull().all()
+        assert pop["address_id"].notnull().all()
+        # 1-to-1 at any given point in time
+        assert (pop.groupby("household_id")["address_id"].nunique() == 1).all()
+        assert (pop.groupby("address_id")["household_id"].nunique() == 1).all()
+
+    # Even over time, there is only 1 household_id for each address_id -- they are not reused
+    all_time_pop = pd.concat(tracked_live_populations, ignore_index=True)
+    assert (all_time_pop.groupby("address_id")["household_id"].nunique() == 1).all()
```
rmudambi, line 45: If you do the test on `all_time_pop` like this in both directions, why do you need the loop?

rmudambi (approved):

integration_tests/test_household_structure.py
zmbc: I only do it in one direction on `all_time_pop` because it is only true in one direction: over time, a household ID can change address IDs in a household move.

integration_tests/test_household_structure.py
```diff
@@ -29,3 +30,18 @@ def test_all_households_have_reference_person(tracked_live_populations, time_ste
...
+
+def test_household_id_and_address_id_correspond(tracked_live_populations):
+    for pop in tracked_live_populations:
+        assert pop["household_id"].notnull().all()
+        assert pop["address_id"].notnull().all()
+        # 1-to-1 at any given point in time
+        assert (pop.groupby("household_id")["address_id"].nunique() == 1).all()
+        assert (pop.groupby("address_id")["household_id"].nunique() == 1).all()
+
+    # Even over time, there is only 1 household_id for each address_id -- they are not reused
```
stevebachmeier, line 43: "They are not reused" - clarify exactly which is not reused (household_id or address_id)

integration_tests/test_household_structure.py
```diff
@@ -29,3 +30,18 @@ def test_all_households_have_reference_person(tracked_live_populations, time_ste
...
+    # Note, however, that the reverse is not true: a household_id can span multiple address_ids
```
stevebachmeier, line 44: So that I understand - you're saying that "a household_id can span multiple address_ids" _in different time steps_?

stevebachmeier (approved):

rmudambi:
I'd suggest a test about not reusing address ids - i.e. a household can't move into a previously vacated address

integration_tests/test_household_structure.py
zmbc: Yes. I will clarify this comment.

zmbc:
> I'd suggest a test about not reusing address ids - i.e. a household can't move into a previously vacated address

That is tested [here](https://github.com/ihmeuw/vivarium_census_prl_synth_pop/pull/137/files#diff-1650a4d8c21b09fcd51aa68dae705380d1980c1399c85fbe5f3dab90d506789fR47).

rmudambi (approved):

PR #135, opened 2022-12-30
Factor out shared code in migration tests

zmbc:
### Description
- *Category*: refactor
- *JIRA issue*: [MIC-3722](https://jira.ihme.washington.edu/browse/MIC-3722)
- *Research reference*: none

PR #138, opened 2022-12-30
Feature/varying income

albrja:
### Gives simulants varying income depending upon age, sex, race/ethnicity, and employer.
- *Category*: Feature
- *JIRA issue*: [MIC-3657](https://jira.ihme.washington.edu/browse/MIC-3657)
- *Research reference*: https://vivarium-research.readthedocs.io/en/latest/models/concept_models/vivarium_census_synthdata/concept_model.html#income-20

### Changes and notes
-Added personal_income_propensity and employer_income_propensity columns to state table. 
-Removed income column from state table and made it a pipeline with the same name.
-Added income distributions csv to raw data directory for lookup table for income pipeline
-Fixed bug in businesses.py that was allowing simulants living and working in military group quarters to change jobs
-Added tests for income propensities and income pipelines.

### Verification and Testing
Successfully ran simulation with 250k simulants (20 minutes) and passed all tests.


integration_tests/conftest.py
```diff
@@ -16,16 +16,23 @@ def sim() -> InteractiveContext:
     return simulation
 
 
-TIME_STEPS_TO_TEST = [0, 1, 10]
+TIME_STEPS_TO_TEST = [0, 1]
```
zmbc, line 19: I don't think we want this.

integration_tests/conftest.py
```diff
@@ -16,16 +16,23 @@ def sim() -> InteractiveContext:
...
     for _ in range(max(TIME_STEPS_TO_TEST) + 1):
         pop = sim.get_population(untracked=True)
+        pipelines = sim.list_values()
+        for pipeline in pipelines:
+            # Metrics is a dict we cannot concat and do not want to
+            if pipeline == "metrics":
+                continue
+            pop = pd.concat([pop, sim.get_value(pipeline)(pop.index)], axis=1)
+            # Handle column names for pipelines
+            pop = pop.set_axis([*pop.columns[:-1], pipeline], axis=1, inplace=False)
```
zmbc, line 34: What is this doing? Is `pipeline` always a string? Couldn't it just be:

`pop[pipeline] = sim.get_value(pipeline)(pop.index)`

integration_tests/test_employment.py
```diff
@@ -73,3 +75,153 @@ def test_movers_change_employment(populations):
...
+        (
+            employer_ids_before,
+            employer_propensity_before,
+            employer_ids_after,
+            employer_propensity_after,
+        ) = get_consecutive_time_step_employer_ids_and_test_column_for_comparable_sims(
+            tracked_live_populations, time_step, "employer_income_propensity"
+        )
+
+        # Find those that change employers
```
zmbc, line 113:
```suggestion
        # Find those that don't change employers
```

integration_tests/test_employment.py
```diff
@@ -73,3 +75,153 @@ def test_movers_change_employment(populations):
...
+            employer_ids_before != employer_ids_after
+        ]
+
+        assert (
+            employer_propensity_before.loc[changed_jobs_idx]
+            != employer_propensity_after.loc[changed_jobs_idx]
+        ).all()
+
+
+def test_no_employment_income_propensity_updates(tracked_live_populations):
```
zmbc, line 102: I think these two tests could become one test pretty easily by making `changed_jobs` and `changed_employer_propensity` booleans and asserting that they are always equal.

integration_tests/test_employment.py
```diff
@@ -73,3 +75,153 @@ def test_movers_change_employment(populations):
...
+
+
+def test_employment_income_propensity_updates(tracked_live_populations):
+    for time_step in range(max(TIME_STEPS_TO_TEST)):
+        (
+            employer_ids_before,
+            employer_propensity_before,
+            employer_ids_after,
+            employer_propensity_after,
+        ) = get_consecutive_time_step_employer_ids_and_test_column_for_comparable_sims(
```
zmbc, line 87: I don't have a suggestion, but this method name seems too long.

integration_tests/test_employment.py
```diff
@@ -73,3 +75,153 @@ def test_movers_change_employment(populations):
...
+        changed_jobs_idx = employer_ids_before.index[
+            employer_ids_before != employer_ids_after
+        ]
+
+        assert (
+            income_before.loc[changed_jobs_idx] != income_after.loc[changed_jobs_idx]
+        ).all()
+
+
+def test_no_income_uupdate(tracked_live_populations):
```
zmbc, line 161:
```suggestion
def test_no_income_update(tracked_live_populations):
```
And as before, I think this test and the previous could be made into one.

integration_tests/test_employment.py
```diff
@@ -73,3 +75,153 @@ def test_movers_change_employment(populations):
...
+        )
+
+        # Find those that change employers
+        static_jobs_idx = employer_ids_before.index[employer_ids_before == employer_ids_after]
+
+        assert (income_before.loc[static_jobs_idx] == income_after.loc[static_jobs_idx]).all()
+
+
+def test_income_updates_when_age_changes(tracked_live_populations):
+    # This test checks whether simulants who have a birthdays and therefore will have a new income distribution
```
zmbc, line 179: I don't understand this comment; rephrase?

integration_tests/test_employment.py
```diff
@@ -73,3 +75,153 @@ def test_movers_change_employment(populations):
...
+    # Note: The other tests for income implementation are contigent on comparing simulants who are the same age and
+    #  therefore ignore this edge case.
+    for time_step in range(max(TIME_STEPS_TO_TEST)):
+        pop_before = tracked_live_populations[time_step]
+        pop_after = tracked_live_populations[time_step]
```
zmbc, line 184: Isn't this comparing a population with itself?

Add an assert that there are `birthday_sims`.

src/vivarium_census_prl_synth_pop/constants/data_values.py
```diff
@@ -77,6 +79,9 @@
 
 BUSINESS_MOVE_RATE_YEARLY = 0.1  # 10 changes per 100 py
 
+PERSONAL_INCOME_PROPENSITY_DISTRIBUTION = stats.norm(loc=0.0, scale=0.812309 ** 0.5)
+EMPLOYER_INCOME_PROPENSITY_DISTRIBUTION = stats.norm(loc=0.0, scale=0.187691 ** 0.5)
```
mattkappel, line 83: Get rid of the spaces in the RHS of `scale`.

integration_tests/test_employment.py
```diff
@@ -73,3 +75,153 @@ def test_movers_change_employment(populations):
...
+    time_step,
+    test_column,
+) -> tuple:
+    pop_before = tracked_live_populations[time_step]
+    pop_after = tracked_live_populations[time_step + 1]
+    common_sims_idx = pop_before.index.intersection(pop_after.index)
+
+    # Deal with sims who change age bins on time step
+    # This is handling the cases where simulants enter the next age pin so their distribution for income changes
+    age_before = np.floor(pop_before.loc[common_sims_idx, "age"])
```
zmbc, line 212: This code is duplicated with the above, is there a refactor we can do here?

integration_tests/test_employment.py
```diff
@@ -73,3 +75,153 @@ def test_movers_change_employment(populations):
...
+    tracked_live_populations,
+    time_step,
+    test_column,
+) -> tuple:
+    pop_before = tracked_live_populations[time_step]
+    pop_after = tracked_live_populations[time_step + 1]
+    common_sims_idx = pop_before.index.intersection(pop_after.index)
+
+    # Deal with sims who change age bins on time step
+    # This is handling the cases where simulants enter the next age pin so their distribution for income changes
```
zmbc, line 211:
```suggestion
    # This is handling the cases where simulants enter the next age bin so their distribution for income changes
```

src/vivarium_census_prl_synth_pop/components/businesses.py
```diff
@@ -106,9 +124,19 @@ def on_initialize_simulants(self, pop_data: SimulantData) -> None:
...
+            )
+            pop["employer_income_propensity"] = self.randomness.get_draw(
+                pop.index,
+                additional_key="employer_income_propensity",
+            )
+
             # merge on employer addresses and names
             pop = pop.merge(
-                self.businesses[self.columns_created[:-1]], on="employer_id", how="left"
+                self.businesses[self.columns_created[:3]], on="employer_id", how="left"
```
zmbc, line 139: This seems like a brittle change.

src/vivarium_census_prl_synth_pop/components/businesses.py
```diff
@@ -173,9 +204,14 @@ def on_time_step(self, event: Event):
             address_id_col_name="employer_address_id",
         )
 
-        # change jobs if of working age already and either selected by rate or moved
-        # on this timestep
-        working_age_idx = pop.loc[pop["age"] >= data_values.WORKING_AGE].index
+        # change jobs if of working age already but exclude simulants living in military group quarters
+        working_age_idx = pop.index[
```
zmbc, line 208: Is this change related to the other changes in this PR? If so, I don't immediately see how.

src/vivarium_census_prl_synth_pop/components/businesses.py
```diff
@@ -309,3 +353,32 @@ def _update_employer_metadata(
...
+        employed_idx = pop.index[pop["employer_id"] != data_values.UNEMPLOYED_ID]
+
+        # Get propensities for two components to get income propensity
+        personal_income_component = data_values.PERSONAL_INCOME_PROPENSITY_DISTRIBUTION.ppf(
+            pop.loc[employed_idx, "personal_income_propensity"]
+        )
+        employer_income_component = data_values.EMPLOYER_INCOME_PROPENSITY_DISTRIBUTION.ppf(
+            pop.loc[employed_idx, "employer_income_propensity"]
+        )
+        # Income propensity = probit(personal_component + employer_component)
```
zmbc, line 370: This comment is no longer correct, and I am not sure it adds anything on top of what line 372 already says.

integration_tests/test_employment.py
```diff
@@ -73,3 +75,153 @@ def test_movers_change_employment(populations):
...
+        employer_ids_before = pop_before.loc[common_sims_idx, "employer_id"]
+        employer_ids_after = pop_after.loc[common_sims_idx, "employer_id"]
+        same_jobs_idx = employer_ids_before.index[employer_ids_before == employer_ids_after]
+
+        assert (
+            pop_before.loc[same_jobs_idx, "income"] != pop_after.loc[same_jobs_idx, "income"]
+        ).all()
+
+
+def get_consecutive_time_step_employer_ids_and_test_column_for_comparable_sims(
```
zmbc, line 201: I think this probably needs to be merged with #135 -- whether by you or me (depends who merges to main first, I guess).

zmbc (changes requested):

src/vivarium_census_prl_synth_pop/constants/data_values.py
albrja: Thankls

integration_tests/conftest.py
albrja: Good catch I have reverted this after testing.

integration_tests/conftest.py
albrja: The pipeline wasn't always named so there were columns with no names in the dataframe.  This is giving the columns the name of the pipeline.

zmbc: Right, but if the `pipeline` variable is always a string name for the column, why wouldn't my suggestion work to replace lines 32-34?

src/vivarium_census_prl_synth_pop/components/businesses.py
albrja: Yes.  This is excluding military simulants otherwise they were potentially changing jobs >> getting a new employer propensity >> then being assigned back to military employer.  That is why this was changed and military index chunk of code was moved.

zmbc: Ah, makes sense! Carry on.

integration_tests/conftest.py
albrja: Oh you are right.  Thanks.

albrja: So this doesnt work because one of the pipelines seems to be a dataframe which doesnt allow it to be a column.

integration_tests/test_employment.py
zmbc: @albrja Specifically, as discussed this morning, this should make use of the `simulants_on_adjacent_timesteps` fixture, which is now merged into main.

integration_tests/conftest.py
```diff
@@ -24,8 +24,22 @@ def populations(sim) -> List[pd.DataFrame]:
...
+        for pipeline in pipelines:
+            p = sim.get_value(pipeline)(pop.index)
+            # Metrics is a dict we cannot concat and do not want to
+            if isinstance(p, dict):
+                continue
+            elif isinstance(p, pd.DataFrame):
+                renamed_cols = [f"{pipeline}.{col}" for col in p.columns]
+                rename_dict = dict(zip(p.columns, renamed_cols))
+                p = p.rename(columns=rename_dict)
+                pop = pd.concat([pop, p], axis=1)
```
zmbc, line 37:
```suggestion
                pop = pd.concat([pop, p.add_prefix(f"{pipeline}.")], axis=1)
```

integration_tests/test_employment.py
```diff
@@ -73,3 +75,80 @@ def test_movers_change_employment(populations):
         assert (
             before[should_change]["employer_id"] != after[should_change]["employer_id"]
         ).all()
+
+
+def test_employment_income_propensity_updates(simulants_on_adjacent_timesteps):
```
zmbc, line 80: This test and the next can be combined pretty easily: make a boolean for whether employment changed, and a boolean for whether the employer_income_propensity changed, and assert that those two boolean Series are always equal.

integration_tests/test_employment.py
```diff
@@ -73,3 +75,80 @@ def test_movers_change_employment(populations):
...
+
+
+def test_personal_income_propensity_is_constant(simulants_on_adjacent_timesteps):
+    for before, after in simulants_on_adjacent_timesteps:
+        assert (
+            before["personal_income_propensity"] == after["personal_income_propensity"]
+        ).all()
+
+
+def test_income_updates(simulants_on_adjacent_timesteps):
```
zmbc, line 113: Ditto for these two tests.

integration_tests/test_employment.py
```diff
@@ -73,3 +75,80 @@ def test_movers_change_employment(populations):
...
+            (before["employer_id"] != after["employer_id"])
+            & (np.floor(before["age"]) == np.floor(after["age"]))
+        ]
+
+        assert (
+            before.loc[changed_jobs_idx, "income"] != after.loc[changed_jobs_idx, "income"]
+        ).all()
+
+
+def test_no_income_uupdate(simulants_on_adjacent_timesteps):
```
zmbc, line 125:
```suggestion
def test_no_income_update(simulants_on_adjacent_timesteps):
```

zmbc (commented):
A few comments from the previous review still apply; I have left those unresolved.

src/vivarium_census_prl_synth_pop/components/businesses.py
rmudambi: I agree. It's much better to name the columns than refer to them by index like this. If you want to exclude specific columns from `self.columns_created` you should do that explicitly.

src/vivarium_census_prl_synth_pop/components/businesses.py
```diff
@@ -125,17 +153,20 @@ def on_initialize_simulants(self, pop_data: SimulantData) -> None:
                 ] = data_values.MilitaryEmployer.EMPLOYER_ID
                 pop = self._update_employer_metadata(pop, military_index)
```
rmudambi, line 154: You should move the code common to both branches of the if statement outside and call them once after the `if`. This includes setting the personal income propensity and calling `population_view.update`. You could also move the employer propensity outside, though that would mean drawing for the newborns rather than just setting 0.0.

src/vivarium_census_prl_synth_pop/components/businesses.py
```diff
@@ -200,23 +235,32 @@ def on_time_step(self, event: Event):
             pop.loc[turning_working_age, "employer_id"] = self.assign_random_employer(
                 turning_working_age
             )
-            pop = self._update_employer_metadata(pop, turning_working_age)
 
         # Give military gq sims military employment
-        military_index = pop.loc[
+        military_index = pop.index[
```
rmudambi, line 240: `new_military_idx`?

src/vivarium_census_prl_synth_pop/components/observers.py
```diff
@@ -69,6 +69,10 @@ def output_filename(self):
     def input_columns(self):
         pass
 
+    @property
+    def input_values(self):
+        return []
```
rmudambi, line 74: If you define `INPUT_VALUES` as an empty list in the base class, you can have this method return `self.INPUT_VALUES` and then you don't need to define the property in the `WICObserver`

src/vivarium_census_prl_synth_pop/model_specifications/model_spec.yaml
```diff
@@ -16,7 +16,7 @@ components:
 
 configuration:
     input_data:
-        artifact_path: '/home/rmudambi/ihme/team/priv/engineering/vivarium_census_prl_synth_pop/data/artifacts/categorical-test/florida.hdf'
+        artifact_path: '/mnt/team/simulation_science/priv/engineering/vivarium_census_prl_synth_pop/data/artifacts/categorical-test/florida.hdf'
```
rmudambi, line 19: oops sorry about that.

rmudambi (changes requested):

integration_tests/test_employment.py
```diff
@@ -73,3 +75,153 @@ def test_movers_change_employment(populations):
...
+        static_jobs_idx = employer_ids_before.index[employer_ids_before == employer_ids_after]
+
+        assert (
+            employer_propensity_before.loc[static_jobs_idx]
+            == employer_propensity_after.loc[static_jobs_idx]
+        ).all()
+
+
+def test_personal_income_propensity_is_constant(tracked_live_populations):
+    starting_personal_income_propensity = tracked_live_populations[min(TIME_STEPS_TO_TEST)][
```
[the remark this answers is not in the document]
albrja: This is outdated and I have updated it for the consecutive time step populations like the other tests.

src/vivarium_census_prl_synth_pop/components/businesses.py
albrja: I am not sure I understand this here.  The if branch is based on simulation initialization vs on time step when newborns are initialized.  Wouldn't moving any assignment for newborns to the if branch never be hit because that happens after start time?

integration_tests/test_employment.py
```diff
@@ -74,27 +75,18 @@ def test_movers_change_employment(populations):
 
 def test_employment_income_propensity_updates(simulants_on_adjacent_timesteps):
     for before, after in simulants_on_adjacent_timesteps:
-        changed_jobs_idx = before.index[
-            (before["employer_id"] != after["employer_id"])
-            & (np.floor(before["age"]) == np.floor(after["age"]))
-        ]
+        changed_jobs = before["employer_id"] != after["employer_id"]
 
+        # Employer income propensity changes when employer changes.
```
zmbc, line 80: I was imagining:

```
changed_employer_propensity = before["employer_income_propensity"] != after["employer_income_propensity"]
assert (changed_jobs == changed_employer_propensity).all()
```
instead of lines 81-90.

integration_tests/test_employment.py
```diff
@@ -105,43 +97,33 @@ def test_personal_income_propensity_is_constant(simulants_on_adjacent_timesteps)
...
-            & (np.floor(before["age"]) == np.floor(after["age"]))
-        ]
+        changed_jobs = (before["employer_id"] != after["employer_id"]) & (
+            np.floor(before["age"]) == np.floor(after["age"])
+        )
 
-        assert (
-            before.loc[static_jobs_idx, "income"] == after.loc[static_jobs_idx, "income"]
-        ).all()
+        # Test income updates for same age simulants when employer changes
```
zmbc, line 106: Similar to above:
```
changed_income = before["income"] != after["income"]
assert (changed_jobs == changed_income).all()
```

zmbc (approved):

rmudambi (approved):

integration_tests/conftest.py
zmbc: @rmudambi Did you and Jim decide that this wouldn't be a good change/wouldn't work? Nbd if so just wanted to make sure!

albrja: I do not think it matters for setup of tests but I have updated it as it is a lot cleaner.

PR #140, opened 2022-12-31
convert housing type, race, sex, and relation to hh head to categoric…

rmudambi:
…al columns

### Description

- *Category*: bugfix
- *JIRA issue*: [MIC-3733](https://jira.ihme.washington.edu/browse/MIC-3733)
- *Research reference*: 

### Changes and notes
Make object columns categorical to improve performance and ensure invalid values can't be set 

### Verification and Testing
Ran simulation and verified outputs were unchanged
Added integration tests

Note: the integration tests fail as written, since they require a re-creation of the artifact. They succeed when run against the newly created artifact here: `/mnt/team/simulation_science/priv/engineering/vivarium_census_prl_synth_pop/data/artifacts/categorical-test/florida.hdf`

### DIFF
--- a/integration_tests/test_demographics.py
+++ b/integration_tests/test_demographics.py
@@ -1,0 +1,37 @@
+import pandas as pd
+
+from vivarium_census_prl_synth_pop.constants import metadata
+
+
+# TODO: Broader test coverage
+def test_initial_population_size(sim, populations):
+    pop_size = sim.configuration.population.population_size
+    pop = populations[0]
+
+    assert pop.index.size == pop_size
+    assert pop["tracked"].all()
+    assert (pop["alive"] == "alive").all()
+
+
+# Todo: test that sex is unvarying
+# Todo: test that race is unvarying
+
+
+def test_sex_is_categorical(tracked_live_populations):
+    for pop in tracked_live_populations:
+        sex = pop["sex"]
+
+        # Assert the dtype is correct and that there are no NaNs
+        assert sex.dtype == pd.CategoricalDtype(categories=metadata.SEXES)
+        assert not sex.isnull().any()
+
+
+def test_race_ethnicity_is_categorical(tracked_live_populations):
+    for pop in tracked_live_populations:
+        race_ethnicity = pop["race_ethnicity"]
+
+        # Assert the dtype is correct and that there are no NaNs
+        assert race_ethnicity.dtype == pd.CategoricalDtype(
+            categories=metadata.RACE_ETHNICITIES
+        )
+        assert not race_ethnicity.isnull().any()
--- a/src/vivarium_census_prl_synth_pop/components/observers.py
+++ b/src/vivarium_census_prl_synth_pop/components/observers.py
@@ -140,7 +140,8 @@ class BaseObserver(ABC):
     # TODO: consider using compressed csv instead of hdf
     def on_simulation_end(self, event: Event) -> None:
         output_dir = utilities.build_output_dir(self.output_dir, subdir="results")
-        self.responses.to_hdf(output_dir / self.output_filename, key="data")
+        # 'fixed' format is not compatible with categorical data
+        self.responses.to_hdf(output_dir / self.output_filename, key="data", format="t")
 
 
 class HouseholdSurveyObserver(BaseObserver):
@@ -290,6 +291,7 @@ class WICObserver(BaseObserver):
     ADDITIONAL_OUTPUT_COLUMNS = ["wic_year"]
     WIC_BASELINE_SALARY = 16_410
     WIC_SALARY_PER_HOUSEHOLD_MEMBER = 8_732
+    WIC_RACE_ETHNICITIES = ["White", "Black", "Latino", "Other"]
 
     def __repr__(self):
         return f"WICObserver()"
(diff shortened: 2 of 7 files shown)


integration_tests/test_demographics.py
```diff
@@ -0,0 +1,45 @@
...
+from vivarium_census_prl_synth_pop.constants import metadata
+
+
+# TODO: Broader test coverage
+def test_initial_population_size(sim, tracked_live_populations):
+    pop_size = sim.configuration.population.population_size
+    pop = tracked_live_populations[0]
+
+    assert pop.index.size == pop_size
+    assert (pop["alive"] == "dead").sum() == 0
```
zmbc, line 12: This population is alive by definition so if you want to check for this you should use `populations` instead of `tracked_live_populations`.

src/vivarium_census_prl_synth_pop/components/population.py
```diff
@@ -270,9 +275,11 @@ def initialize_newborns(self, pop_data: SimulantData) -> None:
         new_births = new_births.merge(
             mothers[inherited_traits], left_on="parent_id", right_index=True
         )
-        new_births["relation_to_household_head"] = new_births[
-            "relation_to_household_head"
-        ].map(metadata.NEWBORNS_RELATION_TO_HOUSEHOLD_HEAD_MAP)
+        new_births["relation_to_household_head"] = (
+            new_births["relation_to_household_head"]
+            .map(metadata.NEWBORNS_RELATION_TO_HOUSEHOLD_HEAD_MAP)
+            .astype(new_births["relation_to_household_head"].dtype)
```
zmbc, line 281: I don't know if this would be bad form to do in a metadata file, but you could consider defining the dtypes themselves there in addition to the lists of values.

src/vivarium_census_prl_synth_pop/constants/data_values.py
```diff
@@ -29,6 +29,8 @@
     4: "Military",
     5: "Other non-institutional",
 }
+HOUSING_TYPES = ["Standard"] + list(GQ_HOUSING_TYPE_MAP.values())
```
zmbc, line 32: This makes me a little nervous because it depends on dictionary order, but I guess that is okay now that we only support versions above 3.7?

src/vivarium_census_prl_synth_pop/constants/metadata.py
```diff
@@ -126,8 +127,11 @@
     8: "Multiracial or Other",
     9: "Multiracial or Other",
 }
+RACE_ETHNICITIES = set(RACE_ETHNICITY_VAR_MAP.values())
```
zmbc, line 130: Why is this one a set while the others are lists? Doesn't this create an issue with depending on the set iteration order?

integration_tests/conftest.py
```diff
@@ -16,7 +16,7 @@ def sim() -> InteractiveContext:
     return simulation
 
 
-TIME_STEPS_TO_TEST = [0, 1, 10]
+TIME_STEPS_TO_TEST = list(range(10))
```
zmbc, line 19: Should this be in this PR? Maybe put it in #141?

zmbc (commented):
Glad we are making this change!

src/vivarium_census_prl_synth_pop/components/observers.py
```diff
@@ -380,19 +381,19 @@ def do_observation(self, event: Event) -> pd.DataFrame:
         # selection of additional simulants to reach the covered
         # probabilities
 
-        simplified_race_ethnicity = pop_u1["race_ethnicity"].copy()
-        simplified_race_ethnicity[
-            ~pop_u1["race_ethnicity"].isin(["Latino", "Black", "White"])
-        ] = "Other"
+        simplified_race_ethnicity = pop_u1["race_ethnicity"].astype(
+            pd.CategoricalDtype(categories=self.WIC_RACE_ETHNICITIES)
```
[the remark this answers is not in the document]
rmudambi: The order doesn't matter in this case. I believe it only really matters for sorting, which we aren't doing anywhere. 

If you attempt to add a value that is not valid per the categories it becomes NaN, so this is equivalent to the leftover category "Other"

src/vivarium_census_prl_synth_pop/components/population.py
rmudambi: I need to think a bit more about where a lot of these constants should live in general.

src/vivarium_census_prl_synth_pop/components/observers.py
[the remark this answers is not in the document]
zmbc: Ah okay, that's much less brittle than what I thought was happening 👍

src/vivarium_census_prl_synth_pop/constants/data_values.py
rmudambi: We don't happen to care about ordering anywhere we're using this, but you're right that if we were to care about ordering later and this was a random ordering it could lead to confusing bugs later. But yeah, dict ordering is deterministic as of 3.7 as you say so it doesn't really matter.

src/vivarium_census_prl_synth_pop/constants/metadata.py
rmudambi: It's a set because I needed to remove duplicates, and as I said above, we don't actually care about the order. That being said, consistency is good, and it's also better to know the order in case do care later

src/vivarium_census_prl_synth_pop/components/population.py
zmbc: Maybe for now you could do `.astype(pd.CategoricalDtype(categories=metadata.RELATIONSHIPS))` here? But, it's also okay how it is. Just took me a second to parse.

zmbc (approved):

src/vivarium_census_prl_synth_pop/components/population.py
stevebachmeier: We should definitely have a centralized schema approach. One option @rmudambi is to rip off what we were doing on the Covid ETL. I can look in more detail but we basically had a Schema class and you could specify dataset required columns (including their dtypes, nullability, etc). Something like:
```
output_schema = Schema(
    RequiredColumn("address_id", dtype=int, nullable=False),
```

And then there were soft and hard `check_constraint` methods attached to that Schema class

stevebachmeier (approved):

PR #141, opened 2022-12-31
reduce testing noise

rmudambi:
### Description

- *Category*: test
- *JIRA issue*: N/A
- *Research reference*: 

### Changes and notes
Use loops in tests that loop over time-steps rather than `pytest.mark.parameterize` so that if there's a failure it doesn't show up 10 times. These tests are such that it is highly likely that they will fail on every time-step if they fail on any.

### Verification and Testing
Ran integration tests on new test artifact `/mnt/team/simulation_science/priv/engineering/vivarium_census_prl_synth_pop/data/artifacts/categorical-test/florida.hdf`

### DIFF
--- a/integration_tests/test_employment.py
+++ b/integration_tests/test_employment.py
@@ -1,50 +1,46 @@
 import pytest
 
-from integration_tests.conftest import TIME_STEPS_TO_TEST
 from vivarium_census_prl_synth_pop.constants import data_values
 
 # TODO: Broader test coverage
 
+
 # TODO stop skipping once MIC-3703 has been implemented
-@pytest.mark.parametrize("time_step", TIME_STEPS_TO_TEST)
 @pytest.mark.skip
-def test_military_gq_employment(tracked_live_populations, time_step):
-    pop = tracked_live_populations[time_step]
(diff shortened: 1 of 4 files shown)


zmbc (approved):

integration_tests/test_employment.py
```diff
@@ -1,50 +1,46 @@
...
 
-from integration_tests.conftest import TIME_STEPS_TO_TEST
 from vivarium_census_prl_synth_pop.constants import data_values
 
 # TODO: Broader test coverage
 
+
 # TODO stop skipping once MIC-3703 has been implemented
-@pytest.mark.parametrize("time_step", TIME_STEPS_TO_TEST)
 @pytest.mark.skip
```
stevebachmeier, line 9: Not related to this PR, but why do we skip this test? We should probably add a `reason=`

stevebachmeier (approved):

integration_tests/test_employment.py
rmudambi: We skip because of the ticket in the comment above (MIC-3703)

PR #142, opened 2023-01-03
implement 'United States of America' location with California+Florida

stevebachmeier:
### Description
- *Category*: artifact; implementation
- *JIRA issue*: [MIC-3683](https://jira.ihme.washington.edu/browse/MIC-3683)
- *Research reference*: na

### Changes and notes
This implements a single national "United States of America" artifact
(currently includes on Florida + California).

Notes:
- I have commented out the synthetic_pii artifact data (which will be moved to post-processing
  and is currently not used at all in the codebase)
- We are now using a single value for health data (that of location
  'United States of America'); this is confirmed by the RT to be
  appropriate (https://ihme.slack.com/archives/C02KUQ9LX32/p1672352664332889)
- We are using a single value for workforce employed percentage
  (https://ihme.slack.com/archives/C02KUQ9LX32/p1672434298456879)

### Verification and Testing
The artifact is generated and a simulation ran just to ensure no runtime
errors (no v&v or profiling done yet - that's for next tickets)

### DIFF
--- a/src/vivarium_census_prl_synth_pop/data/loader.py
+++ b/src/vivarium_census_prl_synth_pop/data/loader.py
@@ -12,7 +12,8 @@ for an example.
 
    No logging is done here. Logging is done in vivarium inputs itself and forwarded.
 """
-from typing import Dict
+from pathlib import Path
+from typing import Dict, List
 
 import pandas as pd
 from vivarium.framework.artifact import EntityKey
@@ -49,9 +50,10 @@ def get_data(lookup_key: str, location: str) -> pd.DataFrame:
         data_keys.POPULATION.TMRLE: load_theoretical_minimum_risk_life_expectancy,
         data_keys.POPULATION.LOCATION: load_location,
         data_keys.POPULATION.ASFR: load_asfr,
-        data_keys.SYNTHETIC_DATA.LAST_NAMES: load_last_name_data,
-        data_keys.SYNTHETIC_DATA.FIRST_NAMES: load_first_name_data,
-        data_keys.SYNTHETIC_DATA.ADDRESSES: load_address_data,
+        # Comment out synthetic_data to be moved into different artifact
+        # data_keys.SYNTHETIC_DATA.LAST_NAMES: load_last_name_data,
+        # data_keys.SYNTHETIC_DATA.FIRST_NAMES: load_first_name_data,
+        # data_keys.SYNTHETIC_DATA.ADDRESSES: load_address_data,
     }
     return mapping[lookup_key](lookup_key, location)
 
@@ -59,7 +61,9 @@ def get_data(lookup_key: str, location: str) -> pd.DataFrame:
 def load_standard_data(key: str, location: str) -> pd.DataFrame:
     key = EntityKey(key)
     entity = get_entity(key)
-    data = interface.get_measure(entity, key.measure, location).droplevel("location")
+    data = interface.get_measure(entity, key.measure, location).droplevel(
+        "location"
+    )
     return data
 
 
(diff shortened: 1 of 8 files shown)


src/vivarium_census_prl_synth_pop/model_specifications/model_spec.yaml
```diff
@@ -37,4 +37,4 @@ configuration:
             day: 1
         step_size: 28 # Days
     population:
-        population_size: 2_000
+        population_size: 100_000
```
albrja, line 40: Is this something we want to change?  I don't have a problem with it personally because we seem to be finding more errors when we increase population size as it allows us to catch more edge cases.

src/vivarium_census_prl_synth_pop/data/loader.py
```diff
@@ -307,3 +292,33 @@ def create_draws(df: pd.DataFrame, key: str, location: str):
...
+    data["SERIALNO"] = data["SERIALNO"].astype(str)
+    data = data.rename(columns=column_map)
+
+    if location == "United States of America":
+        location_ids = [
+            v
+            for k, v in metadata.CENSUS_STATE_IDS.items()
+            if k in metadata.UNITED_STATES_LOCATIONS
+        ]
+        data = data.query(f"state in {location_ids}")
```
zmbc, line 318: I think in general we don't use `query`, but I see that you based this on existing code using `query` so it's not a big deal either way.

src/vivarium_census_prl_synth_pop/data/loader.py
```diff
@@ -307,3 +292,33 @@ def create_draws(df: pd.DataFrame, key: str, location: str):
...
+    elif location in metadata.CENSUS_STATE_IDS.keys():
```
zmbc, line 319: This can be just:

```suggestion
    elif location in metadata.CENSUS_STATE_IDS:
```

but I don't know if that is preferable.

zmbc (approved):

src/vivarium_census_prl_synth_pop/model_specifications/model_spec.yaml
stevebachmeier: Uhh, not intentionally - I just forgot to change it back

src/vivarium_census_prl_synth_pop/data/loader.py
stevebachmeier: Yeah I personally hate `query` and initially wrote this with standard slicing. But then I switched for consistency

src/vivarium_census_prl_synth_pop/data/loader.py
stevebachmeier: it is - I _always_ forget this

albrja (approved):

src/vivarium_census_prl_synth_pop/data/loader.py
rmudambi: It's fine not to use `query`, btw.

src/vivarium_census_prl_synth_pop/data/loader.py
```diff
@@ -189,7 +172,9 @@ def load_asfr(key: str, location: str):
 
     # pivot
     asfr = asfr.reset_index()
-    asfr = asfr[(asfr.year_start == 2019)]  # NOTE: this is the latest year available from GBD
+    asfr = asfr[
+        (asfr["year_start"] == 2019)
+    ]  # NOTE: this is the latest year available from GBD
```
rmudambi, line 177: If you move this comment onto its own line black won't vandalize this line

rmudambi (approved):

PR #144, opened 2023-01-04
Handle edge case in new household migration test

zmbc:
### Description
- *Category*: test
- *JIRA issue*: [MIC-3749](https://jira.ihme.washington.edu/browse/MIC-3749)
- *Research reference*: none

### Verification and Testing

Tests are passing even with 500k population.


integration_tests/test_domestic_migration.py
```diff
@@ -16,12 +16,39 @@ def test_individuals_move(simulants_on_adjacent_timesteps):
...
+        # There is exactly one new-household mover for each new household.
+        assert (
+            new_household_movers[movers_into_new_households]
+            .groupby(after[movers_into_new_households]["household_id"])
+            .sum()
+            == 1
+        ).all()
+        # These should be the vast majority, since non-reference-person moves to
+        # new households will be quite rare.
+        assert new_household_movers.sum() / movers_into_new_households.sum() > 0.95
```
stevebachmeier, line 51: I like the sentiment here, but is 95% a reasonable number?

stevebachmeier (approved):

albrja (approved):

integration_tests/test_domestic_migration.py
rmudambi: I'd think this number should scale with the size of the sim population. If we have fewer than 20 new households formed, a single person moving into one of these new households will cause this test to fail. We really don't want false positives in our tests, since CRNs means we'll fail every time if we fail at all. 

That being said, I don't think it's important to think too hard about what the number should be in this case. If this passes on the sizes we're using, that's fine. We're really just checking that we don't have everyone moving into these new households for some reason.

rmudambi (approved):

integration_tests/test_domestic_migration.py
zmbc: @rmudambi Good point, I didn't think about small population sizes. I think that in general we should consider 20k a minimum -- for very small sample sizes, I think there are other tests that also fail.

PR #146, opened 2023-01-05
Fix data type for survey_date

zmbc:
### Description

- *Category*: bugfix
- *JIRA issue*: none
- *Research reference*: none

### Changes and notes

We store dates/times in our population table and in our observers as datetimes. The one exception was survey_date, and this was causing an error trying to save to tabular HDF.

### Verification and Testing

The simulation did not run to completion before with a 20k population (larger population was necessary to have anyone returned by the survey observers). Now it does.

### DIFF
--- a/tests/components/observers/test_household_survey_observers.py
+++ b/tests/components/observers/test_household_survey_observers.py
@@ -62,7 +62,7 @@ def test_do_observation(observer, mocked_pop_view, mocker):
     expected = mocked_pop_view
     expected["middle_initial"] = ["J", "M"]
     expected[["guardian_1_address_id", "guardian_2_address_id"]] = np.nan
-    expected["survey_date"] = [pd.to_datetime(sim_start_date).date()] * 2
+    expected["survey_date"] = [pd.to_datetime(sim_start_date).floor("D")] * 2
 
     pd.testing.assert_frame_equal(expected[observer.output_columns], observation)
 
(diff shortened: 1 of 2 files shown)


tests/components/observers/test_household_survey_observers.py
```diff
@@ -62,7 +62,7 @@ def test_do_observation(observer, mocked_pop_view, mocker):
     expected = mocked_pop_view
     expected["middle_initial"] = ["J", "M"]
     expected[["guardian_1_address_id", "guardian_2_address_id"]] = np.nan
-    expected["survey_date"] = [pd.to_datetime(sim_start_date).date()] * 2
+    expected["survey_date"] = [pd.to_datetime(sim_start_date).floor("D")] * 2
```
stevebachmeier, line 65: Ew. What is this doing? Flooring to midnight of that date? If so, do you even need anything or would just deleting the `date()` work?

stevebachmeier (approved):

tests/components/observers/test_household_survey_observers.py
zmbc: Yes, `.floor("D")` floors to midnight.

In this particular spot I don't need it because we just defined that it is a round date; I will remove. But in the other places in this PR, we only can guarantee `event.time` is a round date if step size is a round number, which I think in practice is always true but seems risky to depend on.

rmudambi: It is true in this simulation (28 days), but we have had simulations like the MNCH child model and the CIFF model with a time-step of 0.5 days. 

That said, I'm always suspicious of test code that looks too much like the code it is testing. In the test, I think we only care that the date part of the datetime matches the expected date

rmudambi (approved):

PR #150, opened 2023-01-07
Assigns new reference persons and adds test

albrja:
### When a household loses a reference person on time step, assign the oldest living member of that household as the new reference person.
- *Category*: Feature
- *JIRA issue*: [MIC-3753](https://jira.ihme.washington.edu/browse/MIC-3753)
- *Research reference*: https://vivarium-research.readthedocs.io/en/latest/models/concept_models/vivarium_census_synthdata/concept_model.html#component-7-mortality

### Changes and notes
-Assigns oldest member of households without a reference person as the new reference person for that household.
-Adds a test that the oldest member is assigned as the reference person and now utilizes an existing test that was previously skipped that verifies every standard household has a reference person.
-Note: This is assigning a new reference person for these selected households but does not update the relation_to_household_head for other members yet.  

### Verification and Testing
Successfully ran simulation and new reference persons were properly updated.

### DIFF
--- a/integration_tests/test_household_structure.py
+++ b/integration_tests/test_household_structure.py
@@ -54,51 +54,40 @@ def test_household_id_and_address_id_correspond(tracked_live_populations):
     # (over multiple time steps) when the whole house moved as a unit between those time steps.
 
 
-def test_new_reference_person_is_oldest_household_member(tracked_live_populations):
-    for time_step in range(max(TIME_STEPS_TO_TEST)):
-        before_reference_person_idx = (
-            tracked_live_populations[time_step]
-            .loc[
-                tracked_live_populations[time_step]["relation_to_household_head"]
-                == "Reference person"
-            ]
-            .index
-        )
-        after_reference_person_idx = (
-            tracked_live_populations[time_step + 1]
-            .loc[
-                tracked_live_populations[time_step + 1]["relation_to_household_head"]
-                == "Reference person"
-            ]
-            .index
-        )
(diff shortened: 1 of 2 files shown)


integration_tests/test_household_structure.py
```diff
@@ -54,3 +52,53 @@ def test_household_id_and_address_id_correspond(tracked_live_populations):
...
+        )
+        new_reference_person_idx = np.setdiff1d(
+            after_reference_person_idx, before_reference_person_idx
+        )
+
+        # Get households with new reference persons
+        household_ids_with_new_reference_person = tracked_live_populations[time_step + 1].loc[
+            new_reference_person_idx, "household_id"
+        ]
+        assert len(household_ids_with_new_reference_person) == len(
```
rmudambi, line 83: This could only fail if there was a household with multiple reference people in it, right? I think we have a test somewhere else that tests that (if we don't we should and it should be its own test).

integration_tests/test_household_structure.py
```diff
@@ -54,3 +52,53 @@ def test_household_id_and_address_id_correspond(tracked_live_populations):
...
+            household_ids_with_new_reference_person.unique()
+        )
+
+        households_with_new_reference_person_idx = (
+            tracked_live_populations[time_step + 1]
+            .loc[
```
rmudambi, line 89: Use `.index[...]` rather than `.loc[...].index`

rmudambi (changes requested):
Looks good. Just a few comments to address in the tests

rmudambi (approved):

PR #152, opened 2023-01-10
implement domestic move state/puma

stevebachmeier:
### Description
- *Category*: Implementation
- *JIRA issue*: [MIC-3728](https://jira.ihme.washington.edu/browse/MIC-3728)
- *Research reference*: 
  - [Household moves](https://vivarium-research.readthedocs.io/en/latest/models/concept_models/vivarium_census_synthdata/concept_model.html#household-moves)
  - [Business moves](https://vivarium-research.readthedocs.io/en/latest/models/concept_models/vivarium_census_synthdata/concept_model.html#business-address-changes)
  - [Individual moves](https://vivarium-research.readthedocs.io/en/latest/models/concept_models/vivarium_census_synthdata/concept_model.html#individual-moves)

### Changes and notes
Apologies in advance - this probably should have been broken into smaller tasks.

The meat of this is the `utilities.update_addresses` function, which
includes the (already-existing) logic to update address_ids as well as the
new logic to update states and pumas. Note that two entirely new columns
are now added to the state table: "employer_state" and "employer_puma".

What's particularly important to remember is that for this task, we're not
quite doing what is requested by the docs. Specifically, we are not using a
dataset to assign state/pumas upon movement. Instead, we are uniformly
sampling from all existing pumas in the artifact data. This was agreed
upon as a good first cut at domestic travel and a different ticket exists
to improve the domestic migration patterns.

A couple edge notes I clarified with the RT:
- Households and businesses are free to move within the same state/puma
- Individuals moving into a GQ have state/puma re-sampled. This effectively
  "grows" the locations of jails, nursing homes, etc.

Finally: If this PR gets implmented as-is, I intend on creating a followup task
to make this far more readable and possibly less fragile; we should maintain
an address_id-to-state/puma mapping throughout the sim to ease in
adding new addresses, extracting the relevant mappings, etc.

### Verification and Testing
All tests pass with 200k population size

### DIFF
--- a/integration_tests/test_domestic_migration.py
+++ b/integration_tests/test_domestic_migration.py
@@ -126,14 +126,11 @@ def test_only_living_people_move(simulants_on_adjacent_timesteps):
         assert (after[movers]["alive"] == "alive").all()
 
 
-def test_address_ids_states_pumas(simulants_on_adjacent_timesteps, sim) -> None:
-    """Check that address_ids are mapping correctly to states/pumas. Note that
+def test_addresses_during_moves(simulants_on_adjacent_timesteps) -> None:
+    """Check that addresses are handled properly during moves. Note that
     this test does not distinguish between the different types of moves (eg
     household moves, business moves, individual moves, etc)
     """
-    state_puma_map = get_state_puma_map(
-        sim._data.artifact.load("population.households").reset_index()
-    )
     for before, after in simulants_on_adjacent_timesteps:
         mask_movers = before["address_id"] != after["address_id"]
         # TODO: Check that *most* are in a new state when we add all locations
@@ -157,30 +154,24 @@ def test_address_ids_states_pumas(simulants_on_adjacent_timesteps, sim) -> None:
             after.loc[same_non_gq_addresses, ["state", "puma"]],
         )
 
-        # Ensure pumas map to correct state. This will be an imperfect test
-        # because pumas are only unique within states and so one puma can
-        # exist in multiple states. Here we only ensure no impossible pumas
-        # exist in each state
-        for (state, puma) in before[["state", "puma"]].drop_duplicates().values:
-            assert puma in state_puma_map[state]
-        for (state, puma) in after[["state", "puma"]].drop_duplicates().values:
+
+def test_address_uniqueness(populations, sim):
+    # Ensure pumas map to correct state. This will be an imperfect test
+    # because pumas are only unique within states and so one puma can
+    # exist in multiple states. Here we only ensure no impossible pumas
+    # exist in each state
+    state_puma_map = get_state_puma_map(
+        sim._data.artifact.load("population.households").reset_index()
+    )
+    for pop in populations:
+        for (state, puma) in pop[["state", "puma"]].drop_duplicates().values:
             assert puma in state_puma_map[state]
 
         # Each unique address_id has the same puma/state (except for GQ)
         assert (
             (
-                before[before["housing_type"] == "Standard"]
-                .groupby("address_id")["state", "puma"]
-                .nunique()
-                == 1
-            )
-            .all()
-            .all()
-        )
-        assert (
-            (
-                after[after["housing_type"] == "Standard"]
-                .groupby("address_id")["state", "puma"]
+                pop[pop["housing_type"] == "Standard"]
+                .groupby("address_id")[["state", "puma"]]
                 .nunique()
                 == 1
             )
--- a/integration_tests/test_domestic_migration.py
+++ b/integration_tests/test_domestic_migration.py
@@ -1,5 +1,3 @@
-from typing import Dict
-
 import pandas as pd
 
 from vivarium_census_prl_synth_pop.constants import data_values
@@ -61,8 +59,6 @@ def test_individuals_move_into_new_households(simulants_on_adjacent_timesteps):
         new_addresses = after[new_household_movers]["address_id"]
         assert new_addresses.nunique() == len(new_addresses)
         assert not new_addresses.isin(before["address_id"]).any()
-
-        # TODO: Add state/puma tests
 
 
 def test_individuals_move_into_group_quarters(simulants_on_adjacent_timesteps):
--- a/integration_tests/test_domestic_migration.py
+++ b/integration_tests/test_domestic_migration.py
@@ -1,6 +1,9 @@
-import pytest
+from typing import Dict
+
+import pandas as pd
 
 from vivarium_census_prl_synth_pop.constants import data_values
+from vivarium_census_prl_synth_pop.utilities import get_state_puma_map
 
 # TODO: Broader test coverage
 
@@ -59,6 +62,8 @@ def test_individuals_move_into_new_households(simulants_on_adjacent_timesteps):
         assert new_addresses.nunique() == len(new_addresses)
         assert not new_addresses.isin(before["address_id"]).any()
 
+        # TODO: Add state/puma tests
+
 
 def test_individuals_move_into_group_quarters(simulants_on_adjacent_timesteps):
     for before, after in simulants_on_adjacent_timesteps:
@@ -114,7 +119,7 @@ def test_households_move(simulants_on_adjacent_timesteps):
         assert new_addresses.nunique() == len(moved_households)
         assert not before["address_id"].isin(new_addresses).any()
 
-        # Never GQ households
+        # GQ households never move
         assert (before[household_movers]["housing_type"] == "Standard").all()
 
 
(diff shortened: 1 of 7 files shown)


integration_tests/test_domestic_migration.py
```diff
@@ -123,3 +124,66 @@ def test_only_living_people_move(simulants_on_adjacent_timesteps):
...
+        same_addresses = before[before["address_id"] == after["address_id"]].index
+        gq_simulants = after[after["housing_type"] != "Standard"].index
+        same_non_gq_addresses = same_addresses.difference(gq_simulants)
+
+        pd.testing.assert_frame_equal(
+            before.loc[same_non_gq_addresses, ["state", "puma"]],
+            after.loc[same_non_gq_addresses, ["state", "puma"]],
+        )
+
+        # Ensure pumas map to correct state. This will be an imperfect test
```
zmbc, line 160: The rest of this test could be on each timestep's population instead of being on simulants_on_adjacent_timesteps. There is an existing test for address_ids that is similar.

src/vivarium_census_prl_synth_pop/components/businesses.py
```diff
@@ -175,27 +183,27 @@ def on_time_step(self, event: Event):
             query="alive == 'alive'",
         )
         all_businesses = self.businesses.loc[
-            self.businesses["employer_id"] != data_values.UNEMPLOYED_ID, "employer_id"
-        ]
+            self.businesses.index != data_values.UNEMPLOYED_ID
+        ].index
```
zmbc, line 187: This can be `self.businesses.index[...]` (no `loc` required).

src/vivarium_census_prl_synth_pop/components/businesses.py
```diff
@@ -342,10 +359,10 @@ def _update_employer_metadata(
         employer_ids = pop.loc[rows_to_update, "employer_id"]
 
         pop.loc[rows_to_update, "employer_address_id"] = employer_ids.map(
-            self.businesses.set_index("employer_id")["employer_address_id"].to_dict()
+            self.businesses["employer_address_id"].to_dict()
```
zmbc, line 362: I know you didn't introduce this, but I'm pretty sure `.to_dict()` is not necessary here.

src/vivarium_census_prl_synth_pop/components/person.py
```diff
@@ -119,7 +126,10 @@ def on_time_step(self, event: Event) -> None:
             event.index,
             query="alive == 'alive'",
         )
-
+        # Want all addresses, including from dead simulants
+        address_id_to_state_puma_map = self._get_non_gq_address_id_to_state_puma_mapping(
+            self.population_view.get(event.index)
```
zmbc, line 131: I don't know Vivarium well enough to know whether it matters, but would it be better here to call `self.population_view.get(...)` only once and do the alive filtering outside of `query`?

src/vivarium_census_prl_synth_pop/components/person.py
```diff
@@ -146,33 +163,50 @@ def on_time_step(self, event: Event) -> None:
     ##################
 
     def _perform_new_household_moves(
-        self, pop: pd.DataFrame, movers: pd.Index
+        self, pop: pd.DataFrame, movers: pd.Index, address_mapping: pd.DataFrame
     ) -> pd.DataFrame:
         """
         Create a new single-person household for each person in movers and move them to it.
         """
+        pop0 = pop.copy()
```
zmbc, line 171: Is this used?

src/vivarium_census_prl_synth_pop/components/person.py
```diff
@@ -255,8 +299,13 @@ def _perform_non_reference_person_moves(
 
         pop.loc[movers, "household_id"] = household_id_values
         pop.loc[movers, "housing_type"] = "Standard"
-        pop.loc[movers, "address_id"] = household_id_values.map(non_gq_address_ids)
         pop.loc[movers, "relation_to_household_head"] = "Other nonrelative"
+        # Update address and enforce address_id-state/puma link
+        pop.loc[movers, "address_id"] = household_id_values.map(non_gq_address_ids)
+        pop = pop.reset_index().set_index("address_id")
```
zmbc, line 305: I feel like there must be a simpler and less scary way to do this -- does this work?

```
pop.loc[movers, ["state", "puma"]] = pop.loc[movers, "address_id"].map(address_mapping)
```

src/vivarium_census_prl_synth_pop/components/population.py
```diff
@@ -118,13 +118,15 @@ def generate_initial_population(self, pop_data: SimulantData) -> None:
         pop["housing_type"] = pop["housing_type"].astype(
             pd.CategoricalDtype(categories=data_values.HOUSING_TYPES)
         )
+        # set int dtypes to float64 to play nicely with population_view checks
```
zmbc, line 121: Maybe I just lack context, but what do you mean by this? It seems a bit odd to store state and puma as floats.

src/vivarium_census_prl_synth_pop/utilities.py
```diff
@@ -297,89 +267,164 @@ def build_output_dir(output_dir: Path, subdir: Optional[Union[str, Path]] = None
     return output_dir
 
 
+def get_state_puma_map(df: pd.DataFrame) -> Dict[int, int]:
+    return df.groupby("state")["puma"].apply(lambda x: list(x.unique())).to_dict()
```
zmbc, line 271:
```suggestion
    return df.groupby("state")["puma"].unique()
```

Do you really need it to be a dict?

src/vivarium_census_prl_synth_pop/utilities.py
```diff
@@ -297,89 +267,164 @@ def build_output_dir(output_dir: Path, subdir: Optional[Union[str, Path]] = None
...
+    """
+    state_puma_options = []
+    for state in state_puma_map:
+        state_puma_options.extend([(state, puma) for puma in state_puma_map[state]])
+    state_puma_choices = pd.DataFrame(
+        vectorized_choice(
+            options=np.array(state_puma_options, dtype="i,i"),
+            n_to_choose=len(units_that_move_ids),
+            randomness_stream=randomness,
+            additional_key="sampling_state_puma",
```
zmbc, line 316: Since you use this in multiple places, won't it use the same randomness each time? I think you might need to allow passing an `additional_key` to this method.

src/vivarium_census_prl_synth_pop/utilities.py
```diff
@@ -297,89 +267,164 @@ def build_output_dir(output_dir: Path, subdir: Optional[Union[str, Path]] = None
...
+        ),
+        index=units_that_move_ids,
+    )
+    state_puma_choices.columns = [state_col_name, puma_col_name]
+    units.loc[
+        units_that_move_ids, [state_col_name, puma_col_name]
+    ] = state_puma_choices.astype(int)
```
zmbc, line 323: Why is both this and the `dtype="i,i"` above necessary?

src/vivarium_census_prl_synth_pop/utilities.py
```diff
@@ -297,89 +267,164 @@ def build_output_dir(output_dir: Path, subdir: Optional[Union[str, Path]] = None
...
+            randomness=randomness,
+        )
+        starting_address_id += len(movers_idx)
+        return df, starting_address_id
 
-        # update address_id column in the pop table
-        rows_changing_address_id_idx = pop.loc[
-            pop[unit_id_col_name].isin(units_that_move_ids)
-        ].index
+    pop0 = pop.copy()
```
zmbc, line 388: Is this used?

src/vivarium_census_prl_synth_pop/utilities.py
```diff
@@ -297,89 +267,164 @@ def build_output_dir(output_dir: Path, subdir: Optional[Union[str, Path]] = None
...
     """
 
-    if len(units_that_move_ids) > 0:
-        # update the employer address_id in self.businesses
-        # this will update address_id in a households dataframe we are not tracking like we are businesses
-        # todo:  Should add income/salary to mapping function here
-        moving_units = update_address_id(
-            df=moving_units,
-            rows_to_update=units_that_move_ids,
+    def _update_address_id_state_puma(
```
zmbc, line 368: Let's move this outside the method.

src/vivarium_census_prl_synth_pop/components/person.py
stevebachmeier: Yeah, probably, good call

src/vivarium_census_prl_synth_pop/components/person.py
albrja: Can you map to two columns like this?

src/vivarium_census_prl_synth_pop/components/population.py
stevebachmeier: As discussed, it's because the population_view.update method breaks because they somehow get converted to floats while processing. I'll likely put this back to integers (which it really should be) and then coerce to floats where appropriate.

All the more reason to implement a more robust schema system

src/vivarium_census_prl_synth_pop/utilities.py
stevebachmeier: I can work with something else. What's the issue with dict?

It looks like {state: [pumas]}, eg `{6: [10702, 6514, 3759, ...], 12: [9507, 5301, 5704, ...]}`

src/vivarium_census_prl_synth_pop/components/person.py
zmbc: I don't know, I didn't test it -- but even mapping them one at a time would be preferable to the current method, IMO

src/vivarium_census_prl_synth_pop/utilities.py
zmbc: No issue with dict, but I don't see a reason to prefer it over a pandas Series, which is a datatype more commonly used in Vivarium. The suggestion above makes a pandas Series.

zmbc: And if you do prefer dict, I believe you could still use my suggestion with a `.to_dict()` at the end.

src/vivarium_census_prl_synth_pop/components/person.py
stevebachmeier: Zeb, I tried so many things here and couldn't get it to work. I'll poke around a bit more. What's scary, though? All the resetting of index?

Jim, yep! Though there are some "rules" you must follow.

zmbc: Scary: resetting of index, and not limiting what you change to the `movers`. Plus, it's a bit long/hard to follow.

stevebachmeier: @zmbc The reason your proposal doesn't work is b/c it's trying to map different indexes (left side is simulant, right side is address_id). That's why I go through the process of setting address_id as the index, then mapping, then setting back to simulant index.

I believe you can map from column but that is somehow scarier to me (there's no specific `left_on` and `right_on` like in `merge`). And I guess I could use merge but that seems clunky

stevebachmeier: Oh, duh

```
pop.loc[movers, "state"] = pop["address_id"].map(address_mapping["state"])
pop.loc[movers, "puma"] = pop["address_id"].map(address_mapping["puma"])
```

src/vivarium_census_prl_synth_pop/utilities.py
stevebachmeier: Good enough for me, will work on it

src/vivarium_census_prl_synth_pop/utilities.py
stevebachmeier: This was lingering from before I changed to dtype float. Something must be further changing this to float after the fact. I'll clean it up

stevebachmeier:
I'm closing this while I work on a refactor

PR #154, opened 2023-01-11
BUG: fix defects in income

aflaxman:
Only living simulants should get paid each time step, and dependent status should be based on last year's earnings not the annual salary for the simulant's current job.

Thanks to @zmbc for catching this.


src/vivarium_census_prl_synth_pop/components/observers.py
```diff
@@ -610,8 +610,6 @@ def get_observation(self, event: Event) -> pd.DataFrame:
             df_w2[col] = df_w2["simulant_id"].map(pop_full[col])
 
         # Tracked, US population to be dependents or get their SSNs borrowed
-        # TODO: This should only be the living population -- leaving for now because
```
zmbc, line 613: So we are leaving dead simulants listed as dependents, and they are eligible to have their SSNs borrowed?

src/vivarium_census_prl_synth_pop/components/observers.py
```diff
@@ -644,6 +642,11 @@ def get_observation(self, event: Event) -> pd.DataFrame:
         )
 
         # not all simulants with a guardian are eligible to be dependents
+        # need to know earnings of dependents in past year for this
+        dependent_w2 = df_w2[df_w2["simulant_id"].isin(df_dependents.index.unique())]
+        dependent_wages = dependent_w2.groupby("simulant_id")["income"].sum()
+        df_dependents["last_year_income"] = df_dependents.index.map(dependent_wages)
```
zmbc, line 648: Couldn't this just be `= dependent_wages`?

Also, the use of "earnings," "income" and "wages" interchangeably is confusing. This variable could be called "dependent_last_year_income". The comment above should say "income".

What happens here when there is no income for the dependent? I'd expect NaN, which would cause them to not qualify below if they are over 19.

src/vivarium_census_prl_synth_pop/components/observers.py
```diff
@@ -652,10 +655,10 @@ def get_observation(self, event: Event) -> pd.DataFrame:
             | (
                 (df_dependents["age"] < 24)
                 & (df_dependents["housing_type"] == "College")
-                & (df_dependents["income"] < 10_000)
+                & (df_dependents["last_year_income"] < 10_000)
```
rmudambi, line 658: Refresh my memory - are we recording all possible dependents here, or the dependents they are actually claiming? A person could be eligible to be claimed by multiple filers.

aflaxman: I think it is best to think of these as "eligible dependents", maybe I will change the var name to be clearer.  It is already a bit verbose, though.

src/vivarium_census_prl_synth_pop/components/observers.py
aflaxman: oops, no, my mistake.

src/vivarium_census_prl_synth_pop/components/observers.py
aflaxman: all great points. I have attempted to address them in this next commit.

rmudambi (approved):

PR #155, opened 2023-01-11
create simple post-processing cli stub

rmudambi:
### Description

- *Category*: post-processing
- *JIRA issue*: [MIC-3734](https://jira.ihme.washington.edu/browse/MIC-3734)
- *Research reference*: 

### Changes and notes
Created a stub for post-processing that copies existing outputs to final output location 

### Verification and Testing
Tested using `/mnt/team/simulation_science/priv/engineering/vivarium_census_prl_synth_pop/results/vv_observers/popsize_2_500_000/florida/2022_12_23_12_57_17`

Note: I had to revert the change to the raw output location to be compatible with the above run, but this will work with all future runs.

### DIFF
--- a/src/vivarium_census_prl_synth_pop/tools/cli.py
+++ b/src/vivarium_census_prl_synth_pop/tools/cli.py
@@ -57,7 +57,7 @@ def make_artifacts(
 
 
 @click.command()
-@click.argument("output_file", type=click.Path(exists=True))
+@click.argument("output_dir", type=click.Path(exists=True))
 @click.option("-v", "verbose", count=True, help="Configure logging verbosity.")
 @click.option(
     "--pdb",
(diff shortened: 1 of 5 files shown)


stevebachmeier (approved):

src/vivarium_census_prl_synth_pop/tools/cli.py
```diff
@@ -57,25 +57,15 @@ def make_artifacts(
...
     help="Drop into python debugger if an error occurs.",
 )
-@click.option(
-    "-s",
-    "--single",
-    "single_run",
-    default=False,
-    is_flag=True,
-    help="Results are from a single, non-parallel run.",
-)
```
zmbc, line 75: Is this removed because we are not planning to support parallel runs in the future, or just because they don't work _yet_ due to the observer overwriting issue? If the latter, would it make sense to leave this and raise a NotImplementedError or something in build_results?

zmbc (approved):

src/vivarium_census_prl_synth_pop/tools/cli.py
rmudambi: This is removed because I don't want the output of a single sim to be any different from a parallel sim that only has a single seed.

zmbc: That makes sense, but what about the output of a parallel sim with _multiple_ seeds? That will need to be handled here eventually, right?

rmudambi: This currently is broken in the simulaton (since each parallel run tries to write to the same location). That needs to be fixed and I think a ticket has bee created for that. When that is fixed (as part of that ticket), this post-processing work will have to be updated.

PR #156, opened 2023-01-12
add ability to mark best and latest versions

rmudambi:
### Description

- *Category*: post-processing
- *JIRA issue*: [MIC-3759](https://jira.ihme.washington.edu/browse/MIC-3759)
- *Research reference*: 

### Changes and notes
 Add ability to mark post-processing versions as best and latest. Also adds the concept of a "test run" which will mark neither.

### Verification and Testing
Tested running multiple times with each flag combination and verified that it works as expected.

### DIFF
--- a/src/vivarium_census_prl_synth_pop/tools/cli.py
+++ b/src/vivarium_census_prl_synth_pop/tools/cli.py
@@ -65,7 +65,19 @@ def make_artifacts(
     is_flag=True,
     help="Drop into python debugger if an error occurs.",
 )
-def make_results(output_dir: str, verbose: int, with_debugger: bool) -> None:
+@click.option(
+    "-b",
+    "--mark-best",
+    is_flag=True,
+)
+@click.option(
+    "-t",
+    "--test-run",
+    is_flag=True,
+)
+def make_results(
+    output_dir: str, verbose: int, with_debugger: bool, mark_best: bool, test_run: bool
+) -> None:
     configure_logging_to_terminal(verbose)
     main = handle_exceptions(build_results, logger, with_debugger=with_debugger)
-    main(output_dir)
+    main(output_dir, mark_best, test_run)
--- a/src/vivarium_census_prl_synth_pop/tools/make_results.py
+++ b/src/vivarium_census_prl_synth_pop/tools/make_results.py
@@ -8,12 +8,36 @@ from vivarium_census_prl_synth_pop import utilities
 from vivarium_census_prl_synth_pop.constants import paths
 
 
-def build_results(results_dir: str) -> None:
+def build_results(results_dir: str, mark_best: bool, test_run: bool) -> None:
+    if mark_best and test_run:
+        logger.error(
+            "A test run can't be marked best. "
+            "Please remove either the mark best or the test run flag."
+        )
+        return
+
     logger.info("Creating final results directory.")
-    final_output_dir = utilities.build_output_dir(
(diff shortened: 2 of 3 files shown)


src/vivarium_census_prl_synth_pop/tools/cli.py
```diff
@@ -65,7 +65,19 @@ def make_artifacts(
     is_flag=True,
     help="Drop into python debugger if an error occurs.",
 )
-def make_results(output_dir: str, verbose: int, with_debugger: bool) -> None:
+@click.option(
+    "-b",
+    "--mark-best",
+    is_flag=True,
```
stevebachmeier, line 71: I think `is_flag`s default to True. Probably want to add default=False

src/vivarium_census_prl_synth_pop/tools/cli.py
```diff
@@ -65,7 +65,19 @@ def make_artifacts(
...
+)
+@click.option(
+    "-t",
+    "--test-run",
+    is_flag=True,
```
stevebachmeier, line 76: Same here - want default to be False I'd imagine

src/vivarium_census_prl_synth_pop/tools/make_results.py
```diff
@@ -8,12 +8,36 @@
...
-    ) / datetime.now().strftime("%Y_%m_%d_%H_%M_%S")
+    )
+    final_output_dir = final_output_root_dir / datetime.now().strftime("%Y_%m_%d_%H_%M_%S")
 
     logger.info("Copying raw results to final location.")
     raw_output_dir = Path(results_dir) / paths.RAW_RESULTS_DIR_NAME
     shutil.copytree(raw_output_dir, final_output_dir)
+
+    if test_run:
+        logger.info("Test run - not marking results as latest.")
```
stevebachmeier, line 30: Do we want test runs to default to the usual output dir? I wonder if we specifically DON'T want that.

Or rather...what are you envisioning a "test run" is used for?

stevebachmeier (changes requested):

src/vivarium_census_prl_synth_pop/tools/make_results.py
albrja: + 1.  I usually save test runs to my own folder.  I do like the flexibility to do a run and save it to the team folder and not mark it as latest or best.

stevebachmeier: Some clarifications
1. Latest happens automatically. You don't do anything for that.
2. I agree we need to be able to save to team folder and not mark best. That's exactly what `make_results` would do by default. But I'm wondering in what scenario do we want to `make_results -t` to the official folder?

rmudambi: A test run is a run that outputs to the same location, but doesn't mark latest. That's all. Maybe we never use it.

stevebachmeier (approved):

src/vivarium_census_prl_synth_pop/tools/cli.py
rmudambi: Nope it defaults to False

src/vivarium_census_prl_synth_pop/tools/make_results.py
stevebachmeier: oh, i see now. I guess that's just a fundamentally different mental model than what I had in mind of what "latest" meant.

PR #158, opened 2023-01-13
Feature/update member relationships

albrja:
### Updates relation to household head for member of households that were assigned a new reference person.
- *Category*: Feature
- *JIRA issue*: [MIC-3754](https://jira.ihme.washington.edu/browse/MIC-3754)
- *Research reference*: https://vivarium-research.readthedocs.io/en/latest/models/concept_models/vivarium_census_synthdata/concept_model.html#component-7-mortality

### Changes and notes
-Adds new relationships CSV to map new relationships.
-Updates relationships for all members in households that got a new reference person
-Adds test that verifies mapping is working on dummy population.

### Verification and Testing
Successfully ran simulation and all tests passed.


src/vivarium_census_prl_synth_pop/components/population.py
```diff
@@ -956,3 +943,98 @@ def assign_last_name_ids(self, pop: pd.DataFrame) -> pd.Series:
...
+            new_reference_persons, ["household_id", "relation_to_household_head"]
+        ]
+        population.loc[
+            new_reference_persons, "relation_to_household_head"
+        ] = "Reference person"
+
+        # Update simulants born in simulation as biological children to their mother if mother is new reference person
+        biological_children_idx = population.index[
+            (population["date_of_birth"] > self.start_time)
+            & (
```
rmudambi, line 983: Consider storing these in intermediate variables so black doesn't make it illegible.

We should revisit our black settings to try to avoid things like this.

src/vivarium_census_prl_synth_pop/components/population.py
```diff
@@ -956,3 +943,98 @@ def assign_last_name_ids(self, pop: pd.DataFrame) -> pd.Series:
...
+            population["household_id"].isin(household_ids_without_reference_person)
+        ]
+
+        # Find oldest member in each household and make them new reference person
+        # This is a series with household_id as the index and the new reference person as the value
+        new_reference_persons = (
+            population.loc[households_to_update_idx].groupby(["household_id"])["age"].idxmax()
+        )
+        # Preserve old relationship of new reference person before assigning them as new reference persons
+        new_reference_persons_relationship_to_old_reference_person = population.loc[
```
rmudambi, line 973: Can you come up with a shorter variable name? Maybe just `new_reference_person_prev_relationship`?

src/vivarium_census_prl_synth_pop/components/population.py
```diff
@@ -956,3 +943,98 @@ def assign_last_name_ids(self, pop: pd.DataFrame) -> pd.Series:
...
+        # Update simulants born in simulation as biological children to their mother if mother is new reference person
+        biological_children_idx = population.index[
+            (population["date_of_birth"] > self.start_time)
+            & (
+                population["household_id"].isin(
+                    new_reference_persons_relationship_to_old_reference_person["household_id"]
+                )
+            )
+            & (
+                population["guardian_1"]
```
rmudambi, line 989: Couldn't guardian_2 have become the new reference person? Are we ignoring this because you can only have a guardian 2 if guardian 1 was previously the reference person, and therefore your relationship won't change?

src/vivarium_census_prl_synth_pop/model_specifications/model_spec.yaml
```diff
@@ -38,4 +38,4 @@ configuration:
             day: 1
         step_size: 28 # Days
     population:
-        population_size: 2_000
+        population_size: 250_000
```
zmbc, line 41: Probably not intentional?

src/vivarium_census_prl_synth_pop/components/population.py
```diff
@@ -956,3 +943,98 @@ def assign_last_name_ids(self, pop: pd.DataFrame) -> pd.Series:
...
+                    new_reference_persons_relationship_to_old_reference_person[
+                        "relation_to_household_head"
+                    ]
+                    == relationship,
+                    "household_id",
+                ]
+            )
+            simulants_to_update_idx = (
+                population.index[population["household_id"].isin(household_ids_to_update)]
+                .difference(new_reference_persons_relationship_to_old_reference_person.index)
```
zmbc, line 1025: How does this differ from the `new_reference_persons` variable?

src/vivarium_census_prl_synth_pop/components/population.py
```diff
@@ -956,3 +943,98 @@ def assign_last_name_ids(self, pop: pd.DataFrame) -> pd.Series:
...
+        )
+        # Preserve old relationship of new reference person before assigning them as new reference persons
+        new_reference_persons_relationship_to_old_reference_person = population.loc[
+            new_reference_persons, ["household_id", "relation_to_household_head"]
+        ]
+        population.loc[
+            new_reference_persons, "relation_to_household_head"
+        ] = "Reference person"
+
+        # Update simulants born in simulation as biological children to their mother if mother is new reference person
```
zmbc, line 980: Can all of this relationship updating logic be in its own method or even multiple methods?

src/vivarium_census_prl_synth_pop/components/population.py
```diff
@@ -956,3 +943,98 @@ def assign_last_name_ids(self, pop: pd.DataFrame) -> pd.Series:
...
+            # Update relationships
+            population.loc[
+                simulants_to_update_idx, "relation_to_household_head"
+            ] = population["relation_to_household_head"].map(
+                relationship_map["relationship_to_new_reference_person"]
+            )
+
+        # Handle extreme edge cases where there would not be a value to map to.
+        population.loc[
+            population["relation_to_household_head"].isnull(), "relation_to_household_head"
```
zmbc, line 1037: This seems slightly worrying, from the perspective that it could cover up errors if any other part of the simulation was introducing nulls here. Can we limit it to only the simulants who were updated?

src/vivarium_census_prl_synth_pop/components/population.py
zmbc: @rmudambi -- I discussed this with @albrja and we decided on this, which there is a PR to document more clearly: https://github.com/ihmeuw/vivarium_research/pull/1117.

You are correct that the case with guardian_2 would be very very rare -- that household structure could only be constructed through as sequence of extremely unlikely domestic migration events.

The original intent of the docs was to not use _guardian_ here (which can -- by design -- be a grandparent or another kind of relative) but to use the _tracked parent_. At that time there was an actual column in the state table tracking the biological mother that had the fertility event. It seems that in the meantime that column was superseded by guardian(s). This is a way to "back out" who the biological mother was from the `guardian_1` column.

src/vivarium_census_prl_synth_pop/model_specifications/model_spec.yaml
albrja: Correct.

src/vivarium_census_prl_synth_pop/components/population.py
```diff
@@ -956,3 +943,98 @@ def assign_last_name_ids(self, pop: pd.DataFrame) -> pd.Series:
...
+        population.loc[
+            new_reference_persons, "relation_to_household_head"
+        ] = "Reference person"
+
+        # Update simulants born in simulation as biological children to their mother if mother is new reference person
+        biological_children_idx = population.index[
+            (population["date_of_birth"] > self.start_time)
+            & (
+                population["household_id"].isin(
+                    new_reference_persons_relationship_to_old_reference_person["household_id"]
```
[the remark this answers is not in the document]
albrja: For your first question, there is no difference but we have assigned a new reference person for these households so I thought that might be confusing.  For the second part, yes you are right I can just find the intersection between households_to_update_idx and sims born after self.start_time.

src/vivarium_census_prl_synth_pop/components/population.py
albrja: That seems like a great idea.

src/vivarium_census_prl_synth_pop/components/population.py
```diff
@@ -1034,7 +1023,9 @@ def update_reference_person_and_relationships(
 
         # Handle extreme edge cases where there would not be a value to map to.
         population.loc[
-            population["relation_to_household_head"].isnull(), "relation_to_household_head"
+            (population.index.isin(households_to_update_idx))
```
zmbc, line 1026: I think you can replace lines 1025-1029 with a `.fillna("Other nonrelative")` on line 1022.

tests/components/test_population.py
```diff
@@ -0,0 +1,175 @@
+import numpy as np
+import pandas as pd
+
+from vivarium_census_prl_synth_pop.components.population import Population
+from vivarium_census_prl_synth_pop.constants import paths
+
+fake_population = pd.DataFrame(
+    data={
+        "household_id": np.sort(np.array([1, 2, 3, 4, 5] * 5)),
```
zmbc, line 9: See my comment in the previous review about this test.

src/vivarium_census_prl_synth_pop/components/population.py
zmbc: But can you not just do `.difference(new_reference_persons)` (note how there is no `.index`)? That surprises me, since above you use new_reference_persons to `.loc`.

src/vivarium_census_prl_synth_pop/components/population.py
zmbc: @albrja I still think this would be a good idea.

tests/components/test_population.py
albrja: This is to test that multiple households will update at the same time.  I am not sure I see the utility in creating different tests for different households when we would be testing the same thing.

src/vivarium_census_prl_synth_pop/components/population.py
albrja: It has to be on the index.  You can do .loc on values or an index.  This also surprised me.

zmbc: The `new_reference_persons` variable is only ever used in this way (as an index) -- so, when you define it, you can just do `pd.Index(...)` and then a) it's easier to understand the `.loc` and b) you can use it here.

tests/components/test_population.py
zmbc: I'm sure this test functionally works, but I find it nearly impossible to read and understand what it is doing. I see your point that you want to test whether it handles multiple households correctly. Maybe you should have a few single-household tests for different cases and then have a test that concatenates together the dataframes used in all those tests to check that it works with multiple households at once.

It isn't obvious at all to me how you worked out the `expected_relationships` and that is a symptom of the test being too complex to follow.

In any case, at least defining this DataFrame like `[{"household_id": 1, "date_of_birth": pd.Timestamp("1987-03-01 00:00:00"), ... }, ...]` would be clearer IMO. And you can omit the `data=`, it is the first argument to the `pd.DataFrame` constructor.

src/vivarium_census_prl_synth_pop/components/population.py
albrja: New reference person is used as an index by looking up the values of that series in the population index.  I see what you are saying with pd.Index(new_reference_persons) and this would work but has also been done already when creating the new_reference_person_prev_relationship variable.  We have to create that variable anyways to store the old relationships so it is not doing any extra work here.

tests/components/test_population.py
albrja: Yes the formatting is awful.  I had written it in a way that was more clear but black wants one value per line.  I can update it to this format.

tests/components/test_population.py
```diff
@@ -4,121 +4,125 @@
 from vivarium_census_prl_synth_pop.components.population import Population
 from vivarium_census_prl_synth_pop.constants import paths
 
-fake_population = pd.DataFrame(
-    data={
-        "household_id": np.sort(np.array([1, 2, 3, 4, 5] * 5)),
+fake_household_1 = pd.DataFrame(
```
rmudambi, line 7: Make this a fixture?

rmudambi (approved):

src/vivarium_census_prl_synth_pop/components/population.py
```diff
@@ -956,3 +943,87 @@ def assign_last_name_ids(self, pop: pd.DataFrame) -> pd.Series:
...
+            ]
+            simulants_to_update_idx = (
+                population.index[population["household_id"].isin(household_ids_to_update)]
+                .difference(new_reference_person_prev_relationship.index)
+                .difference(biological_children_idx)
+            )
+            # Update relationships
+            population.loc[
+                simulants_to_update_idx, "relation_to_household_head"
+            ] = population["relation_to_household_head"].map(
```
zmbc, line 1020:
```suggestion
            ] = population.loc[simulants_to_update_idx, "relation_to_household_head"].map(
```

src/vivarium_census_prl_synth_pop/components/population.py
```diff
@@ -956,3 +943,87 @@ def assign_last_name_ids(self, pop: pd.DataFrame) -> pd.Series:
...
+                relationship_map["relationship_to_new_reference_person"]
+            )
```
zmbc, line 1022:
```suggestion
            ).fillna("Other nonrelative")
```

src/vivarium_census_prl_synth_pop/components/population.py
```diff
@@ -956,3 +943,87 @@ def assign_last_name_ids(self, pop: pd.DataFrame) -> pd.Series:
...
+
+        # Handle extreme edge cases where there would not be a value to map to.
```
zmbc, line 1024: With my suggestions above, you could remove this section of code.

tests/components/test_population.py
```diff
@@ -0,0 +1,188 @@
...
+        "household_id": [4] * 5,
+        "date_of_birth": [
+            pd.Timestamp("1995-04-22 00:00:00"),
+            pd.Timestamp("1996-07-02 00:00:00"),
+            pd.Timestamp("1996-04-22 00:00:00"),
+            pd.Timestamp("1996-09-16 00:00:00"),
+            pd.Timestamp("1997-10-30 00:00:00"),
+        ],
+        "age": [23, 22, 22, 21, 20],
+        "guardian_2": [-1, -1, -1, -1, -1],
```
zmbc, line 87: Why do these have `guardian_2`?

tests/components/test_population.py
```diff
@@ -0,0 +1,188 @@
...
+    {
+        "household_id": [1] * 5,
+        "date_of_birth": [
+            pd.Timestamp("1987-03-01 00:00:00"),
+            pd.Timestamp("2023-05-11 00:00:00"),
+            pd.Timestamp("2026-04-01 00:00:00"),
+            pd.Timestamp("2025-11-21 00:00:00"),
+            pd.Timestamp("1992-07-25 00:00:00"),
+        ],
+        "age": [39, 8, 5, 7, 35],
```
zmbc, line 17: Could we just define age _or_ DOB and then calculate the other one? These are not consistent. IIUC, date of birth only matters in that it is sometimes before and sometimes after the simulation start, so the specific days of the year aren't necessary here.

tests/components/test_population.py
```diff
@@ -0,0 +1,188 @@
...
+            fake_household_3,
+            fake_household_4,
+            fake_household_5,
+        ]
+    ).reset_index()
+
+
+def test_update_to_reference_person_and_relationships(fake_population):
+
+    expected_relationships = pd.Series(
```
zmbc, line 138: My idea with splitting the households was that you could then split the expected relationships, which would allow someone to more easily follow _why_ these are the expected relationships.

zmbc (commented):
Oops I see you just merged this so maybe this isn't useful ¯\\\_(ツ)\_/¯

src/vivarium_census_prl_synth_pop/components/population.py
albrja: I'm putting this outside of the for loop so it only has to search for null values and fill them once vs every time it goes through the loop.  I'm not sure if that will actually make it faster because in all of my testing I have yet to get a simulant where I would hit this but that is the idea.

tests/components/test_population.py
albrja: Typo 👍

src/vivarium_census_prl_synth_pop/components/population.py
zmbc: The operation is O(n), so it the only difference between doing it on each partition of `households_to_update_idx` and once on the whole thing would be any constant-time overhead of calling the `fillna` method. I'm guessing this is _very_ minimal in comparison to the operations done in this method. I think within the for loop is more readable, but it's just a style preference so nbd.

tests/components/test_population.py
albrja: I am not sure I understand what you are saying.  It does matter the day because someone could be born the same year as the simulation starts so we would need to know if it is before or after that day.  Age and date of birth were not used to calculate the other because that allows us to test "bad" data if we wanted to (and maybe should?).  Maybe this isn't the best practice here.

albrja: For example, we could test that a simulant is born in the simulation but is also the oldest member in that household.  Not possible but we would expect them to become the new reference person.

tests/components/test_population.py
albrja: Ok I will split this up and concat it.

tests/components/test_population.py
zmbc: [DOB and age will always be consistent within the simulation](https://github.com/ihmeuw/vivarium_census_prl_synth_pop/blob/8c72892600114944b06b7a353befea1a258220a6/src/vivarium_census_prl_synth_pop/components/population.py#L138-L140), but since this is a unit test it is okay not to enforce that.

I actually think what would be clearer for testing this unit would be to only have two dates of birth: one before the simulation start, and one after. That is all the method cares about.

The more information you have here with all these different dates of birth, the less clear it is what is important. That's why stripping it down to the bare minimum would help me follow.

It _is_ possible that someone is born in the simulation and is also the oldest member in the household, if their parent leaves (via death, domestic migration, emigration).

albrja: Oops I deleted part of my last comment.  Yes they could be the oldest member.  I meant to say we could test a simulant is born in the simulation and is the oldest member in their household that includes their biological mother.  

I will update to only have 2 dates of birth because that is a great idea.

PR #159, opened 2023-01-14
Updated military gq employment test and no longer skip test

albrja:
### Updates and no longer skips test that was checking for unemployment in military group quarters.
- *Category*: Test
- *JIRA issue*: [MIC-3703](https://jira.ihme.washington.edu/browse/MIC-3703)
- *Research reference*: N/A

### Changes and notes
-No longer skips test that verified military gq simulants could not be unemployed.
-Changes test to only include simulants of working age.
-Note: The original bug for this was handled when implementing income and this PR is just to update the test.

### Verification and Testing
All tests passed successfully.


integration_tests/test_employment.py
```diff
@@ -7,21 +7,18 @@
...
-            military_gq["employer_id"].isin(
-                # todo they are not allowed to be unemployed.
-                #  Once this bug is fixed, change the test
-                [data_values.MilitaryEmployer.EMPLOYER_ID, data_values.UNEMPLOYED_ID]
+            (
+                pop["household_id"]
+                == data_values.NONINSTITUTIONAL_GROUP_QUARTER_IDS["Military"]
             )
-        ).all()
+            & (pop["age"] == data_values.WORKING_AGE)
```
rmudambi, line 19: Should be `>= WORKING_AGE`

rmudambi (changes requested):

integration_tests/test_employment.py
albrja: Thank you.

src/vivarium_census_prl_synth_pop/components/businesses.py
```diff
@@ -237,7 +237,10 @@ def on_time_step(self, event: Event):
                 pop["household_id"]
                 == data_values.NONINSTITUTIONAL_GROUP_QUARTER_IDS["Military"]
             )
-            & (pop["age"] >= data_values.WORKING_AGE)
+            & (
+                pop["age"]
+                >= data_values.WORKING_AGE - event.step_size.days / utilities.DAYS_PER_YEAR
+            )
```
albrja, line 243: This catches sims turning working age that live in military gq.  Same thing as turning_working_age but we would not update them here to have military employer.

rmudambi: I don't understand why this is necessary. Simulants age on time-step prepare, so they should already be of working age.

albrja: Simulants age on time step clean up.

rmudambi: Gotcha. I'd rather leave the simulation code simple (i.e. leave this line as is) and make the test only test people who are more than one time-step older than 18.

albrja: Ok to confirm - update the test.  This will mean we will have unemployed simulants in the military for one time step because in time step cleanup they will turn 18 then the next time step they then will have their employer_id updated to military because they will not be included in the working_age_idx.

albrja: This will mean the test will add a time step to the appropriate age.

integration_tests/test_employment.py
```diff
@@ -16,7 +16,7 @@ def test_military_gq_employment(tracked_live_populations):
                 pop["household_id"]
                 == data_values.NONINSTITUTIONAL_GROUP_QUARTER_IDS["Military"]
             )
-            & (pop["age"] == data_values.WORKING_AGE)
+            & (pop["age"] >= data_values.WORKING_AGE + 28 / utilities.DAYS_PER_YEAR)
```
rmudambi, line 19: Use the simulation fixture and the simulation's time-step rather than hard-coding 28 days.

rmudambi (approved):
Feel free to merge once you've addressed the comment.

PR #165, opened 2023-01-19
Update to fake population pytest fixture.

albrja:
### Updates fake population pytest fixture.
- *Category*: Test
- *JIRA issue*: [MIC-3754](https://jira.ihme.washington.edu/browse/MIC-3754)
- *Research reference*: N/A

### Changes and notes
-updates pytest fixture fake population to only have 2 different date of birth values.
-Refactored variable in a related test
-This does not change any functionality of the test, it is addressing additional comments in this PR: https://github.com/ihmeuw/vivarium_census_prl_synth_pop/pull/158

### Verification and Testing
All tests passed.


tests/components/test_population.py
```diff
@@ -8,11 +8,11 @@
     {
         "household_id": [1] * 5,
         "date_of_birth": [
-            pd.Timestamp("1987-03-01 00:00:00"),
-            pd.Timestamp("2023-05-11 00:00:00"),
-            pd.Timestamp("2026-04-01 00:00:00"),
-            pd.Timestamp("2025-11-21 00:00:00"),
-            pd.Timestamp("1992-07-25 00:00:00"),
+            pd.Timestamp("1985-01-01 00:00:00"),
```
zmbc, line 11: Can we define constants for `DOB_BEFORE_SIM_START` and `DOB_AFTER_SIM_START` or something? So that it is explicit what the important difference is between these two dates?

tests/components/test_population.py
```diff
@@ -8,11 +8,11 @@
...
+            pd.Timestamp("2025-12-31 00:00:00"),
+            pd.Timestamp("2025-12-31 00:00:00"),
+            pd.Timestamp("2025-12-31 00:00:00"),
+            pd.Timestamp("1985-01-01 00:00:00"),
         ],
         "age": [39, 8, 5, 7, 35],
         "guardian_1": [-1, 0, 300, -1, -1],
```
zmbc, line 18: None of the fake households you initialize here have defined indices (they are assigned implicitly by the `.reset_index()` below), which makes these guardian IDs pretty hard to follow. You could add `index=` like you did below.

Once you did this, you wouldn't need `.reset_index()` in the fixture.

zmbc (commented):
Thanks so much for indulging me with this! I now can follow these tests!

You might consider adding a test for the logic about `guardian_1`. I don't see a test where it assigns them "Biological child" when they wouldn't already get that based on the relationships alone. Also, testing that it _wouldn't_ use `guardian_1` if they weren't born in the sim.

If you do include these additional tests, I'd give them names more descriptive than `fake_household_6`, such as `fake_household_with_biological_mother` and `fake_household_with_non_biological_guardian`.

zmbc (approved):

PR #167, opened 2023-01-20
Fix GQ household definition, load person weights

zmbc:
### Description

- *Category*: bugfix + artifact
- *JIRA issue*: [MIC-3787](https://jira.ihme.washington.edu/browse/MIC-3787)
- *Research reference*: none

### Changes and notes

This PR bundles two changes which both update the artifact:

* Fixing a bug in how GQ households were defined, which requires a new ACS column
* Loading the person weight column from ACS, which will be used in immigration

### Verification and Testing

Ran simulation to completion; ran integration tests; stepped through code to check that GQ household definition appeared correct and both GQ and non-GQ rows span the whole 2016-2020 interval now.

One integration test is failing due to the problem fixed in #168.

PR #171, opened 2023-01-23
Calculate immigration rates

zmbc:
### Description
- *Category*: implementation
- *JIRA issue*: [MIC-3776](https://jira.ihme.washington.edu/browse/MIC-3776)
- *Research reference*: https://vivarium-research.readthedocs.io/en/latest/models/concept_models/vivarium_census_synthdata/concept_model.html#immigration-rate

### Changes and notes

**Branched off #169, review that first!**

Depends on artifact changes in #167.

### Verification and Testing

Stepped through and verified calculated values make sense.

### DIFF
--- a/src/vivarium_census_prl_synth_pop/components/immigration.py
+++ b/src/vivarium_census_prl_synth_pop/components/immigration.py
@@ -1,6 +1,8 @@
+import pandas as pd
 from vivarium.framework.engine import Builder
+from vivarium.framework.utilities import from_yearly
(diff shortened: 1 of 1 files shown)


zmbc:
I no longer think what I wrote above was correct because this uses the ACS population from whatever _subset_ of the ACS we have loaded (currently Florida only). I need to think through and fix this.

rmudambi (approved):

PR #172, opened 2023-01-24
Post processing/first and middle name data

albrja:
### Loads observer raw outputs and preps name data for name generation code in synthetic pii module.
- *Category*: Post-processing
- *JIRA issue*: [MIC-3788](https://jira.ihme.washington.edu/browse/MIC-3788)
- *Research reference*: N/A

### Changes and notes
-Adds necessary fields to observers so they output needed columns
-Adds call in perform post processing to generate_names for make_results tool.
-Creates data structures for first and middle names to pass to name generation code.

### Verification and Testing
Dataframes have no duplicates and have necessary structure for name generation.

### DIFF
--- a/src/vivarium_census_prl_synth_pop/results_processing/generate_names.py
+++ b/src/vivarium_census_prl_synth_pop/results_processing/generate_names.py
@@ -46,15 +46,15 @@ def fromat_data_for_name_generation(
     data["middle_name_id"] = f"{seed}_" + data["middle_name_id"].astype(str)
     data["last_name_id"] = f"{seed}_" + data["last_name_id"].astype(str)
     # Get year of birth - string column and not timestamp
-    data["yob"] = data["date_of_birth"].str[:4]
+    data["year_of_birth"] = data["date_of_birth"].str[:4]
 
-    first_name_data = data[["first_name_id", "date_of_birth", "sex"]].set_index(
+    first_name_data = data[["first_name_id", "year_of_birth", "sex"]].set_index(
         "first_name_id"
     )
-    middle_name_data = data[["middle_name_id", "date_of_birth", "sex"]].set_index(
+    middle_name_data = data[["middle_name_id", "year_of_birth", "sex"]].set_index(
         "middle_name_id"
     )
-
+    breakpoint()
     return first_name_data, middle_name_data
 
 
--- a/src/vivarium_census_prl_synth_pop/results_processing/generate_names.py
+++ b/src/vivarium_census_prl_synth_pop/results_processing/generate_names.py
@@ -54,7 +54,6 @@ def fromat_data_for_name_generation(
     middle_name_data = data[["middle_name_id", "year_of_birth", "sex"]].set_index(
         "middle_name_id"
     )
-    breakpoint()
     return first_name_data, middle_name_data
 
 
--- a/src/vivarium_census_prl_synth_pop/results_processing/generate_names.py
+++ b/src/vivarium_census_prl_synth_pop/results_processing/generate_names.py
@@ -15,6 +15,7 @@ COLUMNS_FOR_NAME_PROCESSING = [
     "date_of_birth",
     "sex",
     "race_ethnicity",
+    "random_seed",
 ]
 
 
@@ -26,7 +27,12 @@ def load_data(raw_results_dir: Path) -> pd.DataFrame:
     ]
     for obs_dir in observer_directories:
         logger.info(f"Processing data for {obs_dir.name}...")
-        [obs_data.append(pd.read_csv(file)) for file in obs_dir.glob("*.csv.bz2")]
+        obs_files = obs_dir.glob("*.csv.bz2")
+        for file in obs_files:
+            df = pd.read_csv(file)
+            df["random_seed"] = file.name.split(".")[0].split("_")[-1]
+            obs_data.append(df)
+
     data = pd.concat(obs_data)
 
     # Drop unnecessary columns (Social Security Observer does not have sex  and race ethnicity)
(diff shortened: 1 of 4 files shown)


src/vivarium_census_prl_synth_pop/results_processing/generate_names.py
```diff
@@ -0,0 +1,63 @@
...
+def load_data(raw_results_dir: Path):
+    data = pd.DataFrame()
+
+    observer_directories = [f.path for f in os.scandir(raw_results_dir) if f.is_dir()]
+    for obs_dir in observer_directories:
+        logger.info(f'Processing data for {obs_dir.split("/")[-1]}...')
+        results = glob.glob(os.path.join(obs_dir, "*.csv.bz2"))
+        for file in results:
+            tmp = pd.read_csv(file)
+            data = pd.concat([data, tmp])
```
zmbc, line 30: Note: This is going to create a non-unique index. Haven't read far enough yet to know if it matters.

src/vivarium_census_prl_synth_pop/results_processing/generate_names.py
```diff
@@ -0,0 +1,63 @@
...
+    data: pd.DataFrame,
+) -> Tuple:
+    # todo: update to handle with seed column once we are doing parallel runs
+    seed = 0
+
+    # Clean data
+    data["first_name_id"] = f"{seed}_" + data["first_name_id"].astype(str)
+    data["middle_name_id"] = f"{seed}_" + data["middle_name_id"].astype(str)
+    data["last_name_id"] = f"{seed}_" + data["last_name_id"].astype(str)
+    # Get year of birth - string column and not timestamp
```
zmbc, line 48: Was this the outcome of the difficulties we were having with saving to CSV format?

src/vivarium_census_prl_synth_pop/tools/make_results.py
```diff
@@ -24,7 +27,8 @@ def build_results(results_dir: str, mark_best: bool, test_run: bool) -> None:
 
     logger.info("Copying raw results to final location.")
     raw_output_dir = Path(results_dir) / paths.RAW_RESULTS_DIR_NAME
-    shutil.copytree(raw_output_dir, final_output_dir)
+    # Perform post-processing
```
zmbc, line 30: This comment isn't necessary given the name of the method.

src/vivarium_census_prl_synth_pop/tools/make_results.py
```diff
@@ -6,6 +6,9 @@
 
 from vivarium_census_prl_synth_pop import utilities
 from vivarium_census_prl_synth_pop.constants import paths
+from vivarium_census_prl_synth_pop.results_processing.generate_names import (
+    generate_names,
```
zmbc, line 10: I think if we made an `__init__.py` in results_processing we could eliminate this repetition of "generate_names," which is also going to look nicer when we have multiple postprocessing steps.

zmbc (approved):

src/vivarium_census_prl_synth_pop/results_processing/generate_names.py
albrja: No that was for changing the date of birth string in a separate ticket and the problem occurs when you read in the csv.  For example, March 1st, 2020 written as 03012020 gets interpreted as an integer when you do read_csv so the leading 0 is dropped.  Quoting option for write_csv seemed to not work with the date format changed to a string either.

src/vivarium_census_prl_synth_pop/results_processing/generate_names.py
```diff
@@ -0,0 +1,63 @@
...
+
+    observer_directories = [f.path for f in os.scandir(raw_results_dir) if f.is_dir()]
+    for obs_dir in observer_directories:
+        logger.info(f'Processing data for {obs_dir.split("/")[-1]}...')
+        results = glob.glob(os.path.join(obs_dir, "*.csv.bz2"))
+        for file in results:
+            tmp = pd.read_csv(file)
+            data = pd.concat([data, tmp])
+
+    # Drop unnecessary columns (Social Security Observer does not have sex  and race ethnicity)
```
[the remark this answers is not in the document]
albrja: Yes this is an old comment.

src/vivarium_census_prl_synth_pop/tools/make_results.py
albrja: This is a good idea.  Doing this I wasn't sure how exactly we would want to set up the results processing stage since we will have multiple steps and I am not sure what would be necessary for each step to do.

src/vivarium_census_prl_synth_pop/results_processing/generate_names.py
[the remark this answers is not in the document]
rmudambi: It's actually even better to keep a running list of all the DataFrames you are compiling and call a single `pd.concat` on the whole list at once, as that's actually faster.

rmudambi (changes requested):

src/vivarium_census_prl_synth_pop/results_processing/generate_names.py
```diff
@@ -0,0 +1,63 @@
...
+) -> Tuple:
+    # todo: update to handle with seed column once we are doing parallel runs
+    seed = 0
+
+    # Clean data
+    data["first_name_id"] = f"{seed}_" + data["first_name_id"].astype(str)
+    data["middle_name_id"] = f"{seed}_" + data["middle_name_id"].astype(str)
+    data["last_name_id"] = f"{seed}_" + data["last_name_id"].astype(str)
+    # Get year of birth - string column and not timestamp
+    data["year_of_birth"] = data["date_of_birth"].str[:4]
```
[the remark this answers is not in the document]
albrja: It was a time stamp when we wrote the file but now it is just a string when we read it in.

src/vivarium_census_prl_synth_pop/results_processing/generate_names.py
```diff
@@ -0,0 +1,63 @@
...
+    # Drop unnecessary columns (Social Security Observer does not have sex  and race ethnicity)
+    data = data[COLUMNS_FOR_NAME_PROCESSING].drop_duplicates()
+
+    return data
+
+
+def fromat_data_for_name_generation(
+    data: pd.DataFrame,
+) -> Tuple:
+    # todo: update to handle with seed column once we are doing parallel runs
```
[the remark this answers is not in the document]
albrja: These files do not have a random draw or random seed column. like other models.  Should I be getting that from the file name?  Doing so would mean I would need to add that column when the file is read in and not later when the data is cleaned.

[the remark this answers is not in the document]
albrja: I am always hesitant to have data dependent on filenames because of how easily they can be modified.

src/vivarium_census_prl_synth_pop/results_processing/generate_names.py
```diff
@@ -18,26 +18,26 @@
...
+        obs_dir for obs_dir in raw_results_dir.glob("*") if obs_dir.is_dir()
+    ]
     for obs_dir in observer_directories:
-        logger.info(f'Processing data for {obs_dir.split("/")[-1]}...')
-        results = glob.glob(os.path.join(obs_dir, "*.csv.bz2"))
-        for file in results:
-            tmp = pd.read_csv(file)
-            data = pd.concat([data, tmp])
+        logger.info(f"Processing data for {obs_dir.name}...")
+        [obs_data.append(pd.read_csv(file)) for file in obs_dir.glob("*.csv.bz2")]
```
rmudambi, line 29: This expression shouldn't be enclosed in `[ ]`?

src/vivarium_census_prl_synth_pop/results_processing/generate_names.py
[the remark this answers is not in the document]
rmudambi: We should be able to rely on members of our team not maliciously renaming these files.

src/vivarium_census_prl_synth_pop/results_processing/generate_names.py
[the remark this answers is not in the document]
rmudambi: You can use `pd.to_datetime` to convert it to a datetime and then do `.dt.year`. That is cleaner than relying on the specific substring matching.

src/vivarium_census_prl_synth_pop/results_processing/generate_names.py
```diff
@@ -48,16 +48,14 @@ def fromat_data_for_name_generation(
     # Get year of birth - string column and not timestamp
     data["year_of_birth"] = data["date_of_birth"].str[:4]
 
+    # We only need one dataframe here since first_name_id and middle_name_id will be the same.
```
rmudambi, line 51: Let's not rely on this, since it could change in the future.

albrja: That is fine but will require some updates to the name generator and related tests.

rmudambi (approved):

PR #173, opened 2023-01-25
Fix definition of GQ vs non-GQ immigrant

zmbc:
### Description

- *Category*: bugfix
- *JIRA issue*: none
- *Research reference*: none

### Changes and notes

I had a typo in "institutionalized" -- I have replaced this with using the newly loaded "household_type" column, which is clearer.

I will squash this commit with #169 before we merge the epic into main.

### Verification and Testing

Stepped through and verified that gq_immigrants had the expected relationships, and non_gq_immigrants had the expected relationships.

### DIFF
--- a/src/vivarium_census_prl_synth_pop/components/immigration.py
+++ b/src/vivarium_census_prl_synth_pop/components/immigration.py
@@ -1,4 +1,6 @@
+import pandas as pd
 from vivarium.framework.engine import Builder
+from vivarium.framework.utilities import from_yearly
 
 from vivarium_census_prl_synth_pop.constants import data_keys
 
@@ -25,6 +27,7 @@ class Immigration:
 
     def setup(self, builder: Builder):
         persons_data = builder.data.load(data_keys.POPULATION.PERSONS)
+        self.total_person_weight = persons_data["person_weight"].sum()
 
         immigrants = persons_data[persons_data["immigrated_in_last_year"]]
 
@@ -35,6 +38,10 @@ class Immigration:
             ]
         )
         self.gq_immigrants = immigrants[is_gq]
+        self.gq_immigrants_per_time_step = self._immigrants_per_time_step(
+            self.gq_immigrants,
+            builder.configuration,
+        )
 
         non_gq_immigrants = immigrants[~is_gq]
         immigrant_reference_people = non_gq_immigrants[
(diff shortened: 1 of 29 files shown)


src/vivarium_census_prl_synth_pop/components/immigration.py
```diff
@@ -29,10 +29,7 @@ def setup(self, builder: Builder):
         immigrants = persons_data[persons_data["immigrated_in_last_year"]]
 
         is_gq = immigrants["relation_to_household_head"].isin(
-            [
-                "Insitutionalized GQ pop",
-                "Noninstitutionalized GQ pop",
-            ]
+            data_values.GROUP_QUARTER_IDS.keys()
```
rmudambi, line 32: I thought this was supposed to key off of the new housing type column?

zmbc: That column is in the households, not persons, data. In other places, we use the relationship column for this purpose when working with the persons data. I suppose to be more technically correct we could always join to the households data and use the newly-loaded column, but that might overcomplicate things.

Just to note: the reason for adding that column in #167 was not to use it for immigration; it was to fix an unrelated bug in our handling of the households data.

rmudambi: To clarify, this column should be correct once we're in the simulation. The issue is that we couldn't rely on this column in the source ACS data-set. Am I understanding this correctly? If so, feel free to merge.

rmudambi (approved):

src/vivarium_census_prl_synth_pop/components/immigration.py
zmbc: Nope not quite, and I see how this is confusing.

I think what you said is a clearer way to think about it, and I will update this code to use the newly-loaded housing type column from ACS.

zmbc:
@rmudambi I've essentially completely replaced this with using the `household_type` column since I agree that is clearer. I also tested this again to ensure it is actually working.

rmudambi (approved):

src/vivarium_census_prl_synth_pop/components/immigration.py
```diff
@@ -0,0 +1,87 @@
...
+    @property
+    def name(self):
+        return "immigration"
+
+    #################
+    # Setup methods #
+    #################
+
+    def setup(self, builder: Builder):
+        persons_data = builder.data.load(data_keys.POPULATION.PERSONS)
```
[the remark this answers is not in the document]
zmbc: I was thinking about this too! I like your solution, and I don't think it's too much trouble.

PR #176, opened 2023-01-25
make reference person moving/death atomic with reassigning a referenc…

rmudambi:
…e person

### Description

- *Category*: implementation
- *JIRA issue*: [MIC-3796](https://jira.ihme.washington.edu/browse/MIC-3796)
- *Research reference*: 

### Changes and notes
Previously reference people would die or move and the simulation would account for that having happened on time-step cleanup. This meant that we would be in the correct state at the end of the simulation, but it meant that during a time-step there would be situations where households had no living reference person, which makes several actions difficult to reason about. This makes all of these reference persons atomic with the new assignment of a reference person

Fixed some related tests that were breaking at larger population sizes. Increased the size of the default integration test population size so that these failures don't get missed going forward.

Also removed an unused function from utilities. 

### Verification and Testing
Got unit and integration tests working.

### DIFF
--- a/integration_tests/test_domestic_migration.py
+++ b/integration_tests/test_domestic_migration.py
@@ -1,5 +1,3 @@
-import pytest
-
 from vivarium_census_prl_synth_pop.constants import data_values
 
 # TODO: Broader test coverage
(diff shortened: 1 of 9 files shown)


integration_tests/test_domestic_migration.py
```diff
@@ -75,6 +73,18 @@ def test_individuals_move_into_group_quarters(simulants_on_adjacent_timesteps):
         )
 
 
+def get_households_with_reference_person(after, before, movers):
```
zmbc, line 76: I find the name of this method a bit confusing -- maybe `get_households_with_stable_reference_person` or something?

integration_tests/test_domestic_migration.py
```diff
@@ -75,6 +73,18 @@ def test_individuals_move_into_group_quarters(simulants_on_adjacent_timesteps):
...
+    reference_persons = before["relation_to_household_head"] == "Reference person"
+    movers = before["household_id"] != after["household_id"]
+    deaths = before["alive"] != after["alive"]
+    emigrants = before["in_united_states"] != after["in_united_states"]
+    reference_people_movers = reference_persons & (movers | deaths | emigrants)
```
zmbc, line 81: Likewise, this variable could be `changed_reference_people` or something.

zmbc (approved):
This looks great!

albrja (approved):

integration_tests/test_domestic_migration.py
zmbc: Wait, I missed this on first read, but this `movers` arg is not used.

rmudambi: I'll rename these in my next pr as it'll be easier to manage the git history

PR #177, opened 2023-01-25
Feature/rmudambi/implement hh details

rmudambi:
### Description

- *Category*: refactor
- *JIRA issue*: [MIC-3780](https://jira.ihme.washington.edu/browse/MIC-3780)
- *Research reference*: 

### Changes and notes
Integrate business and household address refactor into simulation
Fix failing tests
Prevent simulants from moving into empty households 

### Verification and Testing
Tests pass (both unit and integration) and new tests written

### DIFF
--- a/integration_tests/test_domestic_migration.py
+++ b/integration_tests/test_domestic_migration.py
@@ -11,7 +11,8 @@ def test_individuals_move(simulants_on_adjacent_timesteps):
         individual_movers = before["household_id"] != after["household_id"]
         assert individual_movers.any()
         assert (
-            before[individual_movers]["address_id"] != after[individual_movers]["address_id"]
+            before[individual_movers]["household_details.address_id"]
+            != after[individual_movers]["household_details.address_id"]
         ).all()
 
 
@@ -23,7 +24,7 @@ def test_individuals_move_into_new_households(simulants_on_adjacent_timesteps):
         # established by a new-household mover (within the same time step).
         movers_into_new_households = (
             (before["household_id"] != after["household_id"])
-            & (after["housing_type"] == "Standard")
+            & (after["household_details.housing_type"] == "Standard")
             & (~after["household_id"].isin(before["household_id"]))
         )
         assert movers_into_new_households.any()
@@ -56,18 +57,18 @@ def test_individuals_move_into_new_households(simulants_on_adjacent_timesteps):
         assert new_households.nunique() == len(new_households)
 
         # Address IDs moved to are unique and new
-        new_addresses = after[new_household_movers]["address_id"]
+        new_addresses = after[new_household_movers]["household_details.address_id"]
         assert new_addresses.nunique() == len(new_addresses)
-        assert not new_addresses.isin(before["address_id"]).any()
+        assert not new_addresses.isin(before["household_details.address_id"]).any()
 
 
 def test_individuals_move_into_group_quarters(simulants_on_adjacent_timesteps):
     for before, after in simulants_on_adjacent_timesteps:
         gq_movers = (before["household_id"] != after["household_id"]) & (
-            after["housing_type"] != "Standard"
+            after["household_details.housing_type"] != "Standard"
         )
         assert gq_movers.any()
-        assert (before[gq_movers]["housing_type"] == "Standard").any()
+        assert (before[gq_movers]["household_details.housing_type"] == "Standard").any()
         assert after[gq_movers]["household_id"].isin(data_values.GQ_HOUSING_TYPE_MAP).all()
         assert (
             after[gq_movers]["relation_to_household_head"]
@@ -92,7 +93,7 @@ def test_individuals_move_into_existing_households(simulants_on_adjacent_timeste
     for before, after in simulants_on_adjacent_timesteps:
         non_reference_person_movers = (
             (before["household_id"] != after["household_id"])
-            & (after["housing_type"] == "Standard")
+            & (after["household_details.housing_type"] == "Standard")
             & (after["household_id"].isin(before["household_id"]))
         )
         assert non_reference_person_movers.any()
@@ -124,7 +125,7 @@ def test_individuals_move_into_existing_households(simulants_on_adjacent_timeste
 def test_households_move(simulants_on_adjacent_timesteps):
     for before, after in simulants_on_adjacent_timesteps:
         household_movers = (before["household_id"] == after["household_id"]) & (
-            before["address_id"] != after["address_id"]
+            before["household_details.address_id"] != after["household_details.address_id"]
         )
         assert household_movers.any()
 
@@ -144,36 +145,36 @@ def test_households_move(simulants_on_adjacent_timesteps):
 
         # Address IDs moved to are new
         moved_households = before[household_movers]["household_id"].unique()
-        new_addresses = after[household_movers]["address_id"]
+        new_addresses = after[household_movers]["household_details.address_id"]
         assert new_addresses.nunique() == len(moved_households)
-        assert not before["address_id"].isin(new_addresses).any()
+        assert not before["household_details.address_id"].isin(new_addresses).any()
 
         # Never GQ households
-        assert (before[household_movers]["housing_type"] == "Standard").all()
+        assert (
+            before[household_movers]["household_details.housing_type"] == "Standard"
+        ).all()
 
 
-def test_only_living_people_move(simulants_on_adjacent_timesteps):
+def test_only_living_people_change_households(simulants_on_adjacent_timesteps):
     for before, after in simulants_on_adjacent_timesteps:
-        movers = before["address_id"] != after["address_id"]
+        movers = before["household_id"] != after["household_id"]
         assert after[movers]["tracked"].all()
         assert (after[movers]["alive"] == "alive").all()
 
 
-def test_address_uniqueness(populations):
-    """Check that all units (households or employers) have unique details
-    and also that simulants of the same unit have the same details
+@pytest.mark.parametrize(
+    "unit_id_col, address_id_col",
+    [
+        ("employer_id", "business_details.employer_address_id"),
+        ("household_id", "household_details.address_id"),
+    ],
+)
+def test_unit_address_uniqueness(populations, unit_id_col, address_id_col):
+    """Check that all units (households and businesses) have unique details and
+    also that simulants of the same unit have the same details
     """
-    _test_address_uniqueness(
-        pops=populations,
-        unit_id_col="employer_id",
-        address_id_col="business_details.employer_address_id",
-        other_address_cols=[],
-    )
-
-
-def _test_address_uniqueness(pops, unit_id_col, address_id_col, other_address_cols):
-    address_cols = [unit_id_col, address_id_col] + other_address_cols
-    for pop in pops:
+    address_cols = [unit_id_col, address_id_col]  # TODO: add state/puma
+    for pop in populations:
         # Check that all simulants in the same unit have the same address
         assert (pop.groupby(unit_id_col)[address_cols].nunique() == 1).all().all()
         # Check that all units have unique addresses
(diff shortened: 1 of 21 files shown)


src/vivarium_census_prl_synth_pop/components/businesses.py
```diff
@@ -56,18 +52,16 @@ def name(self):
...
         self.columns_created = [
             "employer_id",
-            "employer_name",
-            "employer_address_id",
             "personal_income_propensity",
             "employer_income_propensity",
         ]
         self.columns_used = [
-            "address_id",
+            "tracked",
```
albrja, line 64: In all the places where this is used would it not have queried to tracked simulants only there anyways?  I thought population view's default was tracked simulants?

src/vivarium_census_prl_synth_pop/components/household_migration.py
```diff
@@ -62,18 +55,24 @@ def setup(self, builder: Builder):
         )
 
         self.randomness = builder.randomness.get_stream(self.name)
-        self.columns_created = ["address_id"]
         self.columns_used = [
+            "tracked",
```
albrja, line 59: Same thing here.

tests/components/observers/test_household_survey_observers.py
```diff
@@ -51,40 +62,50 @@ def test_on_simulation_end(observer, mocker):
...
+):
     """Are responses recorded correctly (including concatenation after time steps?"""
     sim_start_date = "2021-01-01"
     # Simulate first time step
     event = mocker.MagicMock()
     event.time = pd.to_datetime(sim_start_date)
     event.index = pd.Index([0, 1])
-    observer.population_view.get.return_value = mocked_pop_view  # FIXME: This returns mocked_pop_view regardless of event.index or alive status
+    # FIXME: This returns mocked_pop_view regardless of event.index or alive status
+    observer.population_view.get.return_value = mocked_pop_view
```
albrja, line 75: Does this mean it is returning all of `mocked_pop_view` no matter what?  If I understand the mocks correctly I think you are just assigning it here.  You need to mock population_view in the mocked observer's builder.

tests/components/observers/test_household_survey_observers.py
```diff
@@ -51,40 +62,50 @@ def test_on_simulation_end(observer, mocker):
...
+):
     """Are responses recorded correctly (including concatenation after time steps)?"""
     sim_start_date = "2021-01-01"
     # Simulate first time step
     event = mocker.MagicMock()
     event.time = pd.to_datetime(sim_start_date)
     event.index = pd.Index([0, 1])
-    observer.population_view.get.return_value = mocked_pop_view  # FIXME: This returns mocked_pop_view regardless of event.index or alive status
+    # FIXME: This returns mocked_pop_view regardless of event.index or alive status
+    observer.population_view.get.return_value = mocked_pop_view
```
albrja, line 103: Same here if I am right.

src/vivarium_census_prl_synth_pop/components/businesses.py
rmudambi: This column is added because we want to be able to record untracked simulants in the business details pipeline.

integration_tests/test_domestic_migration.py
```diff
@@ -46,25 +50,27 @@ def test_individuals_move_into_new_households(simulants_on_adjacent_timesteps):
         ).all()
         # These should be the vast majority, since non-reference-person moves to
         # new households will be quite rare.
-        assert new_household_movers.sum() / movers_into_new_households.sum() > 0.95
+        # NOTE: This is sensitive to small population size, eg running 20k simulants
+        # was causing it to break w/ a ratio 19/20 = 0.95
+        assert new_household_movers.sum() / movers_into_new_households.sum() >= 0.95
```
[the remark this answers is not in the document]
rmudambi: No the number is arbitrary. It's basically just checking that there are very few people who move into a new household who aren't founding a new one. 95% is very conservative - we expect this to apply to almost everyone.

integration_tests/test_domestic_migration.py
```diff
@@ -114,43 +123,126 @@ def test_individuals_move_into_existing_households(simulants_on_adjacent_timeste
...
+        after_units = after_units.reindex(total_index)
+        mask_moved_units = before_units[address_id_col] != after_units[address_id_col]
+
+        # check that the number of moved units that have the same details is very low
+        # NOTE: we cannot assert that all moved units have different details
+        # because they are free to move within the same state or even puma
+        assert (
+            (
+                before_units.loc[mask_moved_units, other_address_cols]
+                == after_units.loc[mask_moved_units, other_address_cols]
```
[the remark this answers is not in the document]
rmudambi: If you look at line 221 you can see that `mask_moved_units` has had the address id columns change by definition.

src/vivarium_census_prl_synth_pop/components/household_migration.py
rmudambi: Same reason as above.

tests/components/observers/test_household_survey_observers.py
rmudambi: You can do it either way. If we wanted to be able to get variable results as that comment suggests, this approach won't enable it. I'm going to consider that out of scope for this PR ... it's big enough as it is.

albrja: Ahh ok that makes sense.

albrja (approved):

integration_tests/test_household_structure.py
```diff
@@ -90,3 +96,52 @@ def test_households_only_have_one_reference_person(tracked_live_populations):
...
+            .set_index("household_id")
+        )
+
+        pd.testing.assert_frame_equal(before, after)
+        assert not after.index.duplicated().any()
+
+
+def test_initialized_state_complete_coverage(populations, sim):
+    """Initialized states should include all locations from artifact"""
+    initialized_pop = populations[0]
```
zmbc, line 132: Is there a reason this test can't be on every population instead of only the first?

integration_tests/test_household_structure.py
```diff
@@ -90,3 +96,52 @@ def test_households_only_have_one_reference_person(tracked_live_populations):
...
+    states_in_artifact = set(
+        sim._data.artifact.load("population.households").index.unique(level="state")
+    )
+    assert states_in_artifact == set(initialized_pop["state"])
+
+
+def test_initialized_pumas_states(populations):
```
zmbc, line 139: Can we add a note to this test, that it should work on _every_ population once state/PUMA update from moving, and that it should work on the GQ population as well once state/PUMA are made part of household_details?

zmbc (approved):
This seems like a huge improvement!

I think there is some diff in this PR that is related to scaling to multiple states. Not sure if that was on purpose or by mistake.

zmbc:
@rmudambi Since there are observer changes in this PR, which don't get hit by integration tests, you'll want to run the simulation all the way through with these changes (if you haven't already), to ensure the observers run without error.

integration_tests/test_domestic_migration.py
```diff
@@ -114,43 +123,126 @@ def test_individuals_move_into_existing_households(simulants_on_adjacent_timeste
         assert (
             after.loc[
                 mover_to_household_without_reference_person, "relation_to_household_head"
-            ].isin(["Other nonrelative", "Reference person"])
+            ].isin(["Other nonrelative", "Reference person", "Roommate"])
```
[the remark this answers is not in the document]
rmudambi: This kind of comes down to whether we are testing the input data or not. I think we probably should be, and if the input data changes, this test should fail. Of course we wouldn't notice if "Roommate" was removed as an option, but it would at least fail if someone added another option to the file unintentionally.

integration_tests/test_domestic_migration.py
```diff
@@ -114,43 +123,126 @@ def test_individuals_move_into_existing_households(simulants_on_adjacent_timeste
...
+def test_unit_address_uniqueness(populations, unit_id_col, address_id_col):
+    """Check that all units (households and businesses) have unique details and
+    also that simulants of the same unit have the same details
+    """
+    address_cols = [unit_id_col, address_id_col]  # TODO: add state/puma
+    for pop in populations:
+        # Check that all simulants in the same unit have the same address
+        assert (pop.groupby(unit_id_col)[address_cols].nunique() == 1).all().all()
+        # Check that all units have unique addresses
+        assert len(pop[address_cols].drop_duplicates()) == len(pop[unit_id_col].unique())
```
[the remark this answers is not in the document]
rmudambi: This seems cleaner.

integration_tests/test_domestic_migration.py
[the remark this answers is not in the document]
zmbc: I see your point. Maybe we should do something as simple as defining a variable `other_nonrelative_can_update_to = ["Reference person", "Other nonrelative", "Roommate"]` just to be explicit about what is special about these three relationships.

integration_tests/test_domestic_migration.py
```diff
@@ -114,43 +123,126 @@ def test_individuals_move_into_existing_households(simulants_on_adjacent_timeste
...
+def test_addresses_during_moves(simulants_on_adjacent_timesteps, unit_id_col, address_id_col):
+    """Check that unit (household and business) address details change after a move."""
+    other_address_cols = []  # TODO: add state/puma
+    address_cols = [address_id_col] + other_address_cols
+    for before, after in simulants_on_adjacent_timesteps:
+        # get the unique unit (household or employer) dataset for before and after
+        before_units = before.groupby(unit_id_col)[address_cols].first()
+        after_units = after.groupby(unit_id_col)[address_cols].first()
+        # reindex in case there are unique units to one of the datasets
+        total_index = before_units.index.union(after_units.index)
```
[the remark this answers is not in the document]
rmudambi: I agree that intersection is correct rather than union. The `assert_frame_equal` below works because it operates only on households that have the same `address_id` in both `after` and `before` (and therefore must exist in both).

integration_tests/test_household_structure.py
rmudambi: I don't see why not.

src/vivarium_census_prl_synth_pop/components/businesses.py
```diff
@@ -379,3 +358,12 @@ def calculate_income(self, idx: pd.Index) -> pd.Series:
         )
 
         return income
+
+    def get_business_details(self, idx: pd.Index) -> pd.DataFrame:
+        """Source of the business details pipeline"""
+        pop = self.population_view.get(idx)[["employer_id"]]
```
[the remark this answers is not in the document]
rmudambi: That's because we actually need to record them for untracked simlants in the tax observers, whereas we don't need income for those observers, since they use annual earnings on observation rather than income.

src/vivarium_census_prl_synth_pop/constants/data_keys.py
```diff
@@ -38,8 +41,10 @@ def log_name(self):
...
 
 
-POPULATION = __Population()
 SYNTHETIC_DATA = __SyntheticData()
 
 
-MAKE_ARTIFACT_KEY_GROUPS = [POPULATION, SYNTHETIC_DATA]
+MAKE_ARTIFACT_KEY_GROUPS = [
+    POPULATION,
+    # SYNTHETIC_DATA,  Will be moved into separate artifact later
```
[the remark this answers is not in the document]
rmudambi: I think this is a mistake. I'm going to remove commenting it out.

src/vivarium_census_prl_synth_pop/constants/metadata.py
```diff
@@ -16,7 +16,11 @@
 MAKE_ARTIFACT_RUNTIME = "3:00:00"
 MAKE_ARTIFACT_SLEEP = 10
 
-LOCATIONS = ["Florida"]
+LOCATIONS = ["United States of America", "Florida"]
```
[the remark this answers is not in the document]
rmudambi: This enables an artifact to be build using the location USA

PR #180, opened 2023-01-28
Add additional tax observers to support simulated 1040 data

aflaxman:
### Description

- *Category*: observers
- *JIRA issue*: [MIC-3786](https://jira.ihme.washington.edu/browse/MIC-3786)
- *Research reference*: https://vivarium-research.readthedocs.io/en/latest/models/concept_models/vivarium_census_synthdata/concept_model.html#form

### Changes and notes
 
It is a little weird to have the dependents in the W2, so Abie and Rajan decided to refactor this approach to have a composite observer that has three subclasses of BaseObserver.  One will be a refactor of the W2 Observer, another will be the 1040 observer described above, and the third will be the eligible dependent information, organized by dependent (in a long format). See design discussion in JIRA for a sketch https://jira.ihme.washington.edu/browse/MIC-3772

### Verification and Testing

Simulate runs successfully with model_spec.yaml, results look acceptable.  All tests pass.

### DIFF
--- a/src/vivarium_census_prl_synth_pop/components/observers.py
+++ b/src/vivarium_census_prl_synth_pop/components/observers.py
@@ -495,8 +495,34 @@ class SocialSecurityObserver(BaseObserver):
         return pd.concat([df_creation, df_death]).sort_values(["event_date", "date_of_birth"])
 
 
+class TaxObserver:
+    """Holder for three interdependent observers relevant to tax data"""
+
+    @property
+    def name(self):
+        return f"tax_observer"
+
+    def __init__(self):
+        tax_w2 = TaxW2Observer()
+        tax_1040 = Tax1040Observer(tax_w2)
+
+        self._sub_components = [
+            tax_w2,
+            tax_1040,
+        ]  # following pattern from vivarium.examples.disease_model.disease.SISDiseaseModel
+        # TODO: it would be cool if there was more documentation on this, and if it was easy to find!
+
+    @property
+    def sub_components(self):
+        return self._sub_components
+
+
 class TaxW2Observer(BaseObserver):
-    """Class for observing columns relevant to W2 and 1099 tax data."""
+    """Class for observing columns relevant to W2 and 1099 tax data.
+
+    Maintains a pd.Series for last year's income for each (person, employer)-pair,
+    which the TaxDependentObserver uses
+    """
 
     INPUT_VALUES = ["income", "household_details", "business_details"]
     ADDITIONAL_INPUT_COLUMNS = [
@@ -561,6 +587,7 @@ class TaxW2Observer(BaseObserver):
         # this check on_time_step instead of on_time_step__prepare
         builder.event.register_listener("time_step__prepare", self.on_time_step__prepare)
         self.income_to_date = empty_income_series()
+        self.income_last_year = empty_income_series()
         self.time_step = builder.configuration.time.step_size  # in days
 
     def on_time_step__prepare(self, event):
@@ -619,8 +646,10 @@ class TaxW2Observer(BaseObserver):
             "ssn",
         ]:
             df_w2[col] = df_w2["simulant_id"].map(pop_full[col])
+
         df_w2["address_id"] = df_w2["simulant_id"].map(household_details["address_id"])
         df_w2["housing_type"] = df_w2["simulant_id"].map(household_details["housing_type"])
+
         # Tracked, US population to be dependents or get their SSNs borrowed
         pop = pop_full[
             (pop_full["alive"] == "alive")
@@ -695,9 +724,92 @@ class TaxW2Observer(BaseObserver):
             df_w2[col] = df_w2["simulant_id"].map(s)
 
         # re-initialize income-to-date series for next year of income counting
+        self.income_last_year = self.income_to_date
         self.income_to_date = empty_income_series()
 
         return df_w2[self.output_columns]
+
+
+class Tax1040Observer(BaseObserver):
+    """Class for observing columns relevant to 1040 tax data (that is not
+    already recorded by W2 or dependent observer)
+    """
+
+    INPUT_VALUES = ["income", "household_details", "business_details"]
+    ADDITIONAL_INPUT_COLUMNS = [
+        "alive",
+        "tracked",
+        "ssn",
+        "relation_to_household_head",
+    ]
+    OUTPUT_COLUMNS = [
+        "simulant_id",
+        "first_name_id",
+        "middle_name_id",
+        "last_name_id",
+        "age",
+        "date_of_birth",
+        "sex",
+        "ssn",
+        "address_id",
+        "relation_to_household_head",
+        "housing_type",
+    ]
+
+    def __init__(self, w2_observer):
+        super().__init__()
+        self.w2_observer = w2_observer
+
+    def __repr__(self):
+        return f"Tax1040Observer()"
+
+    @property
+    def name(self):
+        return f"tax_1040_observer"
+
+    @property
+    def input_columns(self):
+        return self.DEFAULT_INPUT_COLUMNS + self.ADDITIONAL_INPUT_COLUMNS
+
+    @property
+    def input_values(self):
+        return self.INPUT_VALUES
+
+    @property
+    def output_columns(self):
+        return self.OUTPUT_COLUMNS
+
+    def setup(self, builder: Builder):
+        super().setup(builder)
+        self.clock = builder.time.clock()
+
+    def to_observe(self, event: Event) -> bool:
+        """Observe if April 15 falls during this time step"""
+        tax_date = dt.datetime(event.time.year, 4, 15)
+        return self.clock() < tax_date <= event.time
+
+    def get_observation(self, event: Event) -> pd.DataFrame:
+        pop = self.population_view.get(event.index)
+        household_details = self.pipelines["household_details"](pop.index)
+        pop[["address_id", "housing_type"]] = household_details[
+            ["address_id", "housing_type"]
+        ]
+
+        # add derived columns
+        pop["simulant_id"] = pop.index
+
+        # delete extraneous columns
+        for col in ["alive", "guardian_1", "guardian_2", "race_ethnicity", "tracked"]:
+            del pop[col]
+
+        # to save space, remove rows for simulants with zero income
+        ### get series of all person/employment pairs from w2 observer
+        s_w2 = self.w2_observer.income_last_year
+        non_zero_income_ids = s_w2.reset_index()["simulant_id"].unique()
+        non_zero_income_rows = pop["simulant_id"].isin(non_zero_income_ids)
+        pop = pop[non_zero_income_rows]
+
+        return pop
 
 
 def empty_income_series():
--- a/src/vivarium_census_prl_synth_pop/components/observers.py
+++ b/src/vivarium_census_prl_synth_pop/components/observers.py
@@ -535,7 +535,6 @@ class TaxW2Observer(BaseObserver):
         "employer_id",
     ]
     OUTPUT_COLUMNS = [
-        "simulant_id",
         "first_name_id",
         "middle_name_id",
         "last_name_id",
@@ -846,7 +845,6 @@ class Tax1040Observer(BaseObserver):
         "relation_to_household_head",
     ]
     OUTPUT_COLUMNS = [
-        "simulant_id",
         "first_name_id",
         "middle_name_id",
         "last_name_id",
@@ -900,7 +898,6 @@ class Tax1040Observer(BaseObserver):
         ]
 
         # add derived columns
-        pop["simulant_id"] = pop.index
         pop["tax_year"] = event.time.year - 1
 
         # delete extraneous columns
@@ -911,7 +908,7 @@ class Tax1040Observer(BaseObserver):
         ### get series of all person/employment pairs from w2 observer
         s_w2 = self.w2_observer.income_last_year
         non_zero_income_ids = s_w2.reset_index()["simulant_id"].unique()
-        non_zero_income_rows = pop["simulant_id"].isin(non_zero_income_ids)
+        non_zero_income_rows = pop.index.isin(non_zero_income_ids)
         pop = pop[non_zero_income_rows]
 
         return pop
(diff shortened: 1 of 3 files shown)


src/vivarium_census_prl_synth_pop/components/observers.py
```diff
@@ -680,24 +798,120 @@ def get_observation(self, event: Event) -> pd.DataFrame:
...
+            "middle_name_id",
+            "last_name_id",
+            "age",
+            "date_of_birth",
+            "sex",
+            "ssn",
+            "address_id",
+            "housing_type",
+        ]:
+            df[col] = df["dependent_id"].map(pop_full[col])
```
zmbc, line 817: What is the situation in which these columns are not already present? They seem like they all should be.

src/vivarium_census_prl_synth_pop/components/observers.py
```diff
@@ -680,24 +798,120 @@ def get_observation(self, event: Event) -> pd.DataFrame:
...
 
-        for col, s in [
-            ["eligible_dependent_id_list", s_eligible_dependent_id_list],
-            ["eligible_dependent_address_id_list", s_eligible_dependent_address_id_list],
+        # delete extraneous columns
```
zmbc, line 819: Much cleaner IMO to specify the columns that you _do_ want than to remove the ones you don't. I believe you could use `OUTPUT_COLUMNS` for this, as well.

src/vivarium_census_prl_synth_pop/components/observers.py
```diff
@@ -680,24 +798,120 @@ def get_observation(self, event: Event) -> pd.DataFrame:
...
+        pop = self.population_view.get(event.index)
+        household_details = self.pipelines["household_details"](pop.index)
+        pop[["address_id", "housing_type"]] = household_details[
+            ["address_id", "housing_type"]
+        ]
+
+        # add derived columns
+        pop["tax_year"] = event.time.year - 1
+
+        # delete extraneous columns
```
zmbc, line 903: Similar to above, why not `pop = pop[OUTPUT_COLUMNS]`?

src/vivarium_census_prl_synth_pop/components/observers.py
```diff
@@ -659,8 +778,7 @@ def get_observation(self, event: Event) -> pd.DataFrame:
 
         # not all simulants with a guardian are eligible to be dependents
         # need to know income of dependents in past year for this
-        dependent_w2 = df_w2[df_w2["simulant_id"].isin(df_eligible_dependents.index.unique())]
-        last_year_income = dependent_w2.groupby("simulant_id")["income"].sum()
+        last_year_income = self.w2_observer.income_last_year.groupby("simulant_id").sum()
```
zmbc, line 781: This seems like it is dependent on this observer running _after_ the W2 observer. Is that guaranteed by the order of `sub_components` above, or something else?

src/vivarium_census_prl_synth_pop/components/observers.py
```diff
@@ -680,24 +798,120 @@ def get_observation(self, event: Event) -> pd.DataFrame:
...
 
-        # re-initialize income-to-date series for next year of income counting
-        self.income_to_date = empty_income_series()
+        return df
 
-        return df_w2[self.output_columns]
+
+class Tax1040Observer(BaseObserver):
+    """Class for observing columns relevant to 1040 tax data (that is not
+    already recorded by W2 or dependents observer)
```
zmbc, line 837: If I'm understanding this code correctly, I think this comment is misleading. The only extra column observed here is `relation_to_household_head`.

Instead, the way I understand this observer is that it is different in the time dimension: it observes the taxpayers as of April instead of January, so things may have changed in the meantime.

src/vivarium_census_prl_synth_pop/components/observers.py
```diff
@@ -680,24 +798,120 @@ def get_observation(self, event: Event) -> pd.DataFrame:
...
+        household_details = self.pipelines["household_details"](pop.index)
+        pop[["address_id", "housing_type"]] = household_details[
+            ["address_id", "housing_type"]
+        ]
+
+        # add derived columns
+        pop["tax_year"] = event.time.year - 1
+
+        # delete extraneous columns
+        for col in ["alive", "guardian_1", "guardian_2", "race_ethnicity", "tracked"]:
```
zmbc, line 904: It doesn't seem wise to me to return the full population (including those who died/left the country) and not include which of them were alive and in the US. Without that information, their `age` and `address_id` columns will be misleading. Although it's not clear to me what the values in these columns should be, we'll at least want the option to treat these simulants differently, no?

src/vivarium_census_prl_synth_pop/components/observers.py
```diff
@@ -680,24 +798,120 @@ def get_observation(self, event: Event) -> pd.DataFrame:
...
+            del pop[col]
+
+        # to save space, remove rows for simulants with zero income
```
zmbc, line 907: I do not think this will support joint filing; what if one spouse has zero income and the other has enough that they need to file?

src/vivarium_census_prl_synth_pop/components/observers.py
```diff
@@ -680,24 +798,120 @@ def get_observation(self, event: Event) -> pd.DataFrame:
...
+    OUTPUT_COLUMNS = [
+        "first_name_id",
+        "middle_name_id",
+        "last_name_id",
+        "age",
+        "date_of_birth",
+        "sex",
+        "ssn",
+        "address_id",
+        "relation_to_household_head",
```
zmbc, line 856: My understanding is that `relation_to_household_head` is for joint filing. Can we make a note of that? And also that we do not need to include `household_id` since we are relying on the 1-to-1 mapping (on a single timestep) with `address_id`.

src/vivarium_census_prl_synth_pop/components/observers.py
```diff
@@ -649,6 +679,95 @@ def get_observation(self, event: Event) -> pd.DataFrame:
...
+
+        return df_w2[self.output_columns]
+
+
+class TaxDependentsObserver(BaseObserver):
+    """Class for observing columns relevant to identifying dependents in tax data.
+
+    This is most important for the 1040 data, but relies on data from
+    the W2 observer, and is better represented in a "long" form with a
+    row for each guardiant/dependent pair
```
zmbc, line 694:
```suggestion
    row for each guardian/dependent pair
```

src/vivarium_census_prl_synth_pop/components/observers.py
```diff
@@ -649,6 +679,95 @@ def get_observation(self, event: Event) -> pd.DataFrame:
...
+    def output_columns(self):
+        return self.OUTPUT_COLUMNS
+
+    def setup(self, builder: Builder):
+        super().setup(builder)
+        self.clock = builder.time.clock()
+
+    def to_observe(self, event: Event) -> bool:
+        """Observe if Jan 1 falls during this time step"""
+        tax_date = dt.datetime(event.time.year, 1, 1)
```
zmbc, line 748: It's a little difficult to think through the way this observer works. As you've noted below in `get_observation`, the technically correct thing would be to observe these dependent relationships throughout the tax year. But even with the approach we've chosen here of observing at the end of the tax year, we will report `age` and `address_id` for dependents on the 1040 (in April) as of January 1st. Let's note this limitation somewhere.

src/vivarium_census_prl_synth_pop/components/observers.py
aflaxman: > What is the situation in which these columns are not already present? They seem like they all should be.

Huh, good point.  I didn't think that through!

src/vivarium_census_prl_synth_pop/components/observers.py
aflaxman: > Much cleaner IMO to specify the columns that you do want than to remove the ones you don't. I believe you could use OUTPUT_COLUMNS for this, as well.

Agree, I'll change to that.

src/vivarium_census_prl_synth_pop/components/observers.py
aflaxman: > Similar to above, why not pop = pop[OUTPUT_COLUMNS]?

Just bad taste on my part, I will change

src/vivarium_census_prl_synth_pop/components/observers.py
aflaxman: > This seems like it is dependent on this observer running after the W2 observer. Is that guaranteed by the order of sub_components above, or something else?

Correct, I have added a comment about this subtle point.

src/vivarium_census_prl_synth_pop/components/observers.py
aflaxman: > If I'm understanding this code correctly, I think this comment is misleading. The only extra column observed here is relation_to_household_head.
> 
> Instead, the way I understand this observer is that it is different in the time dimension: it observes the taxpayers as of April instead of January, so things may have changed in the meantime.

Good point, I have updated the docstring

src/vivarium_census_prl_synth_pop/components/observers.py
aflaxman: > It doesn't seem wise to me to return the full population (including those who died/left the country) and not include which of them were alive and in the US.

Good call, I'll add them to the output

src/vivarium_census_prl_synth_pop/components/observers.py
aflaxman: > I do not think this will support joint filing; what if one spouse has zero income and the other has enough that they need to file?

good point.

src/vivarium_census_prl_synth_pop/components/observers.py
aflaxman: > My understanding is that relation_to_household_head is for joint filing. Can we make a note of that? And also that we do not need to include household_id since we are relying on the 1-to-1 mapping (on a single timestep) with address_id.

good call, will do.

src/vivarium_census_prl_synth_pop/components/observers.py
aflaxman: good catch.  i will add it myself to avoid needing an additional merge on my side.

src/vivarium_census_prl_synth_pop/components/observers.py
aflaxman: > It's a little difficult to think through the way this observer works. As you've noted below in get_observation, the technically correct thing would be to observe these dependent relationships throughout the tax year. But even with the approach we've chosen here of observing at the end of the tax year, we will report age and address_id for dependents on the 1040 (in April) as of January 1st. Let's note this limitation somewhere.

I was thinking that it was the dependent's age and address for the previous year that was relevant, did I misunderstand that?

src/vivarium_census_prl_synth_pop/components/observers.py
```diff
@@ -649,6 +679,95 @@ def get_observation(self, event: Event) -> pd.DataFrame:
...
+        pop_full[["address_id", "housing_type"]] = household_details[
+            ["address_id", "housing_type"]
+        ]
+
+        # Tracked, US population to be dependents
+        pop = pop_full[
+            (
+                pop_full["alive"] == "alive"
+            )  # really should include people who died in last year
+            & pop_full["tracked"]  # ??? should we include untracked, too?
```
rmudambi, line 763: The only way a simulant can become untracked is by emigrating so `in_united_states` and `tracked` are identical, but I think it's good to check both.

src/vivarium_census_prl_synth_pop/components/observers.py
rmudambi: It is not guaranteed. We can guarantee it by settings a priority on the listener, but a better solution is to set last year's income on time-step cleanup so it's guaranteed to happen first (time-step cleanup happens before collect metrics).

src/vivarium_census_prl_synth_pop/components/observers.py
zmbc: Ah, good question. I did not think about that.

All I can find in the docs is:

> Sampling on a single time step is not representative of how tax documents are compiled.

The observer as you have it here observes the filer's age and address as of April and the dependents' age and address as of January. It doesn't seem like those can both be right, but I'm not sure which one is the limitation! Maybe you can just add a note to that effect.

src/vivarium_census_prl_synth_pop/components/observers.py
```diff
@@ -649,6 +679,95 @@ def get_observation(self, event: Event) -> pd.DataFrame:
         for col in ["employer_address_id", "employer_name"]:
             df_w2[col] = df_w2["employer_id"].map(emp[col])
 
+        # re-initialize income-to-date series for next year of income counting
+        self.income_last_year = self.income_this_year
```
rmudambi, line 683: Also, I just remembered that I think the whole point (or one of the main points) of combining these observers was so that the base TaxObserver would manage income, not the W2.

src/vivarium_census_prl_synth_pop/components/observers.py
aflaxman: I will use the `time_step__cleanup` approach, it seems like the safest and clearest way.

src/vivarium_census_prl_synth_pop/components/observers.py
aflaxman: Good idea, I will add a note in the observer's docstring

src/vivarium_census_prl_synth_pop/components/observers.py
```diff
@@ -495,8 +495,36 @@ def get_observation(self, event: Event) -> pd.DataFrame:
...
+
+    def __init__(self):
+        tax_w2 = TaxW2Observer()
+        tax_1040 = Tax1040Observer(tax_w2)
+        tax_dependents = TaxDependentsObserver(tax_w2)
+
+        self._sub_components = [
+            tax_w2,
+            tax_1040,
+            tax_dependents,  # NOTE: this component must come after the TaxW2Observer to function correctly
```
zmbc, line 513: We should remove these notes now, I think?

zmbc (approved):

rmudambi (approved):

src/vivarium_census_prl_synth_pop/components/observers.py
aflaxman: good call

src/vivarium_census_prl_synth_pop/components/observers.py
aflaxman: I was thinking of the income as fundamentally a part of the W2, but we can refactor this in a future PR if you think it will be clearer to move it into the TaxObserver.

PR #182, opened 2023-01-30
move standard household intialization into the households component

rmudambi:
### Description

- *Category*: refactor
- *JIRA issue*: [MIC-3802](https://jira.ihme.washington.edu/browse/MIC-3802)
- *Research reference*: 

### Changes and notes
* Initialize households data structure with standard households in the `Households` component rather than the `Population` component
* Sample households from the artifact to pave the way for not having to carry that data around in memory
* On initialize simulants drop all households that don't have simulants initialized into the sim
* Added setters and getters to centralize access to the households component (was helpful in debugging as I could put a breakpoint in the setter and see every time it was updated).
* Fixed a few more places where the household details pipeline is accessed using the `Household` component
* Modified `vectorized_choice` to function in `setup` (i.e. without a `RandomnessStream`)

### Verification and Testing
Ran simulation and verified tests all pass.

### DIFF
--- a/src/vivarium_census_prl_synth_pop/components/household.py
+++ b/src/vivarium_census_prl_synth_pop/components/household.py
@@ -106,7 +106,7 @@ class Households:
 
         households = pd.concat([gq_households, sampled_standard_households])
         households["housing_type"] = households["housing_type"].astype(
-            pd.CategoricalDtype(categories=data_values.HOUSING_TYPES)
+            HOUSEHOLD_DETAILS_DTYPES["housing_type"]
         )
 
         return households
@@ -138,7 +138,7 @@ class Households:
     def get_household_details(self, idx: pd.Index) -> pd.DataFrame:
         """Source of the household_details pipeline"""
         pop = self.population_view.subview(["tracked", "household_id"]).get(idx)
-        household_details = pop.join(
+        household_details = pop[["household_id"]].join(
             self._households,
             on="household_id",
         )
--- a/src/vivarium_census_prl_synth_pop/components/household.py
+++ b/src/vivarium_census_prl_synth_pop/components/household.py
@@ -4,10 +4,14 @@ from vivarium.framework.engine import Builder
 from vivarium.framework.population import SimulantData
 from vivarium.framework.time import get_time_stamp
 
-from vivarium_census_prl_synth_pop.constants.data_values import (
-    GQ_HOUSING_TYPE_MAP,
-    HOUSING_TYPES,
-)
+from vivarium_census_prl_synth_pop.constants import data_keys, data_values
+from vivarium_census_prl_synth_pop.utilities import vectorized_choice
+
+HOUSEHOLD_DETAILS_DTYPES = {
+    "household_id": int,
+    "address_id": int,
+    "housing_type": pd.CategoricalDtype(categories=data_values.HOUSING_TYPES),
+}
 
 
 class Households:
@@ -29,6 +33,8 @@ class Households:
     #################
 
     def setup(self, builder: Builder):
+        self.pop_config = builder.configuration.population
+        self.seed = builder.configuration.randomness.random_seed
         self.start_time = get_time_stamp(builder.configuration.time.start)
 
         self.randomness = builder.randomness.get_stream(self.name)
@@ -39,7 +45,7 @@ class Households:
 
         self.population_view = builder.population.get_view(self.columns_used)
 
-        self.households = None
+        self._households = self.sample_initial_households(builder)
         self.household_details = builder.value.register_value_producer(
             "household_details",
             source=self.get_household_details,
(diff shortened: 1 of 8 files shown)


src/vivarium_census_prl_synth_pop/components/household.py
```diff
@@ -51,47 +57,100 @@ def setup(self, builder: Builder):
             requires_columns=["household_id"],
         )
 
-    ########################
-    # Event-driven methods #
-    ########################
-
-    def on_initialize_simulants(self, pop_data: SimulantData) -> None:
+    def sample_initial_households(self, builder: Builder) -> pd.DataFrame:
```
zmbc, line 60: I'm curious: did you consider using the `set_households` method within `population.py` to keep this logic there? IMHO, it fits better there and at population initialization time than at setup time.

I'm pretty sure I can make household immigration work with this approach, but I haven't fully thought it through. Certainly in that case, both sampling of households and initialization of simulants will need to happen in one atomic action.

src/vivarium_census_prl_synth_pop/components/household.py
```diff
@@ -51,47 +57,100 @@ def setup(self, builder: Builder):
...
+                "address_id": standard_household_ids,
+                "housing_type": "Standard",
+            }
         )
-        # set housing type dtype
+
+        households = pd.concat([gq_households, sampled_standard_households])
         households["housing_type"] = households["housing_type"].astype(
-            pd.CategoricalDtype(categories=HOUSING_TYPES)
+            pd.CategoricalDtype(categories=data_values.HOUSING_TYPES)
```
zmbc, line 109:
```suggestion
            HOUSEHOLD_DETAILS_DTYPES["housing_type"]
```

src/vivarium_census_prl_synth_pop/components/household.py
```diff
@@ -51,47 +57,100 @@ def setup(self, builder: Builder):
...
 
         return households
 
+    ########################
+    # Event-driven methods #
+    ########################
+
+    def on_initialize_simulants(self, pop_data: SimulantData) -> None:
+        """
+        Remove unused households from the household datastructure immediately
```
zmbc, line 120: Why would there be unused households?

src/vivarium_census_prl_synth_pop/components/household.py
```diff
@@ -51,47 +57,100 @@ def setup(self, builder: Builder):
...
+        following initialization and set household id as index
+        """
+        if pop_data.creation_time < self.start_time:
+            households = self._households.set_index("household_id")
```
zmbc, line 124: Why not do this at `setup`?

src/vivarium_census_prl_synth_pop/components/household.py
```diff
@@ -51,47 +57,100 @@ def setup(self, builder: Builder):
...
+            self.set_households(households.loc[gq_index.union(existing_households)])
+
+    ##################################
+    # Pipeline sources and modifiers #
+    ##################################
+
     def get_household_details(self, idx: pd.Index) -> pd.DataFrame:
         """Source of the household_details pipeline"""
-        pop = self.population_view.get(idx)[["household_id"]]
+        pop = self.population_view.subview(["tracked", "household_id"]).get(idx)
```
zmbc, line 140: Won't this add a `tracked` column within the pipeline?

src/vivarium_census_prl_synth_pop/components/population.py
```diff
@@ -161,32 +158,17 @@ def generate_initial_population(self, pop_data: SimulantData) -> None:
...
-                    len(chosen_households) + data_values.N_GROUP_QUARTER_TYPES,
-                ),
-            }
-        )
+    def populate_standard_households(self) -> pd.DataFrame:
+        households = self.households.get_households()
+        target_number_sims = households.index.size
 
+        standard_households = households.loc[
+            households["census_household_id"] != "N/A",
```
zmbc, line 166: Why not use "housing_type" here?

zmbc (commented):
Side-note: In general it's nice to avoid force-pushing while the PR is open/being reviewed since it invalidates any existing review comments.

src/vivarium_census_prl_synth_pop/components/household.py
rmudambi: Ordering of setup operations is not guaranteed

src/vivarium_census_prl_synth_pop/components/household.py
rmudambi: You're right. I should remove this

src/vivarium_census_prl_synth_pop/components/population.py
rmudambi: I have actually already fixed this in the next pr

src/vivarium_census_prl_synth_pop/components/household.py
rmudambi: We oversample households (i.e. sample as many households as we want people living in standard housing) so that we are guaranteed to have enough. We could be smarter and sample fewer, but it hasn't been a bottleneck at these population sizes. This means many sampled households will have no simulants in them.

src/vivarium_census_prl_synth_pop/components/household.py
zmbc: Note for posterity:

* One end goal of these refactors is to avoid having to keep the ACS households data file in memory in the `Population` component for the whole duration of the sim.
* We'd prefer to only have households updates in "household" components and keep them out of `Population`. (Although `PersonMigration` and (future) `Immigration` components aren't explicitly household-focused, we consider them "household" components.)
* The analogous approach to this for household immigration will be to `set_households` with the new households being created before calling the simulant creator function, keeping the two updates decoupled while doing them as an atomic action (since simulant creation happens immediately).

zmbc (approved):

PR #181, opened 2023-01-30
move household data-structure and pipeline to separate component

rmudambi:
### Description

- *Category*: refactor
- *JIRA issue*: [MIC-3801](https://jira.ihme.washington.edu/browse/MIC-3801)
- *Research reference*: 

### Changes and notes
Creates a new component to manage the households data structure and household details pipeline. 
Fixes typing in a  utilities function
Changed pipeline access in the Businesses component via  the Household component to access via the pipeline name.

### Verification and Testing
Ran sim successfully and all tests pass.

PR #184, opened 2023-01-31
Refactor households state into its own component

zmbc:
### Description
- *Category*: refactor
- *JIRA issue*: [MIC-3801](https://jira.ihme.washington.edu/browse/MIC-3801)
- *Research reference*: none

### Changes and notes

Builds on #181, but pares down the household component's public API. New households are initialized in the same way at population setup as they are on migration (different approach than #182). GQ households are initialized in setup of the households component (different approach than #183).

I suggest we squash this PR into one commit on `main`, because the path we took to get here is not especially enlightening.

### Verification and Testing

Integration and unit tests all passing.

### DIFF
--- a/src/vivarium_census_prl_synth_pop/components/household.py
+++ b/src/vivarium_census_prl_synth_pop/components/household.py
@@ -1,14 +1,13 @@
 import numpy as np
 import pandas as pd
 from vivarium.framework.engine import Builder
-from vivarium.framework.population import SimulantData
 from vivarium.framework.time import get_time_stamp
 
-from vivarium_census_prl_synth_pop.constants import data_keys, data_values
-from vivarium_census_prl_synth_pop.utilities import vectorized_choice
+from vivarium_census_prl_synth_pop.constants import data_values
+from vivarium_census_prl_synth_pop.utilities import update_address_ids
 
+HOUSEHOLD_ID_DTYPE = int
 HOUSEHOLD_DETAILS_DTYPES = {
-    "household_id": int,
     "address_id": int,
     "housing_type": pd.CategoricalDtype(categories=data_values.HOUSING_TYPES),
 }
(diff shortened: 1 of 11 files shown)


mattkappel (approved):

src/vivarium_census_prl_synth_pop/components/household.py
```diff
@@ -0,0 +1,148 @@
...
+            address_id_col_name="address_id",
+        )
+
+    ##################
+    # Helper methods #
+    ##################
+
+    @staticmethod
+    def _next_available_ids(num_ids: int, taken: pd.Series):
+        return Households._first_int_available(taken) + np.arange(num_ids)
```
rmudambi, line 141: Note: this approach is assuming households are never deleted from the households data structure (they aren't, but we should be clear that we are relying on this).

src/vivarium_census_prl_synth_pop/components/household.py
```diff
@@ -0,0 +1,148 @@
...
+            )
+        )
+        household_details = household_details.astype(HOUSEHOLD_DETAILS_DTYPES)
+        return household_details
+
+    ##################
+    # Public methods #
+    ##################
+
+    def create_households(
```
rmudambi, line 83: I like the access restrictions you've introduced here.

src/vivarium_census_prl_synth_pop/components/household.py
```diff
@@ -0,0 +1,148 @@
...
+        """
+        Changes the address_id associated with each household_id in household_ids
+        to a new, unique value.
+
+        Parameters
+        ----------
+        household_ids
+            The IDs of the households that should move to new addresses.
+        """
+        self._households, _ = update_address_ids(
```
rmudambi, line 128: I'm a bit conflicted about the utility of this function. It's used twice - here and in businesses, but there we need the max id and here we don't (and the function itself is quite simple). Would it be better just have the code duplicated rather than calling out to a function in utilities?

src/vivarium_census_prl_synth_pop/components/population.py
```diff
@@ -192,14 +194,25 @@ def choose_standard_households(self, target_number_sims: int) -> pd.DataFrame:
...
-        # get rid simulants in excess of desired pop size
+        # get rid of simulants and households in excess of desired pop size
         households_to_discard = chosen_persons.loc[
-            target_number_sims:, "household_id"
+            target_number_sims:, "census_sample_household_id"
         ].unique()
-
         chosen_persons = chosen_persons.loc[
-            ~chosen_persons["household_id"].isin(households_to_discard)
+            ~chosen_persons["census_sample_household_id"].isin(households_to_discard)
```
rmudambi, line 202: `households_to_discard` might actually be bigger than `households_to_keep`. If so, it would be more performant to do 
```
chosen_persons.loc[
    chosen_persons["census_sample_household_id"].isin(households_to_keep)
]
```
and same for households below.

src/vivarium_census_prl_synth_pop/components/population.py
```diff
@@ -173,14 +177,12 @@ def choose_standard_households(self, target_number_sims: int) -> pd.DataFrame:
...
+        # create unique id for resampled households -- each census_household_id
+        # can be sampled multiple times.
         chosen_households = pd.DataFrame(
             {
                 "census_household_id": chosen_households,
-                "household_id": np.arange(
-                    data_values.N_GROUP_QUARTER_TYPES,
-                    len(chosen_households) + data_values.N_GROUP_QUARTER_TYPES,
-                ),
+                "census_sample_household_id": np.arange(target_number_sims),
```
rmudambi, line 185: How is this different from our basic "household_id"?

src/vivarium_census_prl_synth_pop/utilities.py
```diff
@@ -272,7 +272,7 @@ def update_address_ids(
     units_that_move_ids: pd.Index,
     starting_address_id: int,
     address_id_col_name: str,
-) -> Tuple[pd.DataFrame, pd.DataFrame, int]:
+) -> Tuple[pd.DataFrame, int]:
```
rmudambi, line 275: Update docstring?

rmudambi (approved):

src/vivarium_census_prl_synth_pop/components/household.py
zmbc: Great point. I think I can write this method more simply/clearly without using the utilities function. That will leave just one use of the utilities function, so I can move it into the businesses component as a helper.

src/vivarium_census_prl_synth_pop/components/population.py
zmbc: IIRC, the mean household size was around 2, so I think they are about the same size. I'm going to keep this how it is because it seems easier to find `households_to_discard` than `households_to_keep`.

src/vivarium_census_prl_synth_pop/components/population.py
zmbc: It's not unique with respect to existing household_ids. Only the `Households` component can "mint" new unique household_ids, and it never deletes them, so we don't want to create one for the excess households.

A bit academic here because these are the first households, but this will be consequential for immigration, when there will be pre-existing households we don't want to collide with.

src/vivarium_census_prl_synth_pop/utilities.py
zmbc: I did! Sorry, I might not have done this until after you started your review 😅

PR #185, opened 2023-01-31
Do not keep ACS data longer than necessary

zmbc:
### Description

- *Category*: performance
- *JIRA issue*: none
- *Research reference*: none

### Changes and notes

This should free up the memory (once it is garbage collected) after we are done with the ACS data, which is quite large.

### Verification and Testing

Integration tests passing.

### DIFF
--- a/src/vivarium_census_prl_synth_pop/components/population.py
+++ b/src/vivarium_census_prl_synth_pop/components/population.py
@@ -66,6 +66,9 @@ class Population:
         ]
         self.register_simulants = builder.randomness.register_simulants
         self.population_view = builder.population.get_view(self.columns_created)
+
+        # HACK: ACS data must be loaded in setup, since it comes from the artifact;
+        # however, we only use it while creating the initial population.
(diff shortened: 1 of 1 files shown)


mattkappel:
Did you observe that garbage was collected? `mprof -CM simulate run ...` should produce a graph of usage over time if you're interested.

mattkappel (approved):

rmudambi (approved):

zmbc:
>Did you observe that garbage was collected?

Good call -- this appears to be working (population 250k, keep in mind that the ACS data is also about 1/5th of the country):

image

aflaxman:
Beka Stoerts’s recent PRL paper says they used a single machine with 2TB of memory.  Is 5G a big deal these days?

(and cool plot!)

zmbc:
> Is 5G a big deal these days?

No; but, it's actually 25G once we include the whole country, and keeping it in memory means it will be _added_ to the slow accumulation over time in the right side of the plot, which I expect will scale with population size.

PR #187, opened 2023-02-01
Extract methods for initializing simulants

zmbc:
### Description

- *Category*: refactor
- *JIRA issue*: [MIC-3777](https://jira.ihme.washington.edu/browse/MIC-3777)
- *Research reference*: none
 
### Changes and notes

This mostly just pulls out methods, which will be shared between population initialization and immigration.

Two changes worth highlighting:

* Instead of sampling from households, then joining to persons, then re-joining back to households for state and PUMA, I just join the entire households dataframe when sampling, so the columns are already there.
* This PR fixes a very minor bug: previously, we were perturbing _integer_ age straight from ACS and _then_ adding the random draw between 0 and 1. Now, we do it in the opposite order.

If you are wondering why `initialize_simulant_link_columns` isn't just part of `initialize_new_simulants_from_acs`, in the next PR I will extend that method to allow initializing those columns for only a _subset_ of the population, since that will be necessary for immigration.

### Verification and Testing

Integration tests passing; I also stepped through `generate_initial_population` -- I did find one unrelated bug which I will make a PR about shortly.

### DIFF
--- a/src/vivarium_census_prl_synth_pop/components/population.py
+++ b/src/vivarium_census_prl_synth_pop/components/population.py
@@ -966,6 +966,11 @@ class Population:
         simulants.loc[simulants["age"] < 0, "age"] = 0
         simulants.loc[simulants["age"] > 99, "age"] = 99
 
+        # Keep DOB in sync -- should this be a pipeline instead of a column?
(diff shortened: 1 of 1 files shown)


rmudambi (approved):
Looks great!

PR #195, opened 2023-02-04
Speed up PO box test

zmbc:
### Description

- *Category*: performance, test
- *JIRA issue*: none
- *Research reference*: none

### Changes and notes

Probably tons of other low-hanging fruit but I noticed this one because it was new.

### Verification and Testing

This test went from taking over 1.5 minutes (not including simulation running time, just the test body itself) to a few seconds.

### DIFF
--- a/integration_tests/test_domestic_migration.py
+++ b/integration_tests/test_domestic_migration.py
@@ -273,10 +273,10 @@ def test_po_box(tracked_live_populations):
         # Check that actual proportion of households without PO Boxes (i.e., physical
         # address is the same as mailing) is close to the expected proportion
         assert np.isclose(
-            pop[
-                pop["household_details.po_box"] == data_values.NO_PO_BOX
-            ].household_id.nunique()
-            / pop.household_id.nunique(),
+            pop[pop["household_details.po_box"] == data_values.NO_PO_BOX][
+                "household_id"
+            ].nunique()
+            / pop["household_id"].nunique(),
             data_values.PROBABILITY_OF_SAME_MAILING_PHYSICAL_ADDRESS,
             rtol=0.01,
         )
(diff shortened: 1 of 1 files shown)


integration_tests/test_domestic_migration.py
```diff
@@ -268,17 +268,15 @@ def test_addresses_during_moves(simulants_on_adjacent_timesteps, unit_id_col, ad
...
         assert np.isclose(
-            len(
-                pop[pop["household_details.po_box"] == data_values.NO_PO_BOX].groupby(
-                    "household_id"
-                )
-            )
-            / len(pop.groupby("household_id")),
+            pop[
+                pop["household_details.po_box"] == data_values.NO_PO_BOX
+            ].household_id.nunique()
```
rmudambi, line 278: Use bracket notation rather than dot notation in both of these lines

rmudambi (approved):
Thanks for catching this!

integration_tests/test_domestic_migration.py
mattkappel: +1

mattkappel:
Thanks much. TIL.

mattkappel (approved):

mattkappel:
It still seems weird to me that `nunique` is more efficient than `len` but perhaps that's because I'm doing it on the entire data frame rather than a series.

aflaxman:
> It still seems weird to me that `nunique` is more efficient than `len` but perhaps that's because I'm doing it on the entire data frame rather than a series.

I think the speed up comes from avoiding the `.groupby`

integration_tests/test_domestic_migration.py
zmbc: Ah duh, thanks for the reminder!

PR #196, opened 2023-02-04
Move ACS sampling to utilities

zmbc:
### Description

- *Category*: refactor
- *JIRA issue*: [MIC-3810](https://jira.ihme.washington.edu/browse/MIC-3810)
- *Research reference*: none

### Changes and notes

Splitting this out so it can be used from Immigration.

### Verification and Testing

Integration tests passing except for one, which is unrelated: see #197.

### DIFF
--- a/src/vivarium_census_prl_synth_pop/components/population.py
+++ b/src/vivarium_census_prl_synth_pop/components/population.py
@@ -12,9 +12,9 @@ from vivarium_public_health.utilities import DAYS_PER_YEAR, to_years
 
 from vivarium_census_prl_synth_pop.constants import data_keys, data_values, metadata
 from vivarium_census_prl_synth_pop.utilities import (
-    vectorized_choice,
-    sample_acs_standard_households,
     sample_acs_group_quarters,
+    sample_acs_standard_households,
+    vectorized_choice,
 )
 
 # Family/household relationships helper lists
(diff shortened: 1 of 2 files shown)


src/vivarium_census_prl_synth_pop/components/population.py
```diff
@@ -197,32 +175,12 @@ def sample_simulants_from_standard_households(
...
-        chosen_persons = pd.merge(
-            chosen_units,
-            acs_persons[metadata.PERSONS_COLUMNS_TO_INITIALIZE],
-            on="census_household_id",
-            how="left",
-        )
-
-        new_simulants = self.initialize_new_simulants_from_acs(chosen_persons, pop_data)
+        new_simulants = self.initialize_new_simulants_from_acs(acs_persons, pop_data)
         new_simulants["age"] = self.perturb_individual_age(new_simulants)
```
rmudambi, line 184: Is there a reason we're calling `perturb_individual_age` two separate times rather than just once right before we update the pop view?

rmudambi (approved):

src/vivarium_census_prl_synth_pop/components/population.py
zmbc: If I'm understanding the question correctly, yes. When initializing the population, the non-GQ population uses `self.perturb_household_age` which makes a single shift at the household level. The GQ population is individually perturbed (`self.perturb_individual_age`).

Also, note that guardians depend on age. So initializing link columns is another step that must be done between age perturbation and updating the pop view.

rmudambi: Thanks for that. I misread this and thought they were both the same function call, but one is `perturb_household_age` and the other is `perturb_individual_age`. Carry on.

PR #199, opened 2023-02-04
Implement household immigration

zmbc:
### Description
- *Category*: implementation
- *JIRA issue*: [MIC-3778](https://jira.ihme.washington.edu/browse/MIC-3778)
- *Research reference*: https://vivarium-research.readthedocs.io/en/latest/models/concept_models/vivarium_census_synthdata/concept_model.html#id8

### Changes and notes

**Branched off #174, review that first!**

### Verification and Testing

Integration tests all passing except as noted in #197. I added a bit of testing but I couldn't think of anything to assert about household immigrants. The household structure tests cover a lot.


src/vivarium_census_prl_synth_pop/components/immigration.py
```diff
@@ -130,6 +131,30 @@ def on_time_step(self, event: Event):
             event.index,
         )
 
+        num_household_immigrants = self._round_stochastically(
```
rmudambi, line 134: We don't want to do this before individual immigration and then fill in the gaps when we get slightly too few household immigrants due to the last household straddling the threshold of immigrants we want? In initialization we fill that with extra GQ simulants. Are we sure we don't want to fill the gap here, especially since immigration rates are so low?

rmudambi: Also, what values do we see for `self.household_immigrants_per_time_step`. If we have a value between 0 and 1, doesn't that mean we can't ever have a household immigrate with size greater than 1? We might want to draw from a different distribution that also has a mean of `self.household_immigrants_per_time_step`.

zmbc:
This is currently blocked by review/any revisions to https://github.com/ihmeuw/vivarium_research/pull/1134

src/vivarium_census_prl_synth_pop/components/immigration.py
```diff
@@ -60,24 +61,25 @@ def setup(self, builder: Builder):
...
             households_data["census_household_id"].isin(
                 immigrant_reference_people["census_household_id"]
             )
         ]
+        self.households_immigrating_per_time_step = self._rescale_acs_value(
+            self.household_immigrant_households["household_weight"].sum(),
+            builder.configuration,
+        )
+
         # FIXME -- this can go away once state and PUMA are in the households pipeline, because
```
rmudambi, line 83: Hasn't this been done?

src/vivarium_census_prl_synth_pop/components/immigration.py
```diff
@@ -133,21 +135,42 @@ def on_time_step(self, event: Event):
             event.index,
         )
 
+        num_households_immigrating = self._round_stochastically(
```
rmudambi, line 138: Why are GQ and individual immigration extracted to separate methods while household is not?

src/vivarium_census_prl_synth_pop/utilities.py
```diff
@@ -287,27 +287,38 @@ def convert_middle_name_to_initial(pop: pd.DataFrame) -> pd.DataFrame:
 
 
 def sample_acs_standard_households(
-    target_number_sims: int,
+    num_households: Optional[int],
     acs_households: pd.DataFrame,
     acs_persons: pd.DataFrame,
     randomness: RandomnessStream,
+    target_number_sims: Optional[int] = None,
```
rmudambi, line 294: You shouldn't change the order of the parameters unless you really need to, and you don't in this case.

rmudambi (approved):

src/vivarium_census_prl_synth_pop/components/immigration.py
zmbc: Only because they had more in common with each other. I've extracted household immigration to a method, which I agree is nicer.

src/vivarium_census_prl_synth_pop/components/immigration.py
zmbc: Good point. I should have corrected this in #174. I will make another PR.

PR #202, opened 2023-02-08
add state/puma to household_details pipeline

stevebachmeier:
### Description
- *Category*: implementation
- *JIRA issue*: [MIC-3728](https://jira.ihme.washington.edu/browse/MIC-3728)
- *Research reference*: na
- 
### Changes and notes
This adds state_id and puma to the household_details pipeline.
NOTES:
 - It does remove the cols (in the next PR)
 - It does not do the equivalent for the business_details pipeline (in a separate task)

### Verification and Testing
Pytests have been updated to use the pipeline and all pass.

### DIFF
--- a/integration_tests/test_domestic_migration.py
+++ b/integration_tests/test_domestic_migration.py
@@ -193,7 +193,12 @@ def test_only_living_people_change_households(simulants_on_adjacent_timesteps):
             "household_details.state_id",
             "household_details.puma",
         ),
-        ("household_id", "household_details.po_box", "household_details.state_id", "household_details.puma"),
+        (
+            "household_id",
+            "household_details.po_box",
+            "household_details.state_id",
+            "household_details.puma",
+        ),
     ],
 )
 def test_unit_members_share_address(
@@ -218,7 +223,12 @@ def test_unit_members_share_address(
             "household_details.state_id",
             "household_details.puma",
         ),
-        ("household_id", "household_details.po_box", "household_details.state_id", "household_details.puma"),
+        (
+            "household_id",
+            "household_details.po_box",
+            "household_details.state_id",
+            "household_details.puma",
+        ),
     ],
 )
 def test_unit_address_uniqueness(
@@ -235,7 +245,6 @@ def test_unit_address_uniqueness(
             assert (pop.groupby(address_cols)[unit_id_col].nunique() == 1).all()
 
 
-
 @pytest.mark.parametrize(
     "state_id_col, puma_col",
     [
@@ -243,15 +252,14 @@ def test_unit_address_uniqueness(
         ("household_details.state_id", "household_details.puma"),
     ],
 )
-def test_pumas_exist_in_states(
-    sim, populations, state_id_col, puma_col
-):
+def test_pumas_exist_in_states(sim, populations, state_id_col, puma_col):
     """Check that PUMAs map to correct state.
 
     NOTE: This is an imperfect test because PUMAs are only unique within states
     and so one PUMA can exist in multiple states. Here we only ensure no
     impossible PUMAs exist in each state
     """
+    # FIXME: Use the US households static file for states/pumas
     state_puma_map = (
         sim._data.artifact.load("population.households")
         .reset_index()
--- a/integration_tests/test_domestic_migration.py
+++ b/integration_tests/test_domestic_migration.py
@@ -4,8 +4,6 @@ import pytest
 
 from vivarium_census_prl_synth_pop.constants import data_values
 
-# TODO: Broader test coverage
-
 
 def test_individuals_move(simulants_on_adjacent_timesteps):
     for before, after in simulants_on_adjacent_timesteps:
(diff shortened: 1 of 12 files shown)


integration_tests/test_domestic_migration.py
```diff
@@ -239,32 +252,33 @@ def test_addresses_during_moves(simulants_on_adjacent_timesteps, unit_id_col, ad
...
-        # # Some movers are in a new state.
-        # # assert any(before.loc[mask_movers, "state"] != after.loc[mask_movers, "state"])
-
-        # # Most movers are in a different puma
-        # assert (
-        #     before.loc[mask_movers, "puma"] != after.loc[mask_movers, "puma"]
-        # ).sum() / mask_movers.sum() >= 0.95
+        # TODO Check that *most* are in a new state when we add all locations
+        # Some movers are in a new state.
+        # assert any(before.loc[mask_movers, "state"] != after.loc[mask_movers, "state"])
```
zmbc, line 273: Does this need to still be commented?

integration_tests/test_domestic_migration.py
```diff
@@ -185,50 +183,65 @@ def test_only_living_people_change_households(simulants_on_adjacent_timesteps):
...
-@pytest.mark.skip(reason="Wait for state/puma to be added (MIC-3728)")
-def test_addresses_during_moves(simulants_on_adjacent_timesteps, unit_id_col, address_id_col):
-    """Check that unit (household and business) address details change after a move."""
-    other_address_cols = []  # TODO: add state/puma
-    address_cols = [address_id_col] + other_address_cols
+def test_addresses_during_moves(
+    simulants_on_adjacent_timesteps, unit_id_col, address_id_col, state_puma_cols
+):
+    """Check that unit (household and business) address details change after a move.
+    NOTE: This test requires a rather large pop size (>~200k) to work since the
```
zmbc, line 241: Which part of the test fails? Can we just make it more conservative? For example if it's the 0.001 or 0.95 below let's just change those numbers. I'd prefer not to have a test that always fails with a smaller population.

integration_tests/test_domestic_migration.py
```diff
@@ -239,32 +252,33 @@ def test_addresses_during_moves(simulants_on_adjacent_timesteps, unit_id_col, ad
...
-        # assert (
-        #     before.loc[mask_movers, "puma"] != after.loc[mask_movers, "puma"]
-        # ).sum() / mask_movers.sum() >= 0.95
+        # TODO Check that *most* are in a new state when we add all locations
+        # Some movers are in a new state.
+        # assert any(before.loc[mask_movers, "state"] != after.loc[mask_movers, "state"])
+        # Most movers are in a different PUMA
+        puma_col = [x for x in state_puma_cols if "puma" in x]
+        if puma_col:  # TODO: remove conditioinal when puma/state implemented for businesses
+            puma_col = puma_col[0]
```
zmbc, line 277: This is a bit odd; maybe just do this "most" check using **all** the `state_puma_cols` and do the "some" check on each individual column in `state_puma_cols`. Then you don't have to explicitly split this apart.

setup.py
```diff
@@ -4,7 +4,6 @@
 from setuptools import find_packages, setup
 
 if __name__ == "__main__":
-
```
zmbc, line 7: nit: This diff looks extraneous?

src/vivarium_census_prl_synth_pop/components/household.py
```diff
@@ -135,13 +166,23 @@ def update_household_addresses(self, household_ids: pd.Series) -> None:
...
         self._households.loc[household_ids, "address_id"] = new_address_ids
         self._households.loc[household_ids, "po_box"] = po_boxes
+        self._households.loc[household_ids, ["state_id", "puma"]] = states_pumas
 
     ##################
     # Helper methods #
     ##################
 
+    @staticmethod
+    def _get_state_puma_options(df: pd.DataFrame) -> pd.Series:
```
zmbc, line 179: Could this method not be replaced with `df[["state", "puma"]].drop_duplicates()`? I think it will even be cleaner below to have it as a DataFrame instead of a list when using `vectorized_choice`.

src/vivarium_census_prl_synth_pop/components/household.py
```diff
@@ -107,12 +135,15 @@ def create_households(
             num_households, taken=self._households["address_id"]
         )
         po_boxes = self.generate_po_boxes(num_households)
+        states_pumas = self.generate_states_pumas(household_ids)
```
zmbc, line 138: This doesn't seem right because it throws away information about where the ACS-sampled households actually were! I think `states_pumas` should be an argument to the `create_households` method.

And likewise, `generate_states_pumas` should be called `sample_random_states_pumas` or something to that effect.

src/vivarium_census_prl_synth_pop/components/household.py
```diff
@@ -80,6 +93,21 @@ def get_household_details(self, idx: pd.Index) -> pd.DataFrame:
         household_details = household_details.astype(HOUSEHOLD_DETAILS_DTYPES)
         return household_details
 
+    ########################
+    # Event-driven methods #
+    ########################
+
+    def on_initialize_simulants(self, pop_data: SimulantData) -> None:
+        "Sample the GQ household states/pumas on initialization"
```
zmbc, line 101: I don't see why to do this on initialization instead of in setup. Don't you already have everything you need when the GQ households are created in the first place?

src/vivarium_census_prl_synth_pop/components/household.py
zmbc: Perhaps you could still do this random sampling when the `states_pumas` argument is `None`, to streamline the new-household-migration case.

setup.py
mattkappel: I assume this is an inadvertent `black` reformat.

integration_tests/test_domestic_migration.py
stevebachmeier: The issue is that the `mask_moved_units` comes back empty. I thought about skipping the tests if it's empty but thought that might cover up bigger problems.

integration_tests/test_domestic_migration.py
stevebachmeier: Yeah, I'm waiting to add all 50+ locations

setup.py
stevebachmeier: Yep, just a black thing

src/vivarium_census_prl_synth_pop/components/household.py
stevebachmeier: Oof, yeah, this is a big mistake! I'll work on it

integration_tests/test_domestic_migration.py
zmbc: I'd skip if it comes back empty! That is more similar to what we do in other tests.

Actually the most similar to what I have done in emigration/immigration tests (which are also very rare events) would be to `continue` the for loop when `mask_moved_units` is empty, but accumulate `mask_moved_units` into a single big Series aggregated across timesteps and then at the end of the test, assert that it's not all False. I'd imagine that would make it work for a ~10x smaller population size, and 20k is what we've been considering our lower bound.

src/vivarium_census_prl_synth_pop/components/household.py
zmbc: Ah, I was wondering if it had to do with that. Ok, this is fine for now.

integration_tests/test_domestic_migration.py
zmbc: My interpretation of the above comment was that once we had 50, we would test that _most_ move -- but with only 2, it seems like we should be able to test that _some_ move? That said, some of us still regularly run the sim with Florida only. So maybe we should uncomment and explicitly gate on the location in the config?

stevebachmeier: Oh, I see - I just forgot to comment out the "some" movers check. Will fix and keep the TODO to change it later.

I'm going to keep the test as is b/c I'm pretty sure once this gets merged there's no reason to use florida

integration_tests/test_domestic_migration.py
stevebachmeier: ah I see what you mean

integration_tests/test_domestic_migration.py
```diff
@@ -185,50 +183,70 @@ def test_only_living_people_change_households(simulants_on_adjacent_timesteps):
...
+        (
+            "household_id",
+            "household_details.address_id",
+            ["household_details.state_id", "household_details.puma"],
+        ),
+        ("household_id", "household_details.po_box", []),
     ],
 )
-def test_unit_address_uniqueness(populations, unit_id_col, address_id_col):
+def test_unit_address_uniqueness(
```
rmudambi, line 198: Aren't these really three separate tests?

integration_tests/test_domestic_migration.py
```diff
@@ -185,50 +183,70 @@ def test_only_living_people_change_households(simulants_on_adjacent_timesteps):
...
-@pytest.mark.skip(reason="Wait for state/puma to be added (MIC-3728)")
-def test_addresses_during_moves(simulants_on_adjacent_timesteps, unit_id_col, address_id_col):
+def test_addresses_during_moves(
+    simulants_on_adjacent_timesteps, unit_id_col, address_id_col, state_id_col, puma_col
+):
     """Check that unit (household and business) address details change after a move."""
-    other_address_cols = []  # TODO: add state/puma
-    address_cols = [address_id_col] + other_address_cols
+    address_cols = [address_id_col, state_id_col, puma_col]
+    address_cols = [x for x in address_cols if x is not None]
```
rmudambi, line 248: Add a fixme that this line will be removed once business addresses have state/puma.

zmbc (approved):
Awesome! This will really simplify things!

albrja (approved):

src/vivarium_census_prl_synth_pop/components/household.py
```diff
@@ -32,12 +38,17 @@ def name(self):
...
     def setup(self, builder: Builder):
+        self.start_time = get_time_stamp(builder.configuration.time.start)
+        self.randomness = builder.randomness.get_stream(self.name)
         self.columns_used = [
             "tracked",
             "household_id",
         ]
 
         self.population_view = builder.population.get_view(self.columns_used)
+        self.state_puma_options = builder.data.load("population.households")[
```
rmudambi, line 49: We have a csv that contains all possible states and pumas. You can find the csv here: https://vivarium-research.readthedocs.io/en/latest/models/concept_models/vivarium_census_synthdata/concept_model.html#id27. Matt is already using it in ZIP code generation in post-processing, so sync up with him regarding the location where he's putting the file.

src/vivarium_census_prl_synth_pop/components/household.py
```diff
@@ -50,6 +61,11 @@ def setup(self, builder: Builder):
                 "address_id": data_values.GQ_HOUSING_TYPE_MAP.keys(),
                 "housing_type": data_values.GQ_HOUSING_TYPE_MAP.values(),
                 "po_box": data_values.NO_PO_BOX,
+                # Add dummy state_id and puma during setup b/c we do not have
```
rmudambi, line 64: We can actually modify vectorized choice to be able to function without a randomness stream and still preserve CRN. Let's talk tomorrow about how to do this.

rmudambi (commented):
Looks good. I have a few comments and something for to follow up on tomorrow morning, though.

src/vivarium_census_prl_synth_pop/utilities.py
```diff
@@ -169,15 +169,27 @@ def get_norm_from_quantiles(
...
     n_to_choose: int,
-    randomness_stream: RandomnessStream,
+    randomness_stream: RandomnessStream = None,
     weights: Array = None,
     additional_key: Any = None,
+    random_seed: int = None,
 ):
+    if not randomness_stream and not (str(additional_key) and str(random_seed)):
+        raise RuntimeError(
+            "An additonal_key and a ransom_seed are required in 'vectorized_choice'"
```
rmudambi, line 179: typo: "random" not "ransom" and "additional" not "additonal"

src/vivarium_census_prl_synth_pop/utilities.py
```diff
@@ -169,15 +169,27 @@ def get_norm_from_quantiles(
 def vectorized_choice(
     options: Array,
     n_to_choose: int,
-    randomness_stream: RandomnessStream,
+    randomness_stream: RandomnessStream = None,
     weights: Array = None,
     additional_key: Any = None,
+    random_seed: int = None,
 ):
+    if not randomness_stream and not (str(additional_key) and str(random_seed)):
```
rmudambi, line 177: `str(None) == "None"`. This check won't work how you want it to. Just do 
```
if not randomness_stream and (additional_key == None or random_seed == None):
```

rmudambi (approved):

PR #206, opened 2023-02-10
Add ZIP code mapping logic and unit tests

mattkappel:
### Description
- *Category*:  implementation
- *JIRA issue*: [MIC-3741](https://jira.ihme.washington.edu/browse/MIC-3741)
- *Research reference*: https://vivarium-research.readthedocs.io/en/latest/models/concept_models/vivarium_census_synthdata/concept_model.html#id26

### Changes and notes
- Adds PUMA to ZIP csv to paths (and some minor refactoring to raw_data paths)
- Adds an abstraction function for getting all `address_id` mappings to handle that multiple mappings are indexed by `address_id`
- Adds `address_id` to `zipcode` mapping function

### Verification and Testing
CI passes, all unit tests pass, and `make_artifacts` succeeds (tested for the `constants/paths.py` refactoring)


src/vivarium_census_prl_synth_pop/results_processing/addresses.py
```diff
@@ -18,6 +19,98 @@
...
+        RandomnessStream to use in choosing zipcodes proportionally
+
+    Returns
+    -------
+    A pd.Series suitable for pd.Series.map, indexed by column_name, with key "zipcode"
+
+    """
+    zip_map_dict = {}
+    output_cols = [column_name, "state", "puma"]  # columns in the output we use to map
+    address_ids = format_data_for_mapping(
```
rmudambi, line 78: Calling `format_data_for_mapping` separately inseide each of these child functions is inefficient, as we are doing this concat operation multiple times. It'd be better to do it once inside `get_address_id_maps` with all the columns needed by any of the individual parts of the address.

src/vivarium_census_prl_synth_pop/results_processing/addresses.py
```diff
@@ -18,6 +19,98 @@
...
+        .reset_index()
+        .set_index(["state", "puma"])
+    )
+    normalized_groupby = (
+        (map_data.set_index(["state", "puma", "zipcode"]) / proportions)
+        .reset_index()
+        .groupby(["state", "puma"])
+    )
+
+    for (state, puma), df_locale in address_ids.groupby(["state", "puma"]):
```
rmudambi, line 99: We can proceed with this for now, but this is a python loop over a large number of rows - there are a lot of state/puma combinations. I haven't looked at this long enough to have a solution off the top of my head, but I'd be surprised if there wasn't a vectorized operation we could do here rather than looping over the groupby in python code.

rmudambi (approved):

src/vivarium_census_prl_synth_pop/results_processing/addresses.py
mattkappel: I'll make this change. I'll need to coordinate with @albrja to update his unmerged branch based on this branch.

src/vivarium_census_prl_synth_pop/results_processing/addresses.py
mattkappel: I'll run this against real data and see how bad the inefficiency is and log to the profiling page.

PR #209, opened 2023-02-13
skip moving states pytest when only one state in artifact

stevebachmeier:
### Description
- *Category*: bugfix
- *JIRA issue*: na
- *Research reference*: na

### Changes and notes
If a one-state artifact is used, the test to ensure simulants are
moving states breaks; this skips that test.

### Verification and Testing
All tests pass

### DIFF
--- a/integration_tests/test_domestic_migration.py
+++ b/integration_tests/test_domestic_migration.py
@@ -2,7 +2,7 @@ import numpy as np
 import pandas as pd
 import pytest
 
-from vivarium_census_prl_synth_pop.constants import data_values
+from vivarium_census_prl_synth_pop.constants import data_values, metadata
 
 
 def test_individuals_move(simulants_on_adjacent_timesteps):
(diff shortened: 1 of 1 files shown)


integration_tests/test_domestic_migration.py
```diff
@@ -289,9 +289,12 @@ def test_pumas_exist_in_states(sim, populations, state_id_col, puma_col):
     ],
 )
 def test_addresses_during_moves(
-    simulants_on_adjacent_timesteps, unit_id_col, address_id_col, state_id_col, puma_col
+    sim, simulants_on_adjacent_timesteps, unit_id_col, address_id_col, state_id_col, puma_col
 ):
     """Check that unit (household and business) address details change after a move."""
+    number_of_states_in_artifact = (
```
zmbc, line 295: Is it not possible to get this information directly from the configuration? This way will work, but I imagine it may be a bit slow.

zmbc (approved):

rmudambi (approved):

PR #210, opened 2023-02-13
add employer state/puma to sim

stevebachmeier:
### Description
- *Category*: implementation
- *JIRA issue*: [MIC-3831](https://jira.ihme.washington.edu/browse/MIC-3831)
- *Research reference*: na

### Changes and notes
Adds `employer_state_id` and `employer_puma` to the business_details pipeline.
Also updates the w2 observer to include these new columns in the output.

NOTE: This also includes Jira task [MIC-3830](https://jira.ihme.washington.edu/browse/MIC-3830)

### Verification and Testing
All pytests pass and a short sim run.

### DIFF
--- a/src/vivarium_census_prl_synth_pop/components/household.py
+++ b/src/vivarium_census_prl_synth_pop/components/household.py
@@ -8,6 +8,7 @@ from vivarium.framework.time import get_time_stamp
 
 from vivarium_census_prl_synth_pop.constants import data_values
 from vivarium_census_prl_synth_pop.utilities import (
+    get_state_puma_options,
     random_integers,
     randomly_sample_states_pumas,
 )
@@ -43,9 +44,7 @@ class Households:
     def setup(self, builder: Builder):
         self.start_time = get_time_stamp(builder.configuration.time.start)
         self.randomness = builder.randomness.get_stream(self.name)
-        self.states_in_artifact = list(
-            builder.data.load("population.households")["state"].drop_duplicates()
-        )
+        self.state_puma_options = get_state_puma_options(builder)
         self.columns_used = [
             "tracked",
             "household_id",
@@ -62,7 +61,7 @@ class Households:
         )
         states_pumas = randomly_sample_states_pumas(
             unit_ids=gq_household_ids,
-            available_states=self.states_in_artifact,
+            state_puma_options=self.state_puma_options,
             additional_key="gq_states_pumas",
             random_seed=builder.configuration.randomness.random_seed,
         )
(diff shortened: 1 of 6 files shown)


src/vivarium_census_prl_synth_pop/components/household.py
```diff
@@ -227,22 +243,3 @@ def generate_po_boxes(self, num_addresses) -> np.array:
             additional_key="po_box_number",
         )
         return np.array(po_boxes)
-
-    def randomly_sample_states_pumas(
```
stevebachmeier, line 231: NOTE: Since this now gets called by two different components (household and business), I moved it into utilities.py. It's very slightly more than a copy/paste job, though, so I'd recommend reviewing the function there.

zmbc:
> NOTE: This also includes Jira task [MIC-3830](https://jira.ihme.washington.edu/browse/MIC-3830)

Am I wrong that this will start applying states and PUMAs to people besides those actually included in the artifact? Is that an intended change?

stevebachmeier:
> > NOTE: This also includes Jira task [MIC-3830](https://jira.ihme.washington.edu/browse/MIC-3830)
> 
> Am I wrong that this will start applying states to people besides those actually included in the artifact? E.g. I think now someone could move to Arizona even if the artifact is just Florida and California. Is that an intended change?

@zmbc It is protected against moving people into states not in the artifact (the `randomly_sample_states_pumas` method now includes an `available_states` argument). However, there is no protection against moving a person into some PUMA in a given state if that particular state/PUMA combo does not exist in the ACS data. I think that's fine, though.

src/vivarium_census_prl_synth_pop/utilities.py
```diff
@@ -385,3 +385,51 @@ def sample_acs_group_quarters(
...
+
+    Returns:
+        pd.DataFrame: Index are the unit_ids and has columns [['state_id', 'puma']]
+    """
+    if random_seed is None and randomness_stream is None:
+        raise RuntimeError("A randomness_stream or random_seed must be provided")
+    if random_seed is not None and randomness_stream is not None:
+        raise RuntimeError("Only one of randomness_stream or random_seed should be provided")
+
+    state_puma_options = pd.read_csv(paths.PUMA_TO_ZIP_DATA_PATH)[
```
rmudambi, line 416: We don't want to have to re-read from this file every time we sample. We should pass the state/puma combination options as arguments.

mattkappel (approved):

rmudambi (approved):

PR #212, opened 2023-02-14
Include all locations

stevebachmeier:
### Description
- *Category*: implementation
- *JIRA issue*: [MIC-3730](https://jira.ihme.washington.edu/browse/MIC-3730)
- *Research reference*: na

### Changes and notes
This adds all ACS states to the sim.

### Verification and Testing
pytests pass

### DIFF
--- a/integration_tests/test_domestic_migration.py
+++ b/integration_tests/test_domestic_migration.py
@@ -293,6 +293,13 @@ def test_addresses_during_moves(
 ):
     """Check that unit (household and business) address details change after a move."""
     address_cols = [address_id_col, state_id_col, puma_col]
+    us_locs = metadata.UNITED_STATES_LOCATIONS
+    if not us_locs:
+        # Use all ACS states
+        number_states_in_sim = len(pd.read_csv(paths.PUMA_TO_ZIP_DATA_PATH)["state"].unique())
+    else:
+        number_states_in_sim = len(us_locs)
+
     total_num_moved = 0
     for before, after in simulants_on_adjacent_timesteps:
         # get the unique unit (household or employer) dataset for before and after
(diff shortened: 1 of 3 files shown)


integration_tests/test_domestic_migration.py
```diff
@@ -323,13 +330,12 @@ def test_addresses_during_moves(
...
-        # TODO Check that *most* are in a new state when we add all 50+ locations
-        if len(metadata.UNITED_STATES_LOCATIONS) > 1 and mask_moved_units.sum() > 1:
-            assert any(
+        # Check that most movers are in a new state
+        if number_states_in_sim > 1 and mask_moved_units.sum() > 1:
+            assert (
                 before_units.loc[mask_moved_units, state_id_col]
                 != after_units.loc[mask_moved_units, state_id_col]
-            )
+            ).mean() > 0.5  # TODO: pick a smarter number?
```
mattkappel, line 338: What is the rationale for the 0.5?

stevebachmeier: Not very great rationale - it's literally that most are changed states. I tried being smarter by doing something like making sure the mean > 0.9 * 50/51 (51 being the number of states in the US) but that was failing the test in some cases, eg in one time step only 7 businesses moved and one of them moved within state but 6/7 didn't make the cut

albrja (approved):

integration_tests/test_domestic_migration.py
stevebachmeier: @mattkappel Going to merge for now - happy to readdress if you have any better ideas! 

(or @zmbc since you mentioned maybe trying to think of a statistically better way to deal with some of these fuzzy asserts)

PR #213, opened 2023-02-15
Address city map

albrja:
### Maps city to corresponding address id based on state.
- *Category*: Post-processing
- *JIRA issue*: [MIC-3742](https://jira.ihme.washington.edu/browse/MIC-3742)
- *Research reference*: 

### Changes and notes
-Adds city column to final results mapped from address_id and state
-Updates post-processing to work with change to household details pipeline that now includes state_id.

### Verification and Testing
Successfully ran post-processing and expected columns were in final outputs.

### DIFF
--- a/src/vivarium_census_prl_synth_pop/results_processing/addresses.py
+++ b/src/vivarium_census_prl_synth_pop/results_processing/addresses.py
@@ -157,7 +157,7 @@ def get_city_map(
     randomness: RandomnessStream,
 ) -> Dict[str, pd.Series]:
 
-    # Load addresses data fro artifact
+    # Load addresses data from artifact
     addresses = artifact.load(data_keys.SYNTHETIC_DATA.ADDRESSES).reset_index()
     # Get observer data to map
     output_cols = [column_name, "state"]
--- a/src/vivarium_census_prl_synth_pop/results_processing/addresses.py
+++ b/src/vivarium_census_prl_synth_pop/results_processing/addresses.py
@@ -56,6 +56,7 @@ def get_address_id_maps(
     maps.update(
         get_household_address_map(column_name, formatted_obs_data, artifact, randomness)
     )
+    maps.update(get_city_map(column_name, formatted_obs_data, artifact, randomness))
     return maps
 
 
@@ -145,3 +146,29 @@ def get_household_address_map(
         address_map[obs_column] = address_data[obs_column]
 
     return address_map
+
+
+def get_city_map(
+    column_name: str,
+    obs_data: pd.DataFrame,
+    artifact: Artifact,
+    randomness: RandomnessStream,
+) -> Dict[str, pd.Series]:
+
+    # Load addresses data fro artifact
+    addresses = artifact.load(data_keys.SYNTHETIC_DATA.ADDRESSES).reset_index()
+    # Get observer data to map
+    output_cols = [column_name, "state"]
+    city_data = obs_data.reset_index()[output_cols].drop_duplicates().set_index("address_id")
+
+    for state in city_data["state"].str.lower().unique():
+        cities = vectorized_choice(
+            options=addresses.loc[addresses["Province"] == state, "Municipality"],
+            n_to_choose=len(city_data.loc[city_data["state"] == state.upper()]),
+            randomness_stream=randomness,
+            additional_key="city",
+        ).to_numpy()
+        city_data.loc[city_data["state"] == state.upper(), "city"] = cities
+
+    city_map = {"city": city_data["city"]}
+    return city_map
--- a/src/vivarium_census_prl_synth_pop/results_processing/addresses.py
+++ b/src/vivarium_census_prl_synth_pop/results_processing/addresses.py
@@ -46,7 +46,7 @@ def get_address_id_maps(
     if column_name != "address_id":
         raise ValueError(f"Expected `address_id`, got `{column_name}`")
     maps = dict()
-    output_cols_superset = [column_name, "state", "puma"]
+    output_cols_superset = [column_name, "state_id", "state", "puma"]
     formatted_obs_data = format_data_for_mapping(
         index_name=column_name,
         obs_results=obs_data,
@@ -82,7 +82,7 @@ def get_zipcode_map(
 
     """
     zip_map_dict = {}
-    output_cols = [column_name, "state", "puma"]  # columns in the output we use to map
+    output_cols = [column_name, "state_id", "puma"]  # columns in the output we use to map
     simulation_addresses = (
         obs_data.reset_index()[output_cols].drop_duplicates().set_index("address_id")
     )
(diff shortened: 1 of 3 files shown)


src/vivarium_census_prl_synth_pop/results_processing/addresses.py
```diff
@@ -46,7 +46,7 @@ def get_address_id_maps(
     if column_name != "address_id":
         raise ValueError(f"Expected `address_id`, got `{column_name}`")
     maps = dict()
-    output_cols_superset = [column_name, "state", "puma"]
+    output_cols_superset = [column_name, "state_id", "state", "puma"]
```
stevebachmeier, line 49: Do you need state at all?

src/vivarium_census_prl_synth_pop/results_processing/addresses.py
```diff
@@ -145,3 +148,29 @@ def get_household_address_map(
...
+
+
+def get_city_map(
+    column_name: str,
+    obs_data: pd.DataFrame,
+    artifact: Artifact,
+    randomness: RandomnessStream,
+) -> Dict[str, pd.Series]:
+
+    # Load addresses data fro artifact
```
stevebachmeier, line 160: nit: froM artifact

src/vivarium_census_prl_synth_pop/results_processing/addresses.py
albrja: We get the state abbreviation when we read in the raw results (which is the "state" column) and this is used for selecting cities.

stevebachmeier: But raw_results is now state_id, no? You're talking raw_results as in observer data, right?

albrja: Yes but we do some cleaning when we initially read the data into memory for results processing.  State_id is in raw_results and when we read it in we create state which is the state abbreviation we want in final outputs.  I could do that same mapping in here to go from the state_id to the abbreviation but it has already been done so it doesn't hurt to include the extra column.

albrja: The cleaning I am referring to is taking place in load_data in make_results.py and the actual code is in formatter.py

stevebachmeier: ah got it - that roundabout confused me

src/vivarium_census_prl_synth_pop/results_processing/addresses.py
```diff
@@ -145,3 +148,29 @@ def get_household_address_map(
...
+
+    # Load addresses data from artifact
+    addresses = artifact.load(data_keys.SYNTHETIC_DATA.ADDRESSES).reset_index()
+    # Get observer data to map
+    output_cols = [column_name, "state"]
+    city_data = obs_data.reset_index()[output_cols].drop_duplicates().set_index("address_id")
+
+    for state in city_data["state"].str.lower().unique():
+        cities = vectorized_choice(
+            options=addresses.loc[addresses["Province"] == state, "Municipality"],
```
stevebachmeier, line 168: Does each city in a given state have an equal change of being chosen?

tests/test_synthetic_pii.py
```diff
@@ -361,3 +362,42 @@ def test_zipcode_mapping():
         expected_proportion_90706,
         rtol=0.1,
     )
+
+
+def test_city_address(mocker):
+    # This tests that we assign cities based on the correct state
```
stevebachmeier, line 368: nit: typically our docstrings are block-commented w/ """docstring""".

tests/test_synthetic_pii.py
```diff
@@ -361,3 +362,42 @@ def test_zipcode_mapping():
         expected_proportion_90706,
         rtol=0.1,
     )
+
+
+def test_city_address(mocker):
```
stevebachmeier, line 367: Great seeing a unit test!

I do wonder if we need to start building out integration tests for post-processing. It's a long time to wait for tests to pass, though - maybe at the point of post-processing the testing is just done by v&v? @rmudambi?

src/vivarium_census_prl_synth_pop/results_processing/formatter.py
```diff
@@ -31,24 +31,18 @@ def format_address_id(data: pd.DataFrame) -> pd.Series:
...
-    state_name_map = data["state"].map(state_id_map)
+    state_name_map = data["state_id"].map(state_id_map)
     return state_name_map.map(metadata.US_STATE_ABBRV_MAP)
 
 
-def get_state_id(data: pd.DataFrame) -> pd.Series:
-    return data["state"]
-
-
 # Fixme: Add formatting functions as necessary
```
stevebachmeier, line 38: Remove fixme?

src/vivarium_census_prl_synth_pop/results_processing/addresses.py
stevebachmeier, line 168: Does this mean each city has an equal chance of getting chosen?

stevebachmeier: I see - addresses is actually ACS data and so larger cities will have more rows and thus more likely to be sampled

stevebachmeier (approved):

mattkappel:
Isn't there a research reference for this?

src/vivarium_census_prl_synth_pop/results_processing/addresses.py
zmbc: Not to pick nits, but it's not ACS data -- (publicly released) ACS data does not include addresses. This is standalone address data, which comes from `libpostal` and ultimately originates from crowdsourced OpenStreetMap addresses and government records.

src/vivarium_census_prl_synth_pop/results_processing/addresses.py
```diff
@@ -81,7 +82,7 @@ def get_zipcode_map(
 
     """
     zip_map_dict = {}
-    output_cols = [column_name, "state", "puma"]  # columns in the output we use to map
+    output_cols = [column_name, "state_id", "puma"]  # columns in the output we use to map
```
mattkappel, line 85: Thanks for finding this disconnect and correcting @albrja !!

src/vivarium_census_prl_synth_pop/results_processing/addresses.py
mattkappel: That looks like it. The documentation link here doesn't seem to imply they need to match puma https://vivarium-research.readthedocs.io/en/latest/models/concept_models/vivarium_census_synthdata/concept_model.html#id34

@albrja can you please confirm?

zmbc: But your larger point is correct; even though we have no real guarantees given the data sources that what we have is a true distribution of cities within a state, we consider it close enough (certainly better than a uniform random sample of city) so we allow the data to dictate which cities are more common.

src/vivarium_census_prl_synth_pop/results_processing/formatter.py
mattkappel: Nah, we're gonna keep needing it. Maybe `Fixme` is more `NB` or `Nota bene`

mattkappel (approved):

src/vivarium_census_prl_synth_pop/results_processing/addresses.py
zmbc: See [my reply below](https://github.com/ihmeuw/vivarium_census_prl_synth_pop/pull/213#discussion_r1107634138).

zmbc: The documentation for this is here: https://vivarium-research.readthedocs.io/en/latest/models/concept_models/vivarium_census_synthdata/concept_model.html#id34. It could probably be a little more explicit on the reasoning for this behavior.

PR #216, opened 2023-02-17
Add persistent ITIN for stimulants without SSN

mattkappel:
### Description
- *Category*:  implementation
- *JIRA issue*: [MIC-3573](https://jira.ihme.washington.edu/browse/MIC-3573)
- *Research reference*: https://vivarium-research.readthedocs.io/en/latest/models/concept_models/vivarium_census_synthdata/concept_model.html#itin-generation

### Changes and notes
- Adds a generate ITIN function and related tests
- Problem of collisions is deferred for [MIC-3837](https://jira.ihme.washington.edu/browse/MIC-3837)

### Verification and Testing
Unit tests pass

### DIFF
--- a/src/vivarium_census_prl_synth_pop/results_processing/ssn_and_itin.py
+++ b/src/vivarium_census_prl_synth_pop/results_processing/ssn_and_itin.py
@@ -48,7 +48,7 @@ def generate_itin(
     pre_shift_groups = itin["group"].copy()
     itin.loc[pre_shift_groups <= 16, "group"] += 49
     itin.loc[((pre_shift_groups <= 35) & (pre_shift_groups >= 17)), "group"] += 53
-    itin.loc[((pre_shift_groups <= 38) & (pre_shift_groups >= 36)), "group"] += 52
+    itin.loc[((pre_shift_groups <= 38) & (pre_shift_groups >= 36)), "group"] += 54
     itin.loc[pre_shift_groups >= 39, "group"] += 55
 
     # Serial numbers 1-9999 inclusive
--- a/src/vivarium_census_prl_synth_pop/results_processing/ssn_and_itin.py
+++ b/src/vivarium_census_prl_synth_pop/results_processing/ssn_and_itin.py
@@ -8,9 +8,9 @@ from vivarium_census_prl_synth_pop.utilities import random_integers
 
 
 def generate_itin(
-        size: int,
-        additional_key: Any,
-        randomness: RandomnessStream,
+    size: int,
+    additional_key: Any,
+    randomness: RandomnessStream,
 ) -> np.ndarray:
     """
     Generate a np.ndarray of length `size` of hopefully unique ITIN values.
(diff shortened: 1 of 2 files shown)


albrja (approved):

src/vivarium_census_prl_synth_pop/results_processing/ssn_and_itin.py
```diff
@@ -0,0 +1,69 @@
...
+
+    # Serial numbers 1-9999 inclusive
+    itin["serial"] = random_integers(
+        min_val=1,
+        max_val=9999,
+        index=itin.index,
+        randomness=randomness,
+        additional_key=f"{additional_key}_itin_serial",
+    )
+    itin["itin"] = ""
```
zmbc, line 62: Can we not just define "itin" in one expression instead of these intermediate steps?

src/vivarium_census_prl_synth_pop/results_processing/ssn_and_itin.py
```diff
@@ -0,0 +1,69 @@
...
+    itin["itin"] += (
+        itin["area"].astype(str) + "-"
+    )  # no zfill needed, values should be 900-999
+    itin["itin"] += itin["group"].astype(str) + "-"  # no zfill needed likewise
+    itin["itin"] += itin["serial"].astype(str).str.zfill(4)
+
+    return itin["itin"].to_numpy()
```
zmbc, line 69: I've seen this a few places; what is the purpose of `.to_numpy()`?

src/vivarium_census_prl_synth_pop/results_processing/ssn_and_itin.py
```diff
@@ -0,0 +1,69 @@
...
+        max_val=999,
+        index=itin.index,
+        randomness=randomness,
+        additional_key=f"{additional_key}_itin_area",
+    )
+
+    # Group numbers have range 50-65, 70-88, 90-92, or 94-99, inclusive
+    itin["group"] = random_integers(
+        min_val=1,
+        max_val=44,  # (65-50+1)+(88-70+1)+(92-90+1)+(99-94+1)
```
zmbc, line 41: I think you can do this much more cleanly by creating a Series mapping to the acceptable values from consecutive integers; that would remove the need for these manual calculations and the ones below.

src/vivarium_census_prl_synth_pop/results_processing/ssn_and_itin.py
mattkappel: To get it to be an array for easy assignment via `.loc`, in this case at least.

src/vivarium_census_prl_synth_pop/results_processing/ssn_and_itin.py
mattkappel: Yeah, agreed, that would be a cleaner and probably more efficient. I'll keep that in mind, I'll be in this code for SSN as well.

src/vivarium_census_prl_synth_pop/results_processing/ssn_and_itin.py
mattkappel: Probably. I tried do to this as an f-string lambda in an apply and ran into trouble. This was the approach the previous SSN implementation did, and works for now.

PR #218, opened 2023-02-23
implement CRNs for business move dynamics across parallel runs

stevebachmeier:
### Description
- *Category*: Implementation
- *JIRA issue*: [MIC-3836](https://jira.ihme.washington.edu/browse/MIC-3836)
- *Research reference*: na

### Changes and notes
Business are different than households in that new businesses are never generated
during the sim. Currently, parallel simulations have no idea what businesses are
moving in the other sims. This utilizes a new RandomnessStream that does 
not use simulation random_seed for such dynamics

Other notes:

- Added a us_population_size key to the config file (mostly for debugging purposes
    probably) - it overrides the real US population in data_values
- Fundamentally changed the way the "bank" of employers is built. Instead of scaling
    with the number of 18+ simulants (which lead to different sized banks) we now
    rely on the actual up_population_size and a since percentage scalar.
- Changed randomness.choice to vectorized_choice when assigning employer_ids
    b/c it was way too memory-intensive when building up the initial sim using the much
    larger bank of employers

### Verification and Testing
- pytests pass
- ensured that tax w2 observer reports a single address for each employer now
- loads of other qc

### DIFF
--- a/src/vivarium_census_prl_synth_pop/components/businesses.py
+++ b/src/vivarium_census_prl_synth_pop/components/businesses.py
@@ -139,9 +139,7 @@ class Businesses:
 
         pop["employer_id"] = data_values.Unemployed.EMPLOYER_ID
         working_age = pop.loc[pop["age"] >= data_values.WORKING_AGE].index
-        pop.loc[working_age, "employer_id"] = self.assign_random_employer(
-            working_age, randomness_stream=self.business_moves_randomness
-        )
+        pop.loc[working_age, "employer_id"] = self.assign_random_employer(working_age)
 
         # Give military gq sims military employment
         military_index = pop.index[
@@ -186,7 +184,7 @@ class Businesses:
             "moving_businesses",
         )
 
-        self.update_business_addresses(businesses_that_move_idx, event)
+        self.update_business_addresses(businesses_that_move_idx)
 
         # change jobs by rate as well as if the household moves (only includes
         # working-age simulants and excludes simulants living in military GQ)
@@ -285,20 +283,20 @@ class Businesses:
         pct_adults_needing_employers = 1 - known_employers["prevalence"].sum()
         n_need_employers = np.round(
             self.us_population_size
-            * data_values.PROPORTION_18_PLUS
+            * data_values.PROPORTION_WORKING_AGE
             * pct_adults_needing_employers
         )
-        n_businesses = int(n_need_employers // data_values.EXPECTED_EMPLOYEES_PER_BUSINESS)
-        random_employers_idx = np.arange(2, n_businesses + 2)
+        n_businesses = int(n_need_employers / data_values.EXPECTED_EMPLOYEES_PER_BUSINESS)
+        random_employers_ids = np.arange(2, n_businesses + 2)
         employee_count_propensity = self.business_moves_randomness.get_draw(
-            random_employers_idx
+            random_employers_ids
         )
         employee_counts = stats.lognorm.ppf(
             q=employee_count_propensity, s=1, scale=np.exp(4)
         ).round()
         random_employers = pd.DataFrame(
             {
-                "employer_id": random_employers_idx,
+                "employer_id": random_employers_ids,
                 "employer_name": ["not implemented"] * n_businesses,
                 "prevalence": employee_counts / employee_counts.sum(),
                 "employer_address_id": np.arange(
@@ -308,7 +306,7 @@ class Businesses:
         )
 
         # Combine and add details
-        businesses = pd.concat([known_employers, random_employers], ignore_index=True)
+        businesses = pd.concat([known_employers, random_employers]).set_index("employer_id")
         states_pumas = randomly_sample_states_pumas(
             unit_ids=businesses.index,
             state_puma_options=self.state_puma_options,
@@ -321,20 +319,17 @@ class Businesses:
         # Update employer_address_id_count
         self.employer_address_id_count = len(businesses)
 
-        return businesses.set_index("employer_id")
+        return businesses
 
     def assign_random_employer(
         self,
         sim_index: pd.Index,
         additional_seed: Any = None,
-        randomness_stream: RandomnessStream = None,
     ) -> pd.Series:
-        if randomness_stream == None:
-            randomness_stream = self.randomness
         return vectorized_choice(
             options=self.businesses.index,
             n_to_choose=len(sim_index),
-            randomness_stream=randomness_stream,
+            randomness_stream=self.randomness,
             weights=self.businesses["prevalence"],
             additional_key=additional_seed,
         )
@@ -396,6 +391,9 @@ class Businesses:
             ids (Union[pd.Index, pd.Series]): The state table index (ie simulant
                 ids) or series of employer_ids to add the business details
                 to.
+                NOTE: if a pd.Series is passed in, it must be named
+                "employer_id" and the expectation is that it has an index
+                consisting of simulant IDs.
 
         Returns:
             pd.DataFrame: employer details per person or for a series of
@@ -414,7 +412,7 @@ class Businesses:
 
         return business_details
 
-    def update_business_addresses(self, moving_business_ids: pd.Index, event) -> None:
+    def update_business_addresses(self, moving_business_ids: pd.Index) -> None:
         """
         Change the address information associated with each of the provided
         business_ids to a new, unique address.
--- a/src/vivarium_census_prl_synth_pop/components/businesses.py
+++ b/src/vivarium_census_prl_synth_pop/components/businesses.py
@@ -1,4 +1,4 @@
-from typing import Any
+from typing import Any, Union
 
 import numpy as np
 import pandas as pd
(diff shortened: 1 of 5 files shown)


src/vivarium_census_prl_synth_pop/components/businesses.py
```diff
@@ -123,15 +131,17 @@ def on_initialize_simulants(self, pop_data: SimulantData) -> None:
...
 
         pop = self.population_view.subview(["age", "household_id"]).get(
             pop_data.index, query="tracked"
         )
 
         pop["employer_id"] = data_values.Unemployed.EMPLOYER_ID
         working_age = pop.loc[pop["age"] >= data_values.WORKING_AGE].index
-        pop.loc[working_age, "employer_id"] = self.assign_random_employer(working_age)
+        pop.loc[working_age, "employer_id"] = self.assign_random_employer(
+            working_age, randomness_stream=self.business_moves_randomness
```
stevebachmeier, line 143: Now that I've PRed this I'm not convinced we want to use the CRN here.

zmbc: I don't think we want to use it here.

src/vivarium_census_prl_synth_pop/components/businesses.py
```diff
@@ -272,46 +281,61 @@ def generate_businesses(self, pop_data: SimulantData) -> pd.DataFrame():
             }
         )
 
+        # Generate random employers
         pct_adults_needing_employers = 1 - known_employers["prevalence"].sum()
-        n_need_employers = np.round(n_working_age * pct_adults_needing_employers)
-        employee_counts = np.random.lognormal(
```
zmbc, line 277: Yikes, was this not previously using CRN at all?

zmbc (approved):

src/vivarium_census_prl_synth_pop/components/businesses.py
stevebachmeier: It wasn't!

src/vivarium_census_prl_synth_pop/components/businesses.py
rmudambi: We definitely don't want to use this here. We should use the standard randomness stream. The business moves randomness stream should be used only for initializing and modifying the businesses data structure. 

Honestly, assigning employers should really happen in a different component altogether, as it's really an entirely different task, but that refactor is definitely out of scope here.

src/vivarium_census_prl_synth_pop/constants/data_values.py
```diff
@@ -11,6 +11,7 @@
 
 # TODO: implement gbd call (vivarium_inputs.get_population_structure("United States"))
 US_POPULATION = 333339776
+PROPORTION_18_PLUS = 0.7756
```
rmudambi, line 14: Better name would be `PROPORTION_WORKING_AGE`

src/vivarium_census_prl_synth_pop/components/businesses.py
```diff
@@ -272,46 +281,61 @@ def generate_businesses(self, pop_data: SimulantData) -> pd.DataFrame():
...
-            4, 1, size=int(n_need_employers // data_values.EXPECTED_EMPLOYEES_PER_BUSINESS)
+        n_need_employers = np.round(
+            self.us_population_size
+            * data_values.PROPORTION_18_PLUS
+            * pct_adults_needing_employers
+        )
+        n_businesses = int(n_need_employers // data_values.EXPECTED_EMPLOYEES_PER_BUSINESS)
```
rmudambi, line 291: You don't need to do int division ( i.e. `//`) if you are casting to `int`.

src/vivarium_census_prl_synth_pop/components/businesses.py
```diff
@@ -55,8 +57,14 @@ def name(self):
     #################
 
     def setup(self, builder: Builder):
+        self.us_population_size = builder.configuration.us_population_size
         self.start_time = get_time_stamp(builder.configuration.time.start)
         self.randomness = builder.randomness.get_stream(self.name)
+        # create a randomness stream for business moves; use a common random number
+        # for identical business move behavior across parallel sims
+        self.business_moves_randomness = RandomnessStream(
+            key="business_moves", clock=builder.time.clock(), seed=12345
```
aflaxman, line 66: you could use the draw here to allow different draws to have different business dynamics and still have them be common across seeds (did i get the terminology right?)

src/vivarium_census_prl_synth_pop/components/businesses.py
```diff
@@ -272,46 +281,61 @@ def generate_businesses(self, pop_data: SimulantData) -> pd.DataFrame():
...
-        n_need_employers = np.round(n_working_age * pct_adults_needing_employers)
-        employee_counts = np.random.lognormal(
-            4, 1, size=int(n_need_employers // data_values.EXPECTED_EMPLOYEES_PER_BUSINESS)
+        n_need_employers = np.round(
+            self.us_population_size
+            * data_values.PROPORTION_18_PLUS
+            * pct_adults_needing_employers
+        )
+        n_businesses = int(n_need_employers // data_values.EXPECTED_EMPLOYEES_PER_BUSINESS)
+        random_employers_idx = np.arange(2, n_businesses + 2)
```
rmudambi, line 292: This isn't an index, so maybe call it ids rather than idx

src/vivarium_census_prl_synth_pop/components/businesses.py
```diff
@@ -272,46 +281,61 @@ def generate_businesses(self, pop_data: SimulantData) -> pd.DataFrame():
...
+        self.employer_address_id_count = len(businesses)
 
         return businesses.set_index("employer_id")
 
     def assign_random_employer(
-        self, sim_index: pd.Index, additional_seed: Any = None
+        self,
+        sim_index: pd.Index,
+        additional_seed: Any = None,
+        randomness_stream: RandomnessStream = None,
```
rmudambi, line 330: We should always be using `self.randomness_stream` here, so you should remove the `randomness_stream` argument.

src/vivarium_census_prl_synth_pop/components/businesses.py
```diff
@@ -365,10 +389,23 @@ def calculate_income(self, idx: pd.Index) -> pd.Series:
...
 
-    def get_business_details(self, idx: pd.Index) -> pd.DataFrame:
-        """Source of the business details pipeline"""
-        pop = self.population_view.get(idx)[["employer_id"]]
-        business_details = pop.join(
+    def get_business_details(self, ids: Union[pd.Index, pd.Series]) -> pd.DataFrame:
+        """Source of the business details pipeline.
+
+        Args:
+            ids (Union[pd.Index, pd.Series]): The state table index (ie simulant
```
rmudambi, line 396: This will only work if the Series passed in has the name "employer_id", since you don't name the column of the resultant DataFrame.

I suggest adding a note here that the Series is expected to have an index of simulant ids (though obviously this call will still work even if that is not the case).

src/vivarium_census_prl_synth_pop/components/businesses.py
```diff
@@ -377,7 +414,7 @@ def get_business_details(self, idx: pd.Index) -> pd.DataFrame:
 
         return business_details
 
-    def update_business_addresses(self, moving_business_ids: pd.Index) -> None:
+    def update_business_addresses(self, moving_business_ids: pd.Index, event) -> None:
```
rmudambi, line 417: Why are you passing the event here? You don't use it.

src/vivarium_census_prl_synth_pop/components/businesses.py
```diff
@@ -272,46 +281,61 @@ def generate_businesses(self, pop_data: SimulantData) -> pd.DataFrame():
...
-                * pct_adults_needing_employers,
+                "prevalence": employee_counts / employee_counts.sum(),
                 "employer_address_id": np.arange(
                     len(known_employers), n_businesses + len(known_employers)
                 ),
             }
         )
 
+        # Combine and add details
         businesses = pd.concat([known_employers, random_employers], ignore_index=True)
```
rmudambi, line 311: We should set the employer_id column as the index right here. Then there's no need to ignore the index. That also means when we pass businesses.index to randomly_sample_states_pumas() below we're actually sending in the true index. As it turns out both indexes are identical (i.e. they are both equivalent to np.arange(n_businesses + 2)), but it's good to be precise here.

rmudambi (changes requested):

src/vivarium_census_prl_synth_pop/components/businesses.py
rmudambi: I'm not exactly sure what you mean here. We are only running the simulation with a single draw. If we did have different draws they would result in different business dynamics if we had draw level parameterization of those dynamics (say if move rates were specified at the draw level).

stevebachmeier: This brings up a good point - we've kind of set a requirement that we run with one draw; should we actually enforce that and throw a runtime error if otherwise?

rmudambi (approved):

PR #224, opened 2023-02-25
Support arbitrarily small populations

zmbc:
### Description
- *Category*: misc
- *JIRA issue*: none
- *Research reference*: none

### Changes and notes

In working on "fuzzy testing" I discovered a few places where the simulation itself would error with too small of a population.

### Verification and Testing

Integration tests (at normal population size). Ran simulation to completion with population size of 1 (with two different random seeds). There may still be some rare events that I have missed.

### DIFF
--- a/src/vivarium_census_prl_synth_pop/components/businesses.py
+++ b/src/vivarium_census_prl_synth_pop/components/businesses.py
@@ -338,7 +338,7 @@ class Businesses:
     def assign_different_employer(self, changing_jobs: pd.Index) -> pd.Series:
         current_employers = self.population_view.subview(["employer_id"]).get(
             changing_jobs, query="tracked"
-        )["employer_id"]
+        ).squeeze(axis=1)
 
         new_employers = current_employers.copy()
 
--- a/src/vivarium_census_prl_synth_pop/components/businesses.py
+++ b/src/vivarium_census_prl_synth_pop/components/businesses.py
@@ -336,9 +336,11 @@ class Businesses:
         )
 
     def assign_different_employer(self, changing_jobs: pd.Index) -> pd.Series:
-        current_employers = self.population_view.subview(["employer_id"]).get(
-            changing_jobs, query="tracked"
-        ).squeeze(axis=1)
+        current_employers = (
+            self.population_view.subview(["employer_id"])
+            .get(changing_jobs, query="tracked")
+            .squeeze(axis=1)
+        )
 
         new_employers = current_employers.copy()
 
--- a/src/vivarium_census_prl_synth_pop/components/businesses.py
+++ b/src/vivarium_census_prl_synth_pop/components/businesses.py
@@ -2,6 +2,7 @@ from typing import Any, Union
 
 import numpy as np
 import pandas as pd
+from loguru import logger
 from scipy import stats
 from vivarium.framework.engine import Builder
 from vivarium.framework.event import Event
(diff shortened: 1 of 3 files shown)


src/vivarium_census_prl_synth_pop/components/population.py
```diff
@@ -937,7 +943,7 @@ def assign_college_simulants_guardians(
             .reset_index()
             .rename(columns={"index": "partner_id"})
             .groupby("household_id")["partner_id"]
-            .apply(np.random.choice)
+            .first()
```
zmbc, line 946: Note: There should never be more than one item to choose from here. See #225 for a new integration test of this household structure invariant.

The `np.random.choice` version was failing for me with Pandas 1.4.0 due to this bug: pandas-dev/pandas#47985

src/vivarium_census_prl_synth_pop/components/businesses.py
```diff
@@ -335,24 +336,28 @@ def assign_random_employer(
         )
 
     def assign_different_employer(self, changing_jobs: pd.Index) -> pd.Series:
-        # TODO: This limits employers to those existing in the pop table. If
```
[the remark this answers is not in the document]
mattkappel: 😲

src/vivarium_census_prl_synth_pop/components/person_migration.py
```diff
@@ -226,16 +226,19 @@ def _perform_non_reference_person_moves(
...
-                break
-
-            household_id_values[to_move] = self.randomness.choice(
-                to_move,
-                choices=households_with_non_movers,
-                additional_key=f"non_reference_person_move_household_ids_{iteration}",
-            )
-            to_move = movers[household_id_values == pop.loc[movers, "household_id"]]
+        # There is an edge case in very small populations where there are no households
+        # to move to; in that case, the move is not performed
```
stevebachmeier, line 230: Does this require a doc update or is it edgy enough?

src/vivarium_census_prl_synth_pop/components/population.py
stevebachmeier: Why was random.choice ever the thing if there was always only one option? Or did that used to not be the case?

src/vivarium_census_prl_synth_pop/components/person_migration.py
zmbc: Good question! I can make a doc update for this. It is _super_ edgy so I think this can be an "implement first, document later" situation.

src/vivarium_census_prl_synth_pop/components/population.py
zmbc: My best guess is that it wasn't known by the person implementing this whether ACS source data would always respect this constraint -- it does.

src/vivarium_census_prl_synth_pop/components/businesses.py
```diff
@@ -335,24 +336,28 @@ def assign_random_employer(
...
-        # we want all employers that were initialized to be included, we can
-        # instead use the business data structure.
-        current_employers = (
-            self.population_view.subview(["employer_id"])
-            .get(changing_jobs, query="tracked")
-            .squeeze()
-        )
+        current_employers = self.population_view.subview(["employer_id"]).get(
+            changing_jobs, query="tracked"
+        )["employer_id"]
```
rmudambi, line 341: Can do `squeeze(axis=1)` which I prefer.

mattkappel (approved):

src/vivarium_census_prl_synth_pop/components/population.py
```diff
@@ -369,19 +369,25 @@ def initialize_non_reference_person_immigrants(self, pop_data: SimulantData) ->
...
+        else:
+            # TODO: For now, we do a simple random sample of households.
+            # As part of future PUMA perturbation work, we plan to actually use
+            # the PUMA value in the ACS row in choosing the household they should join.
+            new_simulants["household_id"] = vectorized_choice(
+                existing_household_ids,
+                n_to_choose=len(new_simulants),
+                randomness_stream=self.randomness,
+                additional_key="household_id",
+            )
```
rmudambi, line 390: `astype` is no longer needed?

rmudambi (approved):

src/vivarium_census_prl_synth_pop/components/businesses.py
zmbc: TIL! That is nicer.

src/vivarium_census_prl_synth_pop/components/population.py
zmbc: We do `.astype(int)` on the household_id column just below. This `astype` was actually doing nothing, because the type of `existing_simulants["household_id"]` was float anyway.

PR #226, opened 2023-02-27
Employer addresses

albrja:
### Adds employer address columns to final results.
- *Category*: Post-processing
- *JIRA issue*: [MIC-3737](https://jira.ihme.washington.edu/browse/MIC-3737)
- *Research reference*: https://vivarium-research.readthedocs.io/en/latest/models/concept_models/vivarium_census_synthdata/concept_model.html#w2-and-1099-forms

### Changes and notes
-Adds logic for `get_address_id_maps `to work for `employer_address_ids`
-Adds necessary additional key parameter to not reuse the same key.
-Parameterizes necessary pytests.

### Verification and Testing
-Successfully ran parallel simulation and `make_results `and saw desired columns in final results.

### DIFF
--- a/src/vivarium_census_prl_synth_pop/results_processing/addresses.py
+++ b/src/vivarium_census_prl_synth_pop/results_processing/addresses.py
@@ -59,10 +59,7 @@ def get_address_id_maps(
         )
 
     maps = dict()
-<<<<<<< HEAD
     output_cols_superset = [column_name, "state_id", "state", "puma", "po_box"]
-=======
->>>>>>> caaa200 (Implementation refactor for employer address - format state column and refactor to address.py)
     formatted_obs_data = format_data_for_mapping(
         index_name=column_name,
         obs_results=obs_data,
@@ -136,11 +133,7 @@ def get_zipcode_map(
             n_to_choose=len(df_locale),
             randomness_stream=randomness,
             weights=locale_group["proportion"],
-<<<<<<< HEAD
             additional_key=f"{additional_key}zip_map_{state_id}_{puma}",
-=======
-            additional_key=f"{column_name}_zip_map_{state_id}_{puma}",
->>>>>>> caaa200 (Implementation refactor for employer address - format state column and refactor to address.py)
         ).to_numpy()
 
     # Map against obs_data
@@ -207,11 +200,7 @@ def get_city_map(
             options=addresses.loc[addresses["Province"] == state, "Municipality"],
             n_to_choose=len(city_data.loc[city_data["state"] == state.upper()]),
             randomness_stream=randomness,
-<<<<<<< HEAD
             additional_key=f"{additional_key}city_map",
-=======
-            additional_key=f"{column_name}_city_{state}",
->>>>>>> caaa200 (Implementation refactor for employer address - format state column and refactor to address.py)
         ).to_numpy()
         city_data.loc[city_data["state"] == state.upper(), "city"] = cities
 
--- a/src/vivarium_census_prl_synth_pop/results_processing/addresses.py
+++ b/src/vivarium_census_prl_synth_pop/results_processing/addresses.py
@@ -45,7 +45,7 @@ def get_address_id_maps(
     """
 
     if column_name == "address_id":
-        output_cols_superset = [column_name, "state_id", "state", "puma"]
+        output_cols_superset = [column_name, "state_id", "state", "puma", "po_box"]
     elif column_name == "employer_address_id":
         output_cols_superset = [
             column_name,
--- a/src/vivarium_census_prl_synth_pop/results_processing/addresses.py
+++ b/src/vivarium_census_prl_synth_pop/results_processing/addresses.py
@@ -43,17 +43,17 @@ def get_address_id_maps(
     A dictionary of pd.Series suitable for pd.Series.map, indexed by `address_id`
 
     """
-
-    if column_name == "address_id":
-        output_cols_superset = [column_name, "state_id", "state", "puma", "po_box"]
-    elif column_name == "employer_address_id":
-        output_cols_superset = [
-            column_name,
-            "employer_state_id",
-            "employer_state",
-            "employer_puma",
-        ]
-    else:
+    try:
+        output_cols_superset = {
+            "address_id": [column_name, "state_id", "state", "puma", "po_box"],
+            "employer_address_id": [
+                column_name,
+                "employer_state_id",
+                "employer_state",
+                "employer_puma",
+            ],
+        }[column_name]
+    except KeyError:
         raise ValueError(
             f"Expected `address_id` or 'employer_address_id', got `{column_name}`"
         )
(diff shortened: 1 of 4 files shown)


src/vivarium_census_prl_synth_pop/results_processing/addresses.py
```diff
@@ -108,34 +126,36 @@ def get_zipcode_map(
         .groupby(["state", "puma"])
     )
 
-    for (state_id, puma), df_locale in simulation_addresses.groupby(["state_id", "puma"]):
+    for (state_id, puma), df_locale in simulation_addresses.groupby(groupby_cols):
```
rmudambi, line 129: The groupby columns are always the output columns minus the column name. It'd be cleaner to define them here like that rather than defining a variable up top.

src/vivarium_census_prl_synth_pop/results_processing/addresses.py
```diff
@@ -166,19 +189,26 @@ def get_city_map(
...
-    city_data = obs_data.reset_index()[output_cols].drop_duplicates().set_index("address_id")
+    if column_name == "address_id":
+        state_col = "state"
+    else:
+        state_col = "employer_state"
+    output_cols = [column_name, state_col]
+    city_data = obs_data.reset_index()[output_cols].drop_duplicates().set_index(column_name)
 
-    for state in city_data["state"].str.lower().unique():
+    for state in city_data[state_col].str.lower().unique():
```
rmudambi, line 199: If you do the `unique` before the `.str.lower()` you'll be doing the latter operation on far fewer rows.

src/vivarium_census_prl_synth_pop/results_processing/addresses.py
```diff
@@ -88,9 +101,14 @@ def get_zipcode_map(
...
-    output_cols = [column_name, "state_id", "puma"]  # columns in the output we use to map
+    if column_name == "address_id":
+        output_cols = [column_name, "state_id", "puma"]  # columns in the output we use to map
+        groupby_cols = ["state_id", "puma"]
+    else:
+        output_cols = [column_name, "employer_state_id", "employer_puma"]
+        groupby_cols = ["employer_state_id", "employer_puma"]
     simulation_addresses = (
-        obs_data.reset_index()[output_cols].drop_duplicates().set_index("address_id")
+        obs_data.reset_index()[output_cols].drop_duplicates().set_index(column_name)
```
rmudambi, line 111: Didn't we already drop all the duplicates when we formatted `obs_data`? Do we actually still have duplicates at this point? The same question applies in all the other places we do this.

src/vivarium_census_prl_synth_pop/results_processing/addresses.py
```diff
@@ -166,19 +189,26 @@ def get_city_map(
...
         cities = vectorized_choice(
             options=addresses.loc[addresses["Province"] == state, "Municipality"],
-            n_to_choose=len(city_data.loc[city_data["state"] == state.upper()]),
+            n_to_choose=len(city_data.loc[city_data[state_col] == state.upper()]),
             randomness_stream=randomness,
-            additional_key=f"{additional_key}city_map",
+            additional_key=f"{additional_key}{column_name}_city_map",
         ).to_numpy()
-        city_data.loc[city_data["state"] == state.upper(), "city"] = cities
+        city_data.loc[city_data[state_col] == state.upper(), "city"] = cities
```
rmudambi, line 206: This should be `employer_city` for businesses

src/vivarium_census_prl_synth_pop/results_processing/addresses.py
```diff
@@ -43,23 +43,36 @@ def get_address_id_maps(
...
+            column_name,
+            "employer_state_id",
+            "employer_state",
+            "employer_puma",
+        ]
+    else:
+        raise ValueError(
+            f"Expected `address_id` or 'employer_address_id', got `{column_name}`"
+        )
+
```
mattkappel, line 60: A way to avoid the `if/elif` is to try/except a dictionary lookup. I've seen it/used it a bunch in artifact code.   See for example: https://github.com/ihmeuw/vivarium_gates_iv_iron/blob/862b057457cea2410e8d24c2c9c38c43419cd8b2/src/vivarium_gates_iv_iron/data/loader.py#L464-L472

src/vivarium_census_prl_synth_pop/results_processing/addresses.py
albrja: I had to do it this way because doing `unique` first makes it an array so I will not have string functionality.

src/vivarium_census_prl_synth_pop/results_processing/addresses.py
albrja: Yes and I will use this in the other necessary places that it should be used.

src/vivarium_census_prl_synth_pop/results_processing/addresses.py
albrja: This is just creating a column and we are handling the column name when we add the key to the dictionary but I have changed it because I think it makes it a lot clearer.

src/vivarium_census_prl_synth_pop/results_processing/addresses.py
stevebachmeier: for state in map(str.lower, city_data[state_col].unique()):

rmudambi: You could convert the np array to a Series and then have the `.str` functionality.

src/vivarium_census_prl_synth_pop/results_processing/addresses.py
albrja: Yes This one should not be here.  We had added it for zipcode map because state and puma are not unique.  I will remove the unnecessary ones.

rmudambi: Why would there still be duplicates here? Address id should already be unique on it's own, so the combination of address id, puma, and state should definitely be unique. Am I missing something?

src/vivarium_census_prl_synth_pop/results_processing/addresses.py
```diff
@@ -88,54 +101,67 @@ def get_zipcode_map(
...
             randomness_stream=randomness,
             weights=locale_group["proportion"],
-            additional_key=f"{additional_key}zip_map_{state_id}_{puma}",
+            additional_key=f"{additional_key}{column_name}_zip_map_{state_id}_{puma}",
         ).to_numpy()
 
     # Map against obs_data
-    zip_map_dict["zipcode"] = zip_map.astype(int)
+    if column_name == "address_id":
+        zip_map_dict["zipcode"] = zip_map.astype(int)
```
stevebachmeier, line 147: .astype(int) zip_map once during creation

src/vivarium_census_prl_synth_pop/results_processing/addresses.py
```diff
@@ -166,19 +195,28 @@ def get_city_map(
     # Load addresses data from artifact
     addresses = artifact.load(data_keys.SYNTHETIC_DATA.ADDRESSES).reset_index()
     # Get observer data to map
-    output_cols = [column_name, "state"]
-    city_data = obs_data.reset_index()[output_cols].drop_duplicates().set_index("address_id")
+    if column_name == "address_id":
```
rmudambi, line 198: Can use the same dict lookup strategy here that Matt suggested you use above.

src/vivarium_census_prl_synth_pop/results_processing/addresses.py
```diff
@@ -43,23 +43,36 @@ def get_address_id_maps(
...
     maps.update(get_city_map(column_name, formatted_obs_data, artifact, randomness))
-    maps.update(
-        get_mailing_address_map(column_name, formatted_obs_data, artifact, randomness, maps)
-    )
+    if column_name == "address_id":
+        maps.update(
+            get_mailing_address_map(
+                column_name, formatted_obs_data, artifact, randomness, maps
+            )
+        )
```
stevebachmeier, line 75: Maybe just a note saying employers don't have mailing addresses

tests/test_synthetic_pii.py
```diff
@@ -277,7 +277,26 @@ def test_last_name_from_oldest_member(mocker):
...
+        ),
+        (
+            "employer_address_id",
+            "employer_street_number",
+            "employer_street_name",
+            "employer_unit_number",
+        ),
+    ],
+)
+def test_address(
```
stevebachmeier, line 297: consider calling this `test_address_mapping`

src/vivarium_census_prl_synth_pop/results_processing/addresses.py
mattkappel: I think it's not needed actually. I think I originally added the drop_duplicates, but it *should* be unnecessary assuming we've got address_id without duplicates (which should be the case). I think I may have added it as CYA not knowing how we'd build up the column superset for the address_id columns.

src/vivarium_census_prl_synth_pop/results_processing/addresses.py
albrja: I added the note in address.py when we generate mailing address.

rmudambi (approved):

PR #229, opened 2023-03-02
Post processing column handling

albrja:
### Changes how we are writing final results to only include columns desired for each observer outline in the research documentation.
- *Category*: Observers/post-processing
- *JIRA issue*: [MIC-3844](https://jira.ihme.washington.edu/browse/MIC-3844)
- *Research reference*: https://vivarium-research.readthedocs.io/en/latest/models/concept_models/vivarium_census_synthdata/concept_model.html#periodic-observations-of-attributes-through-survey-census-and-registry

### Changes and notes
-Updates observers to return necessary columns desired in final results.
-Changes post-processing to go through each observer, and map only columns in that observer to write to file.
--Improves logging for names mapping.

### Verification and Testing
Successfully ran parallel sim and post-processing.

### DIFF
--- a/src/vivarium_census_prl_synth_pop/components/observers.py
+++ b/src/vivarium_census_prl_synth_pop/components/observers.py
@@ -308,8 +308,16 @@ class WICObserver(BaseObserver):
     """Class for observing columns relevant to WIC administrative data."""
 
     INPUT_VALUES = ["income", "household_details"]
-    ADDITIONAL_INPUT_COLUMNS = ["household_id", "relation_to_household_head",]
-    ADDITIONAL_OUTPUT_COLUMNS = ["wic_year", "household_id", "housing_type", "relation_to_household_head"]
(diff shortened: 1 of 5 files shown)


src/vivarium_census_prl_synth_pop/components/observers.py
```diff
@@ -716,6 +727,8 @@ def get_observation(self, event: Event) -> pd.DataFrame:
             df_w2[col] = df_w2["employer_id"].map(business_details[col])
 
         df_w2 = df_w2.set_index(["simulant_id"])
+
+        df_w2["is_w2"] = True
```
rmudambi, line 731: No need to do anything about this now, but in future when you create a placeholder like this that will be modified in a subsequent PR, it's helpful to add a comment mentioning that this is the case so a reviewer doesn't try to understand why we're always setting this to True

rmudambi (approved):

PR #232, opened 2023-03-06
Feature/sbachmei/mic 3849 current make results to jobmon

stevebachmeier:
### Description
- *Category*: Implementation
- *JIRA issue*: [MIC-3849](https://jira.ihme.washington.edu/browse/MIC-3849)
- *Research reference*: na

### Changes and notes
This first step towards parallelizing make_results is to implement a single-task
jobmon workflow that does nothing more than run the current functionality.
To do so, the main things happen in this PR are:
- The building of the results directory happens immediately so that jobmon
  has a place for its output/error logs.
- The `build_results` function (which is called when you run `make_results`
  from the cli, has been wrapped in click so that jobmon can call it as a
  cli tool (we alternatively could have moved the function in to a runnable
  python script or something).
- A new jobmon tool has been created and registered.
- The jobmon runner script has been set up.
- The `make_results` function has been given additional jobmon-specific
  args.

NOTE: this should be 100% backward compatible. If you want to run
`make_results` the original way on a node you've already checked out,
do so like always. If you want to have jobmon run the job instead,
include the -j, -q, -m, -r, and -P arguments.

NOTE: there may be an issue with jobmon not getting installed correctly.
I've added it to the setup.py, but when I last created my env and installed
the repo, it pulled in a jobmon release candidate for some reason. If this
is indeed a persistent problem, then we may need to just install jobmon
manually.

### Verification and Testing
I ran both with and without -j on a small test simulation and all outputs
were created.


setup.py
```diff
@@ -62,5 +63,6 @@
             [console_scripts]
             make_artifacts=vivarium_census_prl_synth_pop.tools.cli:make_artifacts
             make_results=vivarium_census_prl_synth_pop.tools.cli:make_results
+            build_results=vivarium_census_prl_synth_pop.tools.make_results:build_results
```
rmudambi, line 66: I definitely think it's better to have build results as a runnable python script rather than as a click app. 

Also, click apps should all be defined in `cli.py`.

src/vivarium_census_prl_synth_pop/tools/cli.py
```diff
@@ -69,14 +73,67 @@ def make_artifacts(
...
+@click.option(
+    "-j",
+    "--jobmon",
+    is_flag=True,
+    help="Launches a jobmon workflow to generate results.",
+)
+@click.option(
+    "-q",
+    "--queue",
+    type=click.Choice(["all.q", "long.q"]),
```
rmudambi, line 101: We should have default values for all these arguments.

src/vivarium_census_prl_synth_pop/tools/cli.py
```diff
@@ -69,14 +73,67 @@ def make_artifacts(
...
 )
 @click.option(
     "-t",
     "--test-run",
     is_flag=True,
+    help="Skips updating the 'latest' symlink with this version of results.",
 )
 @click.option(
-    "-a", "--artifact_path", type=click.Path(exists=True), default=paths.DEFAULT_ARTIFACT
+    "-a",
```
mattkappel, line 85: Cosmetic nit: We've been using `-a` for the artifact option since we added it to PRL make_results, but it's been bugging me since we use `-i` for the artifact in all the other commands.

setup.py
```diff
@@ -31,6 +31,7 @@
 
     # use "pip install -e .[dev]" to install required components + extra components
     data_requires = [
+        "jobmon_installer_ihme==10.6.0",
```
stevebachmeier, line 34: Add jobmon stuff to README

src/vivarium_census_prl_synth_pop/tools/cli.py
```diff
@@ -85,7 +142,50 @@ def make_results(
...
+    raw_output_dir, final_output_dir = _build_final_results_directory(output_dir)
+    cluster_requests = {
+        "queue": queue,
+        "peak_memory": peak_memory,
+        "max_runtime": max_runtime,
+        "project": project,
+    }
+    if not jobmon and (queue or peak_memory or max_runtime or project):
+        requests = {k: v for k, v in cluster_requests.items() if v is not None}
+        logger.info(
```
stevebachmeier, line 163: make this warning

src/vivarium_census_prl_synth_pop/tools/cli.py
```diff
@@ -85,7 +142,50 @@ def make_results(
...
+        func = run_make_results_workflow
+        kwargs = cluster_requests
+    else:  # run from current node
+        func = do_build_results
+        kwargs = {}
+    main = handle_exceptions(func=func, logger=logger, with_debugger=with_debugger)
+    main(raw_output_dir, final_output_dir, mark_best, test_run, artifact_path, **kwargs)
+
+
+def _build_final_results_directory(results_dir: str) -> Tuple[Path, Path]:
```
stevebachmeier, line 184: move to make_results.py or utilities.py

src/vivarium_census_prl_synth_pop/tools/make_results.py
```diff
@@ -196,17 +197,56 @@
...
+@click.option("-b", "--mark-best", is_flag=True)
+@click.option(
+    "-t",
+    "--test-run",
+    is_flag=True,
+)
+@click.option(
+    "-a", "--artifact-path", type=click.Path(exists=True), default=paths.DEFAULT_ARTIFACT
+)
 def build_results(
```
stevebachmeier, line 219: Remove cli wrappipng and have workflow call `make_results`

src/vivarium_census_prl_synth_pop/tools/jobmon.py
```diff
@@ -0,0 +1,89 @@
...
+) -> None:
+    """Creates and runs a jobmon workflow to build results datasets
+    from the raw data output by observers
+    """
+    logger.info(f"Starting make_results workflow {final_output_dir}")
+    wf_uuid = uuid.uuid4()
+
+    # Deal with boolean args - click either the flag or nothing, not True/False
+    # TODO: there's got to be a better way to do this?
+    if mark_best:
```
stevebachmeier, line 29: From Hussain: consider `mark_best_arg = "--mark-best" if mark_best else ""`

setup.py
stevebachmeier: As discussed, will try having `make_results -j` just turn around and run `make_results`. That way we can access both jobmony or normal processing w/o an additional cli tool

src/vivarium_census_prl_synth_pop/__about__.py
```diff
@@ -14,6 +14,7 @@
 __uri__ = "https://github.com/ihmeuw/vivarium_census_prl_synth_pop"
 
 __version__ = "1.0.0"
+__jobmon_tool_version__ = 1212
```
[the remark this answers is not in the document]
stevebachmeier: Ok, so what I didn't realize was that when I saw this on the covid team, it was actually being used in the background. I'll remove for now but there's some good discussion on slack about how we might want to wrap the jobmon stuff.

src/vivarium_census_prl_synth_pop/tools/make_results.py
stevebachmeier: This didn't work b/c make_results requires just output_dir but at this point we've already split that into raw_dir and final_dir.

src/vivarium_census_prl_synth_pop/tools/jobmon.py
```diff
@@ -0,0 +1,83 @@
...
+    artifact_path: Union[str, Path],
+    queue: str,
+    peak_memory: int,
+    max_runtime: str,
+    project: str,
+) -> None:
+    """Creates and runs a jobmon workflow to build results datasets
+    from the raw data output by observers
+    """
+    logger.info(f"Starting make_results workflow {final_output_dir}")
```
albrja, line 24: Nit: Make this something like "saving to final_output_dir", "writing to", or :"processing results at".

src/vivarium_census_prl_synth_pop/tools/jobmon.py
```diff
@@ -0,0 +1,83 @@
...
+    template_make_results = tool.get_task_template(
+        template_name="make_results_template",
+        default_compute_resources={
+            "queue": queue,
+            "cores": 1,
+            "memory": peak_memory,
+            "runtime": max_runtime,
+            "project": project,
+            "stdout": str(final_output_dir),
+            "stderr": str(final_output_dir),
```
albrja, line 48: Are these just sending the output to that directory or does a log file get created?

albrja (approved):

src/vivarium_census_prl_synth_pop/tools/jobmon.py
stevebachmeier: Well, this is the final output directory for the real datasets (same as always). This is just telling jobmon to send the output logs there as well.

src/vivarium_census_prl_synth_pop/tools/jobmon.py
stevebachmeier: I don't mind changing, but I do want to clarify that the word "workflow" is a very jobmon-y term and I was hoping to capture that here.

src/vivarium_census_prl_synth_pop/tools/cli.py
```diff
@@ -69,14 +72,75 @@ def make_artifacts(
...
+    help="NOTE: only required if --jobmon. "
+    "The cluster queue to assign jobmon tasks to. long.q allows for much "
+    "longer runtimes although there may be reasons to send jobs to that queue "
+    "even if they don't have runtime constraints (eg node availability). ",
+)
+@click.option(
+    "-m",
+    "--peak-memory",
+    type=int,
+    default=200,
```
rmudambi, line 112: This is a lot of memory? We should be able to decrease this when we actually get parallel, right?

src/vivarium_census_prl_synth_pop/tools/cli.py
```diff
@@ -69,14 +72,75 @@ def make_artifacts(
...
+    show_default=True,
+    help="NOTE: only required if --jobmon. "
+    "The estimated maximum memory usage in GB of an individual simulate job. "
```
rmudambi, line 115: "Simulate job" doesn't make sense as we're not doing a simulation here

src/vivarium_census_prl_synth_pop/tools/cli.py
```diff
@@ -69,14 +72,75 @@ def make_artifacts(
...
+    "--max-runtime",
+    type=str,
+    default="4:00:00",
+    show_default=True,
+    help="NOTE: only required if --jobmon. "
+    "The estimated maximum runtime ('hh:mm:ss') of the jobmon tasks. "
+    "Once this time limit is hit, the cluster will terminate jobs regardless of "
+    "queue. The maximum supported runtime is 3 days. Keep in mind that the "
+    "session you are launching from must be able to live at least as long "
+    "as the simulation jobs, and that runtimes by node vary wildly. ",
```
rmudambi, line 129: Again change from simulation

src/vivarium_census_prl_synth_pop/tools/cli.py
```diff
@@ -69,14 +72,75 @@ def make_artifacts(
...
+)
+@click.option(
```
rmudambi, line 131: Don't give the user a choice. They should always be using `proj_simscience_prod`

src/vivarium_census_prl_synth_pop/tools/cli.py
```diff
@@ -85,7 +149,72 @@ def make_results(
...
+    is_flag=True,
+)
+@click.option(
+    "-i", "--artifact-path", type=click.Path(exists=True), default=paths.DEFAULT_ARTIFACT
+)
+def jobmon_make_results_runner(
+    raw_output_dir: Path,
+    final_output_dir: Path,
+    verbose: int,
+    with_debugger: bool,
```
rmudambi, line 213: Isn't this parameter only relevant in the non-jobmon workflow?

src/vivarium_census_prl_synth_pop/tools/jobmon.py
```diff
@@ -0,0 +1,86 @@
...
+        },
+        default_cluster_name="slurm",
+        command_template=(
+            f"{shutil.which('jobmon_make_results_runner')} "
+            "{raw_output_dir} "
+            "{final_output_dir} "
+            "{mark_best} "
```
rmudambi, line 58: It occurs that we may need a slightly more complicated workflow than this. We will want to wait until all jobs have completed successfully before we mark the results as best or latest. This can be addressed in a later PR though.

src/vivarium_census_prl_synth_pop/tools/jobmon.py
```diff
@@ -0,0 +1,86 @@
...
+            "{test_run} "
+            "--artifact-path {artifact_path} "
+            "-vvv "
```
rmudambi, line 61: Shouldn't this level depend on the logging verbosity set by the user?

rmudambi (approved):

src/vivarium_census_prl_synth_pop/tools/cli.py
stevebachmeier: Yes, def. This is based on the current memory req'ts for a 1 million popsize

src/vivarium_census_prl_synth_pop/tools/cli.py
stevebachmeier: Are you sure? What abvout when debugging? I've been doing all this development on `proj_simscience`, eg.

stevebachmeier: Will either remove the option altogether or change it to the default

PR #234, opened 2023-03-08
Add ssns and itins to artifact

stevebachmeier:
### Description
- *Category*: data artifact
- *JIRA issue*: [MIC-3850](https://jira.ihme.washington.edu/browse/MIC-38520)
- *Research reference*: na

### Changes and notes
This adds a randomly shuffled SSN and ITIN dataframes to the artifact
to later be used to sample from instead of generating on the fly. The end goal
of this approach is that we can prevent unwanted SSN collision in different
shards by slicing the dataframe appropriately.

Some notes about these two dataframes:
- They include ALL POSSIBLE combinations
- They have been randomly shuffled
- They consist of three columns of integers which represent the three
  parts to SSNs and ITINs. Processing these into hyphentated strings
  will happen during post-processing. This was the chosen approach
  because it's 5x quicker to load a dataframe w/ integers comnpared
  to a series of processed string IDs. It's then very quick to
  convert from 3 int columns to "xxx-xx-xxxx".

### Verification and Testing
Added to a test artifact and confirmed the data is there.

### DIFF
--- a/src/vivarium_census_prl_synth_pop/data/loader.py
+++ b/src/vivarium_census_prl_synth_pop/data/loader.py
@@ -12,6 +12,7 @@ for an example.
 
    No logging is done here. Logging is done in vivarium inputs itself and forwarded.
 """
+from itertools import product
 from pathlib import Path
 from typing import Dict, List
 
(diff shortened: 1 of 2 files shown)


src/vivarium_census_prl_synth_pop/data/loader.py
```diff
@@ -305,6 +308,58 @@ def load_address_data(key: str, location: str) -> pd.DataFrame:
...
+
+def _generate_randomized_id_df(
+    area: List[int], group: List[int], serial: List[int]
+) -> pd.DataFrame:
+    """Given lists of allowable area, group, and serial integers, generate a
+    randomly sorted dataframe of the product of the three groups
+    """
+    parts = [(area, group, serial) for area, group, serial in product(area, group, serial)]
+    ids = pd.DataFrame(data=parts, columns=["area", "group", "serial"])
+    ids = ids.sample(frac=1).reset_index(drop=True)
```
rmudambi, line 358: You should provide a seed here so we are guaranteed to get the same artifact every time we rebuild it.

rmudambi (approved):
Feel free to merge once you add the seed.

src/vivarium_census_prl_synth_pop/data/loader.py
stevebachmeier: Whoops, good catch

PR #235, opened 2023-03-09
Feature/sbachmei/mic 3851 sample ssn from artifact

stevebachmeier:
### Description
- *Category*: implementation
- *JIRA issue*: [MIC-3851](https://jira.ihme.washington.edu/browse/MIC-3851)
- *Research reference*: na

### Changes and notes
Instead of using the `generate_ssn` method, this PR uses the SSN artifact data.
That data has already been randomly shuffled and is stored that way in the artifact,
so all that is needed is to pick off the top the correct number of SSNs.

### Verification and Testing
Confirmed sim runs and various qc checks.

### DIFF
--- a/src/vivarium_census_prl_synth_pop/results_processing/names.py
+++ b/src/vivarium_census_prl_synth_pop/results_processing/names.py
@@ -270,8 +270,16 @@ def get_employer_name_map(
         data_values.MilitaryEmployer.EMPLOYER_ID, "employer_name"
     ] = data_values.MilitaryEmployer.EMPLOYER_NAME
 
+    # NOTE: This deduplication stuff will all be deleted in a following PR
     duplicate_mask = employer_names.duplicated(subset=["employer_name"])
     while sum(duplicate_mask) > 0:
+        if counter == 10:
+            logger.info(
+                f"Resampled employer names {counter} times and data still contains {duplicate_mask.sum()}"
+                f"duplicates.  Check data or usage of randomness stream."
+            )
+            break
+        counter += 1
         employer_names.loc[duplicate_mask, "employer_name"] = vectorized_choice(
             options=business_names_data["business_names"],
             n_to_choose=duplicate_mask.sum(),
@@ -279,12 +287,6 @@ def get_employer_name_map(
             additional_key=f"employer_names_{counter}",
         ).to_numpy()
         duplicate_mask = employer_names.duplicated(subset=["employer_name"])
-        counter += 1
-        if counter == 10:
-            logger.info(
-                f"Resampled employer names {counter} times and data still contains {duplicate_mask.sum()}"
-                f"duplicates remaining.  Check data or usage of randomness stream."
-            )
 
     employer_name_map = {"employer_name": employer_names["employer_name"]}
     return employer_name_map
--- a/src/vivarium_census_prl_synth_pop/results_processing/ssn_and_itin.py
+++ b/src/vivarium_census_prl_synth_pop/results_processing/ssn_and_itin.py
@@ -204,7 +204,7 @@ def get_ssn_map(
     )
     ssn_map = pd.Series("", index=simulant_data.index)
     # Load (already-shuffled) SSNs and choose the appropriate number off the top
-    # TODO: look for a more performant method (it took ~3 minutes for ~1 million simulants. Good enough for now.)
+    # TODO: look for a more performant method (it takes several minutes - good enough for now.)
     ssns = artifact.load("synthetic_data.ssns")[: simulant_data["has_ssn"].sum()]
     # Convert to hyphenated string IDs
     ssns = (
--- a/src/vivarium_census_prl_synth_pop/results_processing/ssn_and_itin.py
+++ b/src/vivarium_census_prl_synth_pop/results_processing/ssn_and_itin.py
@@ -46,7 +46,7 @@ def get_simulant_id_maps(
         obs_results=obs_data,
         output_columns=output_cols_superset,
     )
-    maps.update(get_ssn_map(column_name, formatted_obs_data, randomness))
+    maps.update(get_ssn_map(column_name, formatted_obs_data, artifact))
     return maps
 
 
@@ -139,7 +139,6 @@ def generate_ssns(
     size: int,
     additional_key: Any,
     randomness: RandomnessStream,
-    allow_duplicates: bool = False,
 ) -> np.ndarray:
     """
     Generate a np.ndarray of length `size` of unique SSN values.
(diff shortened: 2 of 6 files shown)


tests/test_synthetic_pii.py
```diff
@@ -469,7 +469,8 @@ def test_employer_name_map(mocker):
     assert not (employer_names["employer_name"].isnull().any())
 
 
-def test_ssn_mapping():
+@pytest.mark.slow
```
stevebachmeier, line 472: To run this you'd now need to do `pytest --runslow`. This allows for quick unit tests as well as passing with CI (since CI does not runslow). It's a compromise.

src/vivarium_census_prl_synth_pop/results_processing/ssn_and_itin.py
```diff
@@ -139,7 +139,6 @@ def generate_ssns(
     size: int,
     additional_key: Any,
     randomness: RandomnessStream,
-    allow_duplicates: bool = False,
```
stevebachmeier, line 142: The key difference here in the `generate_ssns` method is that I've removed the "allow_duplicates" arg b/c there's no use for it anymore (the only time this gets called is when a simulant is making an SSN up on the spot and duplicates are allowed in that case.

src/vivarium_census_prl_synth_pop/results_processing/ssn_and_itin.py
```diff
@@ -151,88 +150,79 @@ def generate_ssns(
     :param size: The number of itins to generate
     :param additional_key: Additional key to be used by randomness
     :param randomness: RandomnessStream stream to use
-    :param allow_duplicates: True if duplicates are allowed, False (default) otherwise
-
     """
+    ssn = pd.DataFrame(index=pd.Index(range(size)))
 
-    def generate_ssn(
```
stevebachmeier, line 158: Since we no longer need to check for duplicates, this internal function that was previously being called in a while loop (while there were duplicates) is no longer necessary - all of this change is just removing the function definition and then de-indenting (outdenting?)

src/vivarium_census_prl_synth_pop/results_processing/names.py
```diff
@@ -270,21 +270,23 @@ def get_employer_name_map(
...
 
+    # NOTE: This deduplication stuff will all be deleted in a following PR
     duplicate_mask = employer_names.duplicated(subset=["employer_name"])
     while sum(duplicate_mask) > 0:
+        if counter == 10:
+            logger.info(
+                f"Resampled employer names {counter} times and data still contains {duplicate_mask.sum()}"
+                f"duplicates.  Check data or usage of randomness stream."
+            )
+            break
```
albrja, line 281: Nice

src/vivarium_census_prl_synth_pop/results_processing/ssn_and_itin.py
```diff
@@ -46,7 +46,7 @@ def get_simulant_id_maps(
         obs_results=obs_data,
         output_columns=output_cols_superset,
     )
-    maps.update(get_ssn_map(column_name, formatted_obs_data, randomness))
+    maps.update(get_ssn_map(column_name, formatted_obs_data, artifact))
```
albrja, line 49: I am not sure I understand how this is working without a randomness stream?

albrja: Generate SSNs never gets called it was just updated for the artifact?

tests/test_synthetic_pii.py
stevebachmeier: This is no longer necessary - the pytests now work quickly.

src/vivarium_census_prl_synth_pop/results_processing/ssn_and_itin.py
stevebachmeier: Yeah so the `get_ssn_map` happens without randomness stream b/c it load the already-randomly-shuffled SSN options and just picks off the top (no sampling needed b/c they're already shuffled).

The only time `generate_ssns` is called is when a simulant makes up a SSN on the spot, and that happens after the maps are already generated as part of the `for observer in FINAL_OBSERVERS:` loop

src/vivarium_census_prl_synth_pop/results_processing/ssn_and_itin.py
```diff
@@ -254,10 +246,11 @@ def do_collide_ssns(
         & (obs_data["employer_id"] != -1)
         & (~obs_data["has_ssn"])
     )
+    # TODO: Replace with sampling from the artifact data if we figure out a
+    # more performant way of doing so (eg querying the hdf with randomized idx)
```
rmudambi, line 250: This should be easy? Just generate a N random ints in the range [0:len(ssns_in_artifact)] and store them as an index. Then look up those indices in the artifact.

```
ssn_idx = pd.Index(
    random_integers(
        min_val=0,
        max_val=num_possible_ssns,
        index=pd.Index(range(create_ssn_mask.sum()),
        randomness=randomness,
        additional_key=f"{seed}_collide",
    )
)
obs_data.loc[create_ssn_mask, "ssn"] = ssns_from_artifact.loc[ssn_idx]
```

tests/conftest.py
```diff
@@ -0,0 +1,19 @@
+from pathlib import Path
+
+import pytest
+from vivarium import Artifact
+
+
+@pytest.fixture(scope="function")
```
rmudambi, line 7: Function is the default scope, so you don't need to specify it, fyi

tests/test_synthetic_pii.py
```diff
@@ -478,9 +478,12 @@ def test_ssn_mapping():
     simulants["has_ssn"] = [True if n % 2 == 0 else False for n in range(num_unique_ids)]
 
     # Get the map
-    ssn_map = get_simulant_id_maps(
-        "simulant_id", {"silly_observer": simulants}, None, randomness
-    )["ssn"]
+    try:
```
rmudambi, line 481: Nice I like this.

src/vivarium_census_prl_synth_pop/results_processing/ssn_and_itin.py
stevebachmeier: I don't think this solves the problem because it's still loading all ssns_from_artifact which is what takes a long time. That said, I don't think this takes all that long and isn't a big deal anyway

tests/test_synthetic_pii.py
stevebachmeier: Seemed like a good compromise. If we want this functionality everywhere then we can add to vct or something.

rmudambi (approved):
Conditional on the two nits we discussed.

PR #237, opened 2023-03-09
Feature/sbachmei/mic 3887 map itins

stevebachmeier:
### Description
- *Category*: implementation
- *JIRA issue*: [MIC-3887](https://jira.ihme.washington.edu/browse/MIC-3851)
- *Research reference*: na

### Changes and notes
This adds the ITIN column to the 1040 form results. It also refactors so that
instead of generating ITINs on the fly it samples off the top of the ITIN 
artifact.

### Verification and Testing
tests pass and post-processing a small simulation confirmed that the 1040
form includes ITIN wherever a simulant does not have an SSN

### DIFF
--- a/src/vivarium_census_prl_synth_pop/results_processing/ssn_and_itin.py
+++ b/src/vivarium_census_prl_synth_pop/results_processing/ssn_and_itin.py
@@ -247,6 +247,8 @@ def do_collide_ssns(
         & (obs_data["employer_id"] != -1)
         & (~obs_data["has_ssn"])
     )
+    # TODO: Replace with sampling from the artifact data if we figure out a
+    # more performant way of doing so (eg querying the hdf with randomized idx)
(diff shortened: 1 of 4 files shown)


src/vivarium_census_prl_synth_pop/results_processing/ssn_and_itin.py
```diff
@@ -228,6 +138,47 @@ def get_ssn_map(
     return ssn_map_dict
 
 
+def get_itin_map(
```
rmudambi, line 141: I think `get_itin_map` and `get_ssn_map` are identical, except they read from a different artifact key. That being said, this won't be the case after you account for my comment in the other pr about the `ssn_id` column. Nevertheless, there seems to be scope to consolidate the code.

rmudambi (approved):

PR #238, opened 2023-03-10
Post processing single seed

rmudambi:
### Description

- *Category*: post-processing
- *JIRA issue*: [MIC-3852](https://jira.ihme.washington.edu/browse/MIC-3852)
- *Research reference*: 

### Changes and notes
Adds a new flag to the cli that allows for post processing to be run on a single seed. 

### Verification and Testing
Ran post processing on a recent simulation output with a single seed and verified in a notebook that the outputs look correct and are roughly 10% the size of outputs generated by running without the seed flag.

### DIFF
--- a/src/vivarium_census_prl_synth_pop/tools/cli.py
+++ b/src/vivarium_census_prl_synth_pop/tools/cli.py
@@ -85,8 +85,8 @@ def make_artifacts(
     "--seed",
     type=str,
     default="",
-    help="Provide seed in order to run only on the files corresponding to that seed. "
-         "If no seed it provided, will run on all files."
+    help="Provide seed in order to run only on the files corresponding to that "
+    "seed. If no seed it provided, will run on all files.",
 )
 @click.option(
     "-i",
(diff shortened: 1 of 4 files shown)


albrja (approved):

src/vivarium_census_prl_synth_pop/tools/cli.py
```diff
@@ -168,7 +177,7 @@ def make_results(
         func = build_results
         kwargs = {}
     main = handle_exceptions(func=func, logger=logger, with_debugger=with_debugger)
-    main(raw_output_dir, final_output_dir, mark_best, test_run, artifact_path, **kwargs)
+    main(raw_output_dir, final_output_dir, mark_best, test_run, seed, artifact_path, **kwargs)
```
stevebachmeier, line 180: I think this is going to break the `-j` stuff. I'll fix later if needed

rmudambi: Why would this break `-j`? There's a default value, so seed will just be the empty string in that case.

rmudambi: Good call. Should be fixed now, though we aren't using the argument in the jobmon workflow at all yet.

stevebachmeier (approved):

PR #241, opened 2023-03-13
Feature/sbachmei/mic 3887 map itins

stevebachmeier:
### Description
- *Category*: implementation
- *JIRA issue*: [MIC-3887](https://jira.ihme.washington.edu/browse/MIC-3851)
- *Research reference*: na

### Changes and notes
This adds the ITIN column to the 1040 form results. It also refactors so that
instead of generating ITINs on the fly it samples off the top of the ITIN 
artifact.

### Verification and Testing
tests pass and post-processing a small simulation confirmed that the 1040
form includes ITIN wherever a simulant does not have an SSN

### DIFF
--- a/src/vivarium_census_prl_synth_pop/results_processing/ssn_and_itin.py
+++ b/src/vivarium_census_prl_synth_pop/results_processing/ssn_and_itin.py
@@ -141,9 +141,7 @@ def get_itin_map(
         obs_results=obs_data,
         output_columns=[column_name, "has_ssn"],
     )
-    simulant_data = (
-        formatted_obs_data.reset_index().drop_duplicates().set_index(column_name)
-    )
+    simulant_data = formatted_obs_data.reset_index().drop_duplicates().set_index(column_name)
     # Load (already-shuffled) ITINs and choose the appropriate number off the top
     # NOTE: artifact.load does not currently have the option to select specific rows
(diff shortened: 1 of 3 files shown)


tests/test_synthetic_pii.py
```diff
@@ -31,25 +31,6 @@ def get_draw(self, index, additional_key=None) -> pd.Series:
     return pd.Series(data=s, index=index)
 
 
-def test_itin_generation():
```
stevebachmeier, line 34: Not such a useful test anymore since we are relying on static artifact data

src/vivarium_census_prl_synth_pop/results_processing/ssn_and_itin.py
```diff
@@ -34,95 +34,10 @@ def get_simulant_id_maps(
     """
     if column_name != "simulant_id":
         raise ValueError(f"Expected `simulant_id`, got `{column_name}`")
-
-    ssn_map = get_ssn_map(obs_data, column_name, artifact)
-
-    return ssn_map
-
-
-def generate_itins(
```
stevebachmeier, line 43: We are using the artifact data instead of the function now. There is also no concept of "making up" ITINs like there is for SSNs so we really don't need this

rmudambi (approved):

PR #243, opened 2023-03-13
implement seed arg for jobmon

stevebachmeier:
### Description
- *Category*: Implementation
- *JIRA issue*: [MIC-3853](https://jira.ihme.washington.edu/browse/MIC-3853)
- *Research reference*: na

### Changes and notes
- Adds `seed` argument logic to the jobmon workflow
- If empty, continues to run on all seeds and concat observer results
- If not empty, saves seed-specific files to respective observer folders

### Verification and Testing
works on small sim

### DIFF
--- a/src/vivarium_census_prl_synth_pop/tools/cli.py
+++ b/src/vivarium_census_prl_synth_pop/tools/cli.py
@@ -85,6 +85,7 @@ def make_artifacts(
     "--seed",
     type=str,
     default="",
+    show_default=True,
     help="Provide seed in order to run only on the files corresponding to that "
     "seed. If no seed it provided, will run on all files.",
 )
@@ -191,6 +192,15 @@ def make_results(
     is_flag=True,
 )
 @click.option(
+    "-s",
+    "--seed",
+    type=str,
+    default="",
+    show_default=True,
+    help="Provide seed in order to run only on the files corresponding to that "
+    "seed. If no seed it provided, will run on all files.",
+)
+@click.option(
     "-i", "--artifact-path", type=click.Path(exists=True), default=paths.DEFAULT_ARTIFACT
 )
 def jobmon_make_results_runner(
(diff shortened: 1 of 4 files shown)


src/vivarium_census_prl_synth_pop/tools/cli.py
```diff
@@ -85,6 +85,7 @@ def make_artifacts(
     "--seed",
     type=str,
     default="",
+    show_default=True,
```
rmudambi, line 88: What does show_default do?

src/vivarium_census_prl_synth_pop/tools/jobmon.py
```diff
@@ -53,27 +54,31 @@ def run_make_results_workflow(
...
             "{raw_output_dir} "
             "{final_output_dir} "
             "{verbose} "
             "{mark_best} "
             "{test_run} "
+            "{seed} "
             "--artifact-path {artifact_path} "
         ),
-        node_args=[],
+        # node_args=["seed"],  # TODO: parameterize by seed
```
rmudambi, line 68: What is a node arg?

src/vivarium_census_prl_synth_pop/tools/make_results.py
```diff
@@ -268,8 +268,11 @@ def perform_post_processing(
 
         obs_data = obs_data[list(FINAL_OBSERVERS[observer])]
         logger.info(f"Writing final results for {observer}.")
+        obs_dir = final_output_dir / observer
+        obs_dir.mkdir(parents=True, exist_ok=False)
```
rmudambi, line 272: Why is `exist_ok` False? This seems highly problematic when we're parallelized, since we want all the parallel jobs to write to the same directory. Also, why are we now creating the directory at this stage? When were we creating it before? Are we still creating it at that time, because I don't see any changes that remove directory creation.

src/vivarium_census_prl_synth_pop/tools/cli.py
stevebachmeier: Just shows what the default value is when you `--help`

src/vivarium_census_prl_synth_pop/tools/jobmon.py
stevebachmeier: It's the thing that we will be parameterizing by (like,  nodes in the dag)

https://scicomp-docs.ihme.washington.edu/jobmon/current/core_concepts.html#node-args

src/vivarium_census_prl_synth_pop/tools/make_results.py
stevebachmeier: Oh yeah, good call. 

No, we have never created final results in observer directories. It seemed worthwhile since we are now spitting out many files per observer.

rmudambi (approved):

PR #246, opened 2023-03-14
Enforce uniique SSNs and ITINs across seeds

stevebachmeier:
### Description
- *Category*: implementation
- *JIRA issue*: [MIC-3855](https://jira.ihme.washington.edu/browse/MIC-3855)
- *Research reference*: na

### Changes and notes
Currently we have SSN and ITIN duplication when each simulation seed is
processed in parallel because they all pick off the top of the artifact data.
This PR assigns unique chunks of the artifact data to each seed to
then choose from

### Verification and Testing
added pytest. Ran small make_results to make sure files are still output.

### DIFF
--- a/src/vivarium_census_prl_synth_pop/results_processing/names.py
+++ b/src/vivarium_census_prl_synth_pop/results_processing/names.py
@@ -1,4 +1,4 @@
-from typing import Any, Dict
+from typing import Any, Dict, List
 
 import numpy as np
 import pandas as pd
@@ -19,6 +19,7 @@ def get_given_name_map(
     artifact: Artifact,
     randomness: RandomnessStream,
     seed: str,
+    _: List[int],
 ) -> Dict[str, pd.Series]:
     """
     Parameters:
@@ -57,6 +58,7 @@ def get_middle_initial_map(
     artifact: Artifact,
     randomness: RandomnessStream,
     seed: str,
+    all_seeds: List[int],
 ) -> Dict[str, pd.Series]:
     """
     Parameters:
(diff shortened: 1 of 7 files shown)


stevebachmeier:
Oh, big note: currently this `make_results` returns very different processed data if you happen to run all seeds without jobmon, eg `make_results` (note no `-j` or `-s` arguments). In this case - which is what we've all gotten used to - the final processed results will be one large dataset per observer. In any other situation, you will get seed-specific datasets in their respective observer folder.
- `make_results ...` --> one large dataset per observer
- `make_results -s <seed>` --> one seed-specific dataset per observer, eg <observer>/<observer>_<seed>.csv.bz2
- `make_results -j` --> all seed-specific datasets in their respective observer folders, same format as above
- `make_results -j -s <seed>` --> one seed-specific dataset per observer for <seed>, same format as above

src/vivarium_census_prl_synth_pop/results_processing/names.py
```diff
@@ -19,6 +19,7 @@ def get_given_name_map(
     artifact: Artifact,
     randomness: RandomnessStream,
     seed: str,
+    _: List[int],
```
rmudambi, line 22: I'd recommend doing this for all of the mapper functions (other than ssn and itin obviously). Then you don't have to pass the `all_seeds` argument in `get_middle_initial_map`

```suggestion
    *_: Any,
```

rmudambi (approved):

src/vivarium_census_prl_synth_pop/results_processing/names.py
stevebachmeier: lol well that's much nicer than my ["dummy_list"]s

PR #253, opened 2023-03-28
floor age and enforce zipcode as int

stevebachmeier:
### Description
- *Category*: bugfix 
- *JIRA issue*: [MIC-3932](https://jira.ihme.washington.edu/browse/MIC-3932)
- *Research reference*: na

### Changes and notes
This PR does two simple things. When post-processing:
- Force zipcodes to be integers (currently they are floats)
- Modify the "real" age (which is a floag, eg 39.384623) to int (eg 39)

### Verification and Testing
spot checked a bit. I'm re-running make_results on the full-pop sim that Jim ran last week.

### DIFF
--- a/src/vivarium_census_prl_synth_pop/results_processing/addresses.py
+++ b/src/vivarium_census_prl_synth_pop/results_processing/addresses.py
@@ -154,7 +154,8 @@ def get_zipcode_map(
             weights=locale_group["proportion"],
             additional_key=f"{additional_key}{column_name}_zip_map_{state_id}_{puma}_{seed}",
         ).to_numpy()
-
+    # Convert to 0-padded strings
+    zip_map = zip_map.astype(str).str.zfill(5)
     # Map against obs_data
     if column_name == "address_id":
         zip_map_dict["zipcode"] = zip_map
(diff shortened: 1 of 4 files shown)


src/vivarium_census_prl_synth_pop/tools/make_results.py
```diff
@@ -308,6 +309,8 @@ def load_data(raw_results_dir: Path, seed: str) -> Dict[str, pd.DataFrame]:
                 else:
                     df = pd.read_csv(file)
                 df["random_seed"] = file.name.split(".")[0].split("_")[-1]
+                if "age" in df.columns:
```
rmudambi, line 312: The proper way to do this is to add a formatter to the age column that gets called by `formatter.format_columns()`

rmudambi (changes requested):

src/vivarium_census_prl_synth_pop/results_processing/addresses.py
```diff
@@ -127,7 +127,7 @@ def get_zipcode_map(
         )
 
     simulation_addresses = obs_data.reset_index()[output_cols].set_index(column_name)
-    zip_map = pd.Series(index=simulation_addresses.index, dtype=int)
+    zip_map = pd.Series(-1, index=simulation_addresses.index, dtype=int)
```
rmudambi, line 130: How do we handle zipcodes that start with 0?

aflaxman:
Thank you!

PR #255, opened 2023-04-01
Feature/hdf

rmudambi:
### Description

- *Category*: observers/post-processing
- *JIRA issue*: [MIC-3942](https://jira.ihme.washington.edu/browse/MIC-3942)
- *Research reference*: 

### Changes and notes
Write output of sim and of post-processing as bzip2 compressed hdf files. 
Updated default simulation size to 1 million simulants.

### Verification and Testing
Ran simulation and post-processing and verified that files are written out correctly and can be read back using `pd.read_hdf`

### DIFF
--- a/src/vivarium_census_prl_synth_pop/components/observers.py
+++ b/src/vivarium_census_prl_synth_pop/components/observers.py
@@ -171,7 +171,7 @@ class BaseObserver(ABC):
             self.responses.index.names = ["simulant_id"]
             filepath = output_dir / f"{self.name}_{self.seed}.{self.file_extension}"
             if "hdf" == self.file_extension:
-                self.responses.to_hdf(filepath, "data", format="table")
+                self.responses.to_hdf(filepath, "data", format="table", complib=9)
             else:
                 self.responses.to_csv(filepath)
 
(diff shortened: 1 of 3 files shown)


src/vivarium_census_prl_synth_pop/components/observers.py
```diff
@@ -171,7 +171,9 @@ def on_simulation_end(self, _: Event) -> None:
             self.responses.index.names = ["simulant_id"]
             filepath = output_dir / f"{self.name}_{self.seed}.{self.file_extension}"
             if "hdf" == self.file_extension:
-                self.responses.to_hdf(filepath, "data", format="table")
+                self.responses.to_hdf(
+                    filepath, "data", format="table", complib="bzip2", complevel=9
```
stevebachmeier, line 175: Did you profile these options to land on this complib and complevel?

stevebachmeier (approved):

src/vivarium_census_prl_synth_pop/components/observers.py
rmudambi: No. If we have time to do profiling later, we will do that, but we will almost certainly want to use hdf files to preserve dtypes, so I'm proactively changing this.

PR #256, opened 2023-04-03
Add fuzzy checking of proportions and use in a few simple cases

zmbc:
### Description
- *Category*: test
- *Research reference*: none

### Changes and notes

**Terminology**:
- Here, I am considering our existing integration tests to be part of "V&V" (which I write "v_and_v" in file paths). They are the automated part, while our traditional V&V process is the manual part. This PR adds automated V&V/integration testing **of proportions** (that are not 0 or 1).
- "Fuzzy checking" is the term I am using for checking anything that is subject to sampling error/stochastic variation. For the time being, this is only proportions, but could be expanded to other types of quantities in the future.

Statistically, this uses a frequentist approach to test a family of hypotheses with Bonferroni correction. Each hypothesis is a binomial test of a proportion being within a range. A full description of the statistics can be found at https://github.com/ihmeuw/vivarium_research/pull/1181.

Technically, this uses a pytest fixture that is kept active for the entire test session and allows testing these hypotheses while maintaining a log of information about them. This diagnostic information is output at the end of a test run for optional human inspection. I have gitignored the diagnostic output -- is this the right call?

Initially, we only fuzzy-check:
* How addresses change between geographic areas (state, PUMA)
* The distribution of states in the simulation

### Verification and Testing

Ran integration tests with multiple population sizes.


zmbc:

integration_tests/conftest.py
```diff
@@ -74,3 +78,198 @@ def pipeline_columns(sim, populations) -> List[str]:
...
+            Path(os.path.dirname(__file__)) / "v_and_v_output/proportion_tests.csv",
+            index=False,
+        )
+
+    def _two_tailed_binomial_test(
+        self, numerator, denominator, null_hypothesis_min, null_hypothesis_max
+    ):
+        # Adapted from https://github.com/scipy/scipy/blob/c1ed5ece8ffbf05356a22a8106affcd11bd3aee0/scipy/stats/_binomtest.py#L202-L333
+        # This is a totally home-grown way to make the null hypothesis a range.
+        # Is this reasonable?
```
alibow, line 160: This is how I would have done it!

integration_tests/conftest.py
```diff
@@ -74,3 +78,198 @@ def pipeline_columns(sim, populations) -> List[str]:
...
+
+@pytest.fixture(scope="session")
+def fuzzy_checker() -> FuzzyChecker:
+    tester = FuzzyChecker(
+        # NOTE: This will need to be updated when any new tests with fuzzy asserts
+        # are added.
+        # Do not update this if it is not the only thing that is failing!
+        # Early exits from failing tests can make the number of asserts actually performed
+        # different than what it would be with passing tests.
+        num_comparisons=601,
```
alibow, line 264: Where is this coming from? just a manual calculation of the number of comparisons that you implemented elsewhere? I think an inline note would be helpful and it would be cool if we could somehow do some degree of automation here so we don't have to do manual math on such a large number...

I'm not sure something like this is even possible and don't know exactly what implementation would look like if it is, but would be cool if we could just do manual math on each instance that fuzzy checker is called (like "check address on move", for instance) and then that could get compiled here.

alibow (approved):

zmbc:
Closing in favor of #333.

PR #257, opened 2023-04-03
Add statistical power calculation to fuzzy testing

zmbc:
### Description
- *Category*: test
- *Research reference*: none

### Changes and notes

Based on #256.

This adds a power calculation for each hypothesis: the range the underlying value would need to be outside for our test to be 80% powered to detect the difference. This has no effect on whether the tests pass or fail but adds context to the diagnostic output.

Obviously, the statistical power greatly depends on population size.

This is currently logging warnings when a hypothesis test is so underpowered that no result, no matter how extreme, would lead to rejecting the null hypothesis. There are a handful of these warnings even at our typical 250k population. I don't know if this will be annoying -- perhaps we should suppress them in specified places where we know our tests are underpowered and think that is okay.

### Verification and Testing

Ran integration tests.

With a slightly older but substantially similar version of the code, I also ran this for multiple population sizes and did [a cursory look at how the statistical power of our hypotheses relates to population size](https://github.com/ihmeuw/vivarium_research_prl/pull/29).

### DIFF
--- a/integration_tests/conftest.py
+++ b/integration_tests/conftest.py
@@ -1,4 +1,5 @@
 import os
+import warnings
 from pathlib import Path
 from typing import List, Optional, Tuple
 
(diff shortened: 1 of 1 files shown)


zmbc:

PR #260, opened 2023-04-03
Fuzzy test domestic migration rates

zmbc:
### Description
- *Category*: test
- *Research reference*: none

### Changes and notes

Based on #256.

Uses V&V target values [calculated on the research side](https://github.com/ihmeuw/vivarium_research_prl/pull/29).

### Verification and Testing

Ran integration tests with several population sizes.

Intentionally doubled the person migration rate and verified that these tests will fail (even with only 20k population size).


zmbc:

PR #261, opened 2023-04-04
Public sample data fixes

zmbc:
### Description
- *Category*: other
- *JIRA issue*: none
- *Research reference*: will be described in pseudopeople docs

### Changes and notes

No longer uses the us_population_size configuration value when calculating the ACS sample rate.
I think this configuration value is a misnomer; it is intended to capture the full size of a parallel run,
and therefore the number of employers that should be generated, since those are shared for the
entire parallel run.

I can also change the name of the configuration option, if desired.

Also, updates the dummy ZIP code to 00000, and makes employer addresses follow the "Anytown, US 00000" pattern.

### Verification and Testing

* Ran simulation with configuration values: population_size = 20_000, us_population_size = 20_000
* Ran post-processing with `--public-sample`
* Did manual V&V of the results: https://github.com/ihmeuw/vivarium_research_prl/pull/39

### DIFF
--- a/src/vivarium_census_prl_synth_pop/components/observers.py
+++ b/src/vivarium_census_prl_synth_pop/components/observers.py
@@ -246,7 +246,7 @@ class HouseholdSurveyObserver(BaseObserver):
             * builder.configuration.time.step_size
             / DAYS_PER_YEAR
             * builder.configuration.population.population_size
-            / builder.configuration.us_population_size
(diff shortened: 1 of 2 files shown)


rmudambi (approved):

rmudambi:
Instead of 99999, could we us 00000? Then we could use this data directly to make sure we're handling leading zeros correctly.

PR #262, opened 2023-04-05
containerize the output cols

stevebachmeier:
### Description
- *Category*: other/misc
- *JIRA issue*: na
- *Research reference*: na

### Changes and notes
This creates an output_schema.COLUMNS object for storing column names.

NOTE: I did NOT replace all column names throughout the repo, though I did
in the observers.py. The real impetus for this change is to import
column names into pseudopeople.

### Verification and Testing
Ran a small simulation to make sure nothing broke.


rmudambi:
We do NOT want to make this model repository a dependency of pseudopeople.

PR #263, opened 2023-04-07
Feature/categorize strings

rmudambi:
### Description

- *Category*: implementation
- *JIRA issue*: [MIC-3984](https://jira.ihme.washington.edu/browse/MIC-3984)
- *Research reference*: 

### Changes and notes
Converts all string columns to categorical to save both space and read time.
Removes support for csvs 

### Verification and Testing
Ran automated tests.
Ran a parallel simulation, made results both for a single seed (non-parallel) and in parallel using jobmon, and verified that results can be read by pseudopeople.

### DIFF
--- a/src/vivarium_census_prl_synth_pop/model_specifications/model_spec.yaml
+++ b/src/vivarium_census_prl_synth_pop/model_specifications/model_spec.yaml
@@ -46,19 +46,19 @@ configuration:
         population_size: 1_000_000
     # us_population_size: 250_000  # Override the default US population size
 
-    household_survey_observer_acs:
-        file_extension: "hdf"
-    household_survey_observer_cps:
-        file_extension: "hdf"
-    decennial_census_observer:
-        file_extension: "hdf"
-    wic_observer:
-        file_extension: "hdf"
-    social_security_observer:
-        file_extension: "hdf"
-    tax_w2_observer:
-        file_extension: "hdf"
-    tax_dependents_observer:
-        file_extension: "hdf"
-    tax_1040_observer:
-        file_extension: "hdf"
+#    household_survey_observer_acs:
+#        file_extension: "hdf"
+#    household_survey_observer_cps:
+#        file_extension: "hdf"
+#    decennial_census_observer:
+#        file_extension: "hdf"
+#    wic_observer:
+#        file_extension: "hdf"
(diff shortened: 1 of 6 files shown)


mattkappel (approved):

src/vivarium_census_prl_synth_pop/model_specifications/model_spec.yaml
```diff
@@ -46,19 +46,19 @@ configuration:
...
-        file_extension: "hdf"
-    social_security_observer:
-        file_extension: "hdf"
-    tax_w2_observer:
-        file_extension: "hdf"
-    tax_dependents_observer:
-        file_extension: "hdf"
-    tax_1040_observer:
-        file_extension: "hdf"
+#    household_survey_observer_acs:
```
stevebachmeier, line 49: delete?

src/vivarium_census_prl_synth_pop/tools/make_results.py
```diff
@@ -334,13 +335,7 @@ def perform_post_processing(
...
         obs_dir = build_output_dir(final_output_dir, subdir=observer)
         seed_ext = f"_{seed}" if seed != "" else ""
-        obs_data.to_hdf(
-            obs_dir / f"{observer}{seed_ext}.hdf",
-            "data",
-            format="table",
-            complib="bzip2",
-            complevel=9,
-        )
+        write_to_disk(obs_data.copy(), obs_dir / f"{observer}{seed_ext}.hdf")
```
stevebachmeier, line 338: Seems like the .copy() should be moved into `write_to_disk`.

stevebachmeier (approved):

src/vivarium_census_prl_synth_pop/tools/make_results.py
rmudambi: We don't need to copy before write to disk in the observers. The copy was just to avoid the `SettingWithCopyWarning` which doesn't apply there.

src/vivarium_census_prl_synth_pop/model_specifications/model_spec.yaml
rmudambi: I'm keeping this here so it's easier to add the parquet override.

PR #278, opened 2023-04-20
add sample data model spec

stevebachmeier:
### Description
- *Category*: feature
- *JIRA issue*: [MIC-4024](https://jira.ihme.washington.edu/browse/MIC-4024)
- *Research reference*: na

### Changes and notes
This adds a sample_data_model_spec.yaml for generating sample data.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -127,3 +127,6 @@ dmypy.json
 
 # Pycharm project settings
 .idea/
+
+# Model specification files
+src/vivarium_census_prl_synth_pop/model_specifications/*.yaml
--- a/.gitignore
+++ b/.gitignore
@@ -127,6 +127,3 @@ dmypy.json
 
 # Pycharm project settings
 .idea/
-
-# Model specification files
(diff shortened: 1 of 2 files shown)


.gitignore
```diff
@@ -127,6 +127,3 @@ dmypy.json
 
 # Pycharm project settings
 .idea/
-
-# Model specification files
-src/vivarium_census_prl_synth_pop/model_specifications/*.yaml
```
mattkappel, line 132: We don't want to change this. Custom model specs shouldn't get added. You shouldn't need this line to add the sample model spec.

stevebachmeier, line 132: I'm unsure if I should add this back once the new yaml gets merged. I'm assuming it's here just so that we're all free to have our own custom yaml files or something?

stevebachmeier: How do I add it w/o not ignoring it?

stevebachmeier: Ah git add --force. Well I cheated it by just undoing the gitignore.

rmudambi (approved):

PR #281, opened 2023-05-02
Nicknames proportion

albrja:
### Adds function to calculate nickname proportion constant to be used in pseudopeople for noise scaling.
- *Category*: Other
- *JIRA issue*: [MIC-3861](https://jira.ihme.washington.edu/browse/MIC-3861)
- *Research reference*: N/A

### Changes and notes
-Adds nicknames csv
-Adds function to calculate proportion of names that have a nickname

### Verification and Testing
N/A


src/vivarium_census_prl_synth_pop/data/nicknames_proportion.py
```diff
@@ -0,0 +1,25 @@
...
+
+def get_nicknames_proportion():
+    # This function calculates the proportion of names that have potential nicknames. This will return a float that
+    # that will be used as a constant to scale the nicknames noise function in the pseudopeople packages.
+
+    art = Artifact(paths.DEFAULT_ARTIFACT)
+    names = art.load(data_keys.SYNTHETIC_DATA.FIRST_NAMES).reset_index()["name"]
+    nicknames = _load_nicknames_data()
+
+    proportion_of_nicknames = len(nicknames.index) / len(names.unique())
```
rmudambi, line 24: You don't want unique names in the denominator. You want all names.

rmudambi (changes requested):

rmudambi (approved):

src/vivarium_census_prl_synth_pop/data/nicknames_proportion.py
```diff
@@ -0,0 +1,20 @@
...
+    # that will be used as a constant to scale the nicknames noise function in the pseudopeople packages.
+
+    art = Artifact(paths.DEFAULT_ARTIFACT)
+    names = art.load(data_keys.SYNTHETIC_DATA.FIRST_NAMES).reset_index()["name"]
+    # Load and format nicknames dataset
+    names_with_nicknames = pd.read_csv(paths.NICKNAMES_DATA_PATH)
+    names_with_nicknames = names_with_nicknames["name"].apply(lambda x: x.title())
+
+    # Proportion
+    nickname_eligible = names.isin(names_with_nicknames)
```
NathanielBlairStahn, line 18: Do you know that all the names in the artifact are title case?

albrja: Yes

rmudambi: Are you sure? There are a lot of African American names like DeAngelo, LaToya etc. that aren't title case. I suspect those names aren't in our nicknames csv though, so it probably doesn't matter.

src/vivarium_census_prl_synth_pop/data/nicknames_proportion.py
```diff
@@ -0,0 +1,20 @@
...
+
+from vivarium_census_prl_synth_pop.constants import data_keys, paths
+
+
+def get_nicknames_proportion():
+    # This function calculates the proportion of names that have potential nicknames. This will return a float that
+    # that will be used as a constant to scale the nicknames noise function in the pseudopeople packages.
+
+    art = Artifact(paths.DEFAULT_ARTIFACT)
+    names = art.load(data_keys.SYNTHETIC_DATA.FIRST_NAMES).reset_index()["name"]
```
zmbc, line 12: This ignores the "freq" column which is what makes some names more common than others. I think we want to correct this.

src/vivarium_census_prl_synth_pop/data/nicknames_proportion.py
NathanielBlairStahn: Yeah, I was assuming we would do a case-insensitive match to determine whether a name is eligible for a nickname, e.g. convert all the simulation names to lower case to match the nicknames CSV. And yeah, converting all the chosen nicknames to title case seems reasonable.

src/vivarium_census_prl_synth_pop/data/nicknames_proportion.py
```diff
@@ -1,15 +1,26 @@
...
 
-    art = Artifact(paths.DEFAULT_ARTIFACT)
-    names = art.load(data_keys.SYNTHETIC_DATA.FIRST_NAMES).reset_index()["name"]
+    data_files = []
+    for file in data_dir.glob("*.parquet"):
+        data_files.append(file)
+    names = []
+    for file in data_files:
+        df = pd.read_parquet(file)
+        df = df.loc[df["year"] == 2030, "first_name"]
```
rmudambi, line 20: Don't worry about changing this, but if we re-run this, we should use the strategy to load only the 2030 year from the file that we are already using to load in our interface functions.

rmudambi (approved):

PR #286, opened 2023-05-10
Feature/sbachmei/mic 4067 implement vivarium v1.1.0

stevebachmeier:
### Description
- *Category*: other
- *JIRA issue*: [MIC-4067](https://jira.ihme.washington.edu/browse/MIC-4067)
- *Research reference*: na

### Changes and notes
Primary changes:
- We lost randomness::random and so I had to implement that functionality
  in `vectorized_choice`
- No more randomness Array either.
- Instead of side-instantiating a randomness stream for business moves, we
  are now generating one properly through builder. To maintain a crn strictly
  for business move dynamics, we overwrite the seed in-code.
- We started getting errors because simulants were not being registered to the
  randomness stream which resulted in the IndexMap not properly updating. We
  chose to solve this be removing the key_columns from the config files (and just
  deleting the unused self.register_updates). This works because if key_columns
  is empty the IndexMap gets instantiated w/ `_use_crn == False` which in turn
  has the IndexMap getter return index.values rather than a map. It's confusing.

### Verification and Testing
pytests pass.

I also confirmed that the manual overwrite of random_seed is working by
running two simulations in interactive mode with different model-spec seeds
and checking that the business details are identical.

### DIFF
--- a/integration_tests/test_domestic_migration.py
+++ b/integration_tests/test_domestic_migration.py
@@ -314,28 +314,23 @@ def test_addresses_during_moves(
             continue
 
         total_num_moved += mask_moved_units.sum()
-        # Check that the number of moved units that have the same details is very low
-        # NOTE: we cannot assert that all moved units have different details
-        # because they are free to move within the same state or even PUMA
-        assert (
-            (
-                before_units.loc[mask_moved_units, address_cols]
-                == after_units.loc[mask_moved_units, address_cols]
-            )
-            .all(axis=1)
-            .mean()
-        ) < 0.001  # TODO: pick a smarter number?
-        # address details do not change if address_id does not change
+
+        # Check that address details do not change if address_id does not change
         pd.testing.assert_frame_equal(
             before_units[~mask_moved_units], after_units[~mask_moved_units]
         )
 
         # Check that most movers are in a new state-PUMA combination
+        # NOTE: we cannot assert that all moved units have different details
+        # because they are free to move within the same state and PUMA. The
+        # threshold applied below is a somewhat arbitrarily-chosen small number
+        # since PUMAs are different sizes and so it is difficult to estimate
+        # the chance of a simulant moving into the same one.
         if mask_moved_units.sum() > 1:
             assert (
                 before_units.loc[mask_moved_units, [state_id_col, puma_col]]
                 == after_units.loc[mask_moved_units, [state_id_col, puma_col]]
-            ).all(axis=1).mean() < 100 * 1 / len(states_pumas)
+            ).all(axis=1).mean() < 0.01
 
     # Check that at least some units moved during the sim
     assert total_num_moved > 0
--- a/src/vivarium_census_prl_synth_pop/components/population.py
+++ b/src/vivarium_census_prl_synth_pop/components/population.py
@@ -40,9 +40,7 @@ class Population:
     def setup(self, builder: Builder):
         self.config = builder.configuration.population
         self.seed = builder.configuration.randomness.random_seed
-        self.randomness = builder.randomness.get_stream(
-            "household_sampling", initializes_crn_attributes=True
-        )
+        self.randomness = builder.randomness.get_stream("household_sampling")
         self.proportion_with_ssn = builder.lookup.build_table(
             data=data_values.PROPORTION_INITIALIZATION_WITH_SSN
         )
--- a/src/vivarium_census_prl_synth_pop/components/population.py
+++ b/src/vivarium_census_prl_synth_pop/components/population.py
@@ -41,7 +41,7 @@ class Population:
         self.config = builder.configuration.population
         self.seed = builder.configuration.randomness.random_seed
         self.randomness = builder.randomness.get_stream(
-            "household_sampling", for_initialization=True
+            "household_sampling", initializes_crn_attributes=True
         )
         self.proportion_with_ssn = builder.lookup.build_table(
             data=data_values.PROPORTION_INITIALIZATION_WITH_SSN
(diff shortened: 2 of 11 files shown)


src/vivarium_census_prl_synth_pop/tools/make_results.py
```diff
@@ -274,6 +275,7 @@ def perform_post_processing(
         key="post_processing_maps",
         clock=lambda: pd.Timestamp("2020-04-01"),
         seed=0,
+        index_map=IndexMap(),
```
albrja, line 278: Don't we need to increase this above the default 1_000_000?  I'm assuming this would have broken if we needed to since we already process files with more than a million rows.

src/vivarium_census_prl_synth_pop/components/population.py
```diff
@@ -41,7 +41,7 @@ def setup(self, builder: Builder):
         self.config = builder.configuration.population
         self.seed = builder.configuration.randomness.random_seed
         self.randomness = builder.randomness.get_stream(
-            "household_sampling", for_initialization=True
+            "household_sampling", initializes_crn_attributes=True
```
rmudambi, line 44: We don't have CRN in this simulation, so `initializes_crn_attributes` should be (the default) False.

src/vivarium_census_prl_synth_pop/model_specifications/model_spec.yaml
```diff
@@ -30,7 +30,7 @@ configuration:
         extrapolate: True
     randomness:
         map_size: 1_000_000
-        key_columns: ['entrance_time', 'age']
+        key_columns: []  # We do not utilize crn
```
albrja, line 33: Is this because we throw everything into vectorized choice and not the normal functionality of CRN?

albrja (approved):

src/vivarium_census_prl_synth_pop/tools/make_results.py
rmudambi: Good question. Did you test post-processing more than 1 million rows? I *think* because there are no key columns, use_crn will be false on the IndexMap, so we shouldn't get an error.

src/vivarium_census_prl_synth_pop/utilities.py
```diff
@@ -194,8 +194,12 @@ def vectorized_choice(
...
     if randomness_stream is None:
         # Generate an additional_key on-the-fly and use that in randomness.random
-        additional_key = f"{additional_key}_{random_seed}"
-        probs = random(str(additional_key), index)
+        # additional_key = f"{additional_key}_{random_seed}"
+        # probs = random(str(additional_key), index)
+        random_state = np.random.RandomState(seed=get_hash(f"{additional_key}_{random_seed}"))
+        sample_size = index.max() + 1
+        raw_draws = random_state.random_sample(sample_size)
+        probs = pd.Series(raw_draws[index], index=index)
```
rmudambi, line 202: If the index is something like `pd.Index([1_000_000, 1_000_001])` we will do 1,000,002 samples to get two numbers we'll use. This seems quite wasteful. Why not do this?
```
raw_draws = random_state.random_sample(len(index))
probs = pd.Series(raw_draws, index=index)
```

src/vivarium_census_prl_synth_pop/tools/make_results.py
stevebachmeier: Hmm, good point - I def forgot to test `make_results`. I think it won't matter here for the same reason it didn't matter in pseudopeople...which I still don't entirely understand

rmudambi (commented):
Some minor comments and echoing Jim's question.

src/vivarium_census_prl_synth_pop/model_specifications/model_spec.yaml
stevebachmeier: No, it boils down to the INdexMap getter function https://github.com/ihmeuw/vivarium/blob/main/src/vivarium/framework/randomness/index_map.py#L243-L247.

I'm still trying to wrap my head around it, but basically if you don't properly register new simulants to the randomness stream (which we were not) and you have `key_columns`, then this getter tries to return self._map.loc[index].values except we don't have a map (b/c that's part of registering new simulants).

So we could have just registered new simulants to the randomness stream every time we create them, but that's unnecessary b/c we don't actually use crn. So instead, we "cheated" by just removing the `key_columns`.

src/vivarium_census_prl_synth_pop/utilities.py
stevebachmeier: That makes sense. I did just copy this from vivarium - though I think in that case it works b/c you're dealing with a state table with a range index?

src/vivarium_census_prl_synth_pop/components/population.py
stevebachmeier: @rmudambi Interesting, changing this actually causes the `test_addresses_during_moves` integration test to fail.

integration_tests/test_domestic_migration.py
```diff
@@ -314,28 +314,23 @@ def test_addresses_during_moves(
             continue
 
         total_num_moved += mask_moved_units.sum()
-        # Check that the number of moved units that have the same details is very low
```
stevebachmeier, line 317: This test is useless because `address_cols` includes address_id but address_id is what was used to define `mask_moved_units`, i.e. it is always 0 < 0.001.

albrja (approved):

rmudambi (approved):

PR #287, opened 2023-05-11
output 'household_id' for all observers

stevebachmeier:
### Description
- *Category*: observers
- *JIRA issue*: [MIC-4036](https://jira.ihme.washington.edu/browse/MIC-4036)
- *Research reference*: https://github.com/ihmeuw/vivarium_research/pull/1164/files

### Changes and notes

This adds `household_id` to the output columns for all observers.

### Verification and Testing
integration tests pass

### DIFF
--- a/src/vivarium_census_prl_synth_pop/components/observers.py
+++ b/src/vivarium_census_prl_synth_pop/components/observers.py
@@ -31,8 +31,10 @@ class BaseObserver(ABC):
         "race_ethnicity",
         "guardian_1",
         "guardian_2",
+        "household_id",
     ]
     DEFAULT_OUTPUT_COLUMNS = [
+        "household_id",  # Included in every dataset as a source of truth
         "first_name_id",
         "middle_name_id",
         "last_name_id",
@@ -48,7 +50,6 @@ class BaseObserver(ABC):
         "guardian_2",
         "guardian_1_address_id",
         "guardian_2_address_id",
-        # todo: add necessary guardian address columns
     ]
 
     configuration_defaults = {"observer": {"file_extension": "hdf"}}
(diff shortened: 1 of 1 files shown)


mattkappel (approved):

src/vivarium_census_prl_synth_pop/components/observers.py
```diff
@@ -497,16 +494,18 @@ class SocialSecurityObserver(BaseObserver):
...
         "first_name_id",
         "middle_name_id",
         "last_name_id",
         "date_of_birth",
-        "event_type",
-        "event_date",
         "sex",
         "race_ethnicity",
+        "event_type",
+        "event_date",
```
rmudambi, line 505: Is there a reason for the reordering?

src/vivarium_census_prl_synth_pop/components/observers.py
```diff
@@ -549,21 +548,22 @@ def get_observation(self, event: Event) -> pd.DataFrame:
             event.index,
             query="has_ssn == True",  # only include simulants with a SSN
         )
-        df_creation = pop.filter(self.POST_PROCESSING_FIRST_NAME_METADATA_COLS)
+
+        df_creation = pop
```
rmudambi, line 552: Why are you assigning the variable like this? Do you mean to do `df_creation = pop.copy()`?

rmudambi (approved):

src/vivarium_census_prl_synth_pop/components/observers.py
stevebachmeier: Oh, I forgot to delete the now-unued POST_PROCESSING_FIRST_NAME_METADATA_COLS which I think will make it more clear; see the new commit. I basically just re-ordered to match that.

src/vivarium_census_prl_synth_pop/components/observers.py
stevebachmeier: Oh, I just deleted the filter. In this case is there  a difference whether I .copy() or not? We don't below for df_death but I still get confused sometimes as to when we need to copy or not

rmudambi: Copying means you create a new dataframe in a new place in memory. What you are doing is just assigning a new variable to point to the same location in memory. `df_creation` and `pop` are both pointing at the same dataframe, so the changes you are making to `df_creation` below are also being made to `pop`. This assignment doesn't really do anything. `pop.filter` like most pandas operations is not an in-place operation by default, so it is implicitly doing a copy.

PR #288, opened 2023-05-12
include household_id in post-processing

stevebachmeier:


rmudambi (approved):

src/vivarium_census_prl_synth_pop/tools/make_results.py
```diff
@@ -205,7 +210,6 @@
         "mailing_address_po_box",
         "sex",
         "ssn",
-        "tax_year",
```
stevebachmeier, line 208: This was duplicated (see three rows down)

PR #290, opened 2023-05-18
Update directory name

albrja:
### Updates directory name for final results and adds optional version argument to make_results.
- *Category*: Feature
- *JIRA issue*: [MIC-4029](https://jira.ihme.washington.edu/browse/MIC-4029)
- *Research reference*: 

### Changes and notes
-adds additional version flag for cli command make_results
-updates directory name and incorporates version into directory name if one is provided.

### Verification and Testing
Successfully ran make_results and make_state_results and saw desired directory names.


src/vivarium_census_prl_synth_pop/tools/cli.py
```diff
@@ -150,6 +150,13 @@ def make_artifacts(
     "queue. The maximum supported runtime is 3 days. Keep in mind that "
     "runtimes by node vary wildly. ",
 )
+@click.option(
```
stevebachmeier, line 153: Move this arg to right after --output-dir.
(1) it makes sense to be next to the output directory stuff
(2) it should definitely come before the --jobmon arg

src/vivarium_census_prl_synth_pop/tools/cli.py
```diff
@@ -150,6 +150,13 @@ def make_artifacts(
...
+    "--version",
+    type=str,
+    default=None,
+    help="Provide a version number for final results. If None is provided, a datetime "
```
stevebachmeier, line 157: nit: show_default=True,

src/vivarium_census_prl_synth_pop/tools/cli.py
albrja: That makes sense.

src/vivarium_census_prl_synth_pop/tools/cli.py
```diff
@@ -62,6 +62,14 @@ def make_artifacts(
 
 @click.command()
 @click.argument("output_dir", type=click.Path(exists=True))
+@click.option(
+    "--version",
+    type=str,
+    default=None,
+    show_default=True,
+    help="Provide a version number for final results. If None is provided, a datetime "
```
rmudambi, line 70: We aren't generating a datetime if a version is not provided. This help text should be updated.

src/vivarium_census_prl_synth_pop/tools/make_results.py
```diff
@@ -488,8 +488,21 @@ def subset_results_by_state(processed_results_dir: str, state: str) -> None:
     abbrev_name_dict = {v: k for k, v in metadata.US_STATE_ABBRV_MAP.items()}
     state_name = sanitize_location(abbrev_name_dict[state.upper()])
     processed_results_dir = Path(processed_results_dir)
-    usa_results_dir = processed_results_dir / "usa"
-    state_dir = processed_results_dir / "states" / state_name
+    dirs = [x for x in processed_results_dir.iterdir() if x.is_dir()]
```
rmudambi, line 491: Nit: don't abbreviate unnecessarily. Call these variables `directories` and `d` should become `directory`

src/vivarium_census_prl_synth_pop/tools/cli.py
stevebachmeier, line 70: Oh, probably want to specify in this help string the format of version, e.g.
```
help="Provide a version number for final results. Must be formatted like '#.#.#'. If None is provided..."
```

And then throw an error if they provide a different version formatting.

src/vivarium_census_prl_synth_pop/utilities.py
```diff
@@ -288,12 +288,25 @@ def build_output_dir(output_dir: Path, subdir: Optional[Union[str, Path]] = None
     return output_dir
 
 
-def build_final_results_directory(results_dir: str) -> Tuple[Path, Path]:
+def build_final_results_directory(
+    results_dir: str,
+    version: Optional[str] = None,
+) -> Tuple[Path, Path]:
+    if version is None:
```
stevebachmeier, line 295: I think you should add a check that this version doesn't already exist and, if so, raise an error with an informative message.

src/vivarium_census_prl_synth_pop/utilities.py
```diff
@@ -288,12 +288,25 @@ def build_output_dir(output_dir: Path, subdir: Optional[Union[str, Path]] = None
...
+        subdir = (
+            paths.FINAL_RESULTS_DIR_NAME
+            / datetime.now().strftime("%Y_%m_%d_%H_%M_%S")
+            / "pseudopeople_input_data_usa"
+        )
+    else:
+        subdir = (
+            paths.FINAL_RESULTS_DIR_NAME
+            / datetime.now().strftime("%Y_%m_%d_%H_%M_%S")
```
stevebachmeier, line 304: @rmudambi Do we want the version number to replace this timestamp altogether? It would make it easier to find (at least until we fix the symlink generation).

src/vivarium_census_prl_synth_pop/utilities.py
```diff
@@ -288,12 +288,25 @@ def build_output_dir(output_dir: Path, subdir: Optional[Union[str, Path]] = None
...
+            / f"pseudopeople_input_data_usa_{version}"
+        )
```
rmudambi, line 306: There's a lot of duplicated code here. `paths.FINAL_RESULTS_DIR_NAME / datetime.now().strftime("%Y_%m_%d_%H_%M_%S")` is the same in both branches

src/vivarium_census_prl_synth_pop/tools/make_results.py
```diff
@@ -488,8 +488,21 @@ def subset_results_by_state(processed_results_dir: str, state: str) -> None:
...
+    # Get USA results directory
+    for d in dirs:
```
stevebachmeier, line 493: I think this all makes the assuption that there is only one directory in this folder, correct? That may  not be the case for various reasons.

This may be tougher than I thought.

To get the version number you'll need to sort dirnames and pick the last one (b/c 1.0.1 > 1.0.0).

But I'm not sure offhand how that will work for runs done without --version.

stevebachmeier (changes requested):
I think it would be useful to fully test the outputs by doing something like the following.

1. run a small sim
2. `make_results` on the sim (no --version)
3. `make_results --version 1.0.0`
4. `make_results` (no --version)
5. `make_results --version 1.0.1`

And make sure it works. Would be useful if you `tree` the results here so we can see them as well.

A few extra things to consider:

- Confirm whether we want the version to be 1.1.0 or 1_1_0. Probably the latter since that's what the zip file ends up being anyway. Maybe the code should be flexible for the user (us) to pass in `--version 1.1.0` or `--version 1_1_0` and then it changes it to underscores?
- Do we want to ensure that the version passed in is greater than any existing ones?

src/vivarium_census_prl_synth_pop/tools/cli.py
stevebachmeier: Yeah that was my bad to suggest that. I didn't realize we were going to keep these results in datestamped directories.

src/vivarium_census_prl_synth_pop/tools/make_results.py
albrja: There will only be one "USA" directory in here as it is timestamp when you run make results.

src/vivarium_census_prl_synth_pop/utilities.py
albrja: See other comment but this is making a new directory.

src/vivarium_census_prl_synth_pop/utilities.py
albrja: We are leaving it since we will update the symlink generation at some point.

t
```diff
@@ -0,0 +1,31 @@
+[1mdiff --git a/src/vivarium_census_prl_synth_pop/constants/data_values.py b/src/vivarium_census_prl_synth_pop/constants/data_values.py[m
+[1mindex 82e883f..9211c28 100644[m
+[1m--- a/src/vivarium_census_prl_synth_pop/constants/data_values.py[m
+[1m+++ b/src/vivarium_census_prl_synth_pop/constants/data_values.py[m
+[36m@@ -96,8 +96,8 @@[m [mYEARLY_JOB_CHANGE_RATE = 0.5  # 50 changes per 100 py[m
+ [m
+ BUSINESS_MOVE_RATE_YEARLY = 0.1  # 10 changes per 100 py[m
+ [m
+[31m-PERSONAL_INCOME_PROPENSITY_DISTRIBUTION = stats.norm(loc=0.0, scale=0.812309 ** 0.5)[m
+[31m-EMPLOYER_INCOME_PROPENSITY_DISTRIBUTION = stats.norm(loc=0.0, scale=0.187691 ** 0.5)[m
```
stevebachmeier, line 10: delete this file

stevebachmeier (approved):
I still don't love that we're hiding versioned files within date-stamped files (which is already a versioning method) but looks good assuming it's what we really want

t
albrja: We don't need this anymore?

stevebachmeier: Isn't it a junk file? The filename is just "t" and it appears to be a `git diff`

PR #292, opened 2023-05-30
Copy from household member

albrja:
### Adds copy from household member functionality to observers.
- *Category*: Feature
- *JIRA issue*: [MIC-4042](https://jira.ihme.washington.edu/browse/MIC-4042)
- *Research reference*: https://vivarium-research.readthedocs.io/en/latest/models/concept_models/vivarium_census_synthdata/concept_model.html#id115

###
-Adds copy from household member function to utilities
-Implements copying of necessary columns to all observers
-adds related tests

### Verification and Testing
Successfully ran simulation and all tests pass.


.gitignore
```diff
@@ -130,3 +130,6 @@ dmypy.json
 
 # Model specification files
 src/vivarium_census_prl_synth_pop/model_specifications/*.yaml
+
+# VSCode test cache
+**/.vscode/
```
stevebachmeier, line 135: This is odd since I've been using vscode forever and never needed to add this

src/vivarium_census_prl_synth_pop/components/observers.py
```diff
@@ -1016,6 +1057,7 @@ def get_observation(self, event: Event) -> pd.DataFrame:
             ],
             additional_key="joint_filing_1040",
         )
+        # todo: add function here to copy each of the necessary columns
```
stevebachmeier, line 1060: I don't understand this todo. isn't that what copy_from_household_member does?

src/vivarium_census_prl_synth_pop/results_processing/formatter.py
```diff
@@ -116,6 +132,8 @@ def format_age(data: pd.DataFrame) -> pd.Series:
     "guardian_id": (get_guardian_id, ["guardian_id", "random_seed"]),
     "ssn_id": (format_ssn_id, ["ssn_id", "random_seed"]),
     "age": (format_age, ["age"]),
+    "copy_age": (format_copy_age, ["copy_age"]),
```
stevebachmeier, line 135: There's no copy_dob - does this mean it's already in the correct format?

src/vivarium_census_prl_synth_pop/results_processing/formatter.py
```diff
@@ -92,6 +92,22 @@ def format_age(data: pd.DataFrame) -> pd.Series:
     return data["age"].astype(int)
 
 
+def format_copy_age(data: pd.DataFrame) -> pd.Series:
+    no_copy_idx = data.index[data["copy_age"].isna()]
```
stevebachmeier, line 96: Can't you just do the same thing as format_age? `return data["copy_age"].astype(int)`. This way you don't need to convert to string and split.

src/vivarium_census_prl_synth_pop/results_processing/formatter.py
```diff
@@ -92,6 +92,22 @@ def format_age(data: pd.DataFrame) -> pd.Series:
...
+    column = data["copy_age"].astype(str)
+    column = column.apply(lambda row: row.split(".")[0])
+    column.loc[no_copy_idx] = np.nan
```
stevebachmeier, line 99: Isn't this already NaN by definition of `no_copy_idx`?

src/vivarium_census_prl_synth_pop/results_processing/formatter.py
```diff
@@ -92,6 +92,22 @@ def format_age(data: pd.DataFrame) -> pd.Series:
...
+    return column
+
+
+def format_copy_ssn(data: pd.DataFrame) -> pd.Series:
+    no_copy_idx = data.index[data["copy_ssn"].isna()]
```
stevebachmeier, line 104: Same general question as the copy age formatting - why is this doing different things than the original `format_snn_id` function?

src/vivarium_census_prl_synth_pop/results_processing/formatter.py
```diff
@@ -92,6 +92,22 @@ def format_age(data: pd.DataFrame) -> pd.Series:
...
+    column = data["random_seed"].astype(str) + "_" + data["copy_ssn"].astype(str)
+    column = column.apply(lambda row: row[:-2])
```
stevebachmeier, line 106: Wht are the last two characters that you're not including?

src/vivarium_census_prl_synth_pop/results_processing/ssn_and_itin.py
```diff
@@ -218,3 +218,11 @@ def do_collide_ssns(
         randomness=randomness,
     )
     return obs_data
+
+
+def copy_ssn_from_household_member(
```
stevebachmeier, line 223: I'd move this into make_results.py since it's only called there. And then if we were to start calling it elsewhere I'd move into utilities.py

src/vivarium_census_prl_synth_pop/tools/cli.py
```diff
@@ -162,7 +162,7 @@ def make_artifacts(
 )
 def make_results(
     output_dir: str,
-    version: str,
+    label_version: str,
```
stevebachmeier, line 165: Why this change?

src/vivarium_census_prl_synth_pop/tools/make_results.py
```diff
@@ -348,6 +368,16 @@ def perform_post_processing(
             # For w2, we need to post-process to allow for SSN collisions in the data in cases where
             #  the simulant has no SSN but is employed (they'd need to have supplied an SSN to their employer)
             obs_data = do_collide_ssns(obs_data, maps["simulant_id"]["ssn"], randomness)
+        if observer in [
```
stevebachmeier, line 371: Why does this need to be outside of the main loop in lines 359-365?

t
```diff
@@ -0,0 +1,31 @@
+[1mdiff --git a/src/vivarium_census_prl_synth_pop/constants/data_values.py b/src/vivarium_census_prl_synth_pop/constants/data_values.py[m
```
stevebachmeier, line 1: Delete this file. I bet you have an alias that got jumbled since this is exactly the same thing that happened in your last PR (a diff is generated and saved to a file called `t`)

src/vivarium_census_prl_synth_pop/utilities.py
```diff
@@ -500,3 +500,47 @@ def write_to_disk(data: pd.DataFrame, path: Path):
...
+    copy_cols = metadata.COPY_HOUSEHOLD_MEMBER_COLS
+    for col in [column for column in copy_cols.keys() if column in pop.columns]:
+        pop[copy_cols[col]] = np.nan
+        if col == "has_ssn":
+            # Subset to rows that have SSN so we do not try and map nans.
+            households = (
+                pop[pop[col]].groupby("household_id").apply(lambda df_g: list(df_g.index))
+            )
+        else:
+            # This makes the assumption age and date of birth will not have nans at this poitn
```
stevebachmeier, line 520: Is this actually an assumption at this point? Do you have reason to believe this isn't the case? I honestly don't remember

src/vivarium_census_prl_synth_pop/utilities.py
```diff
@@ -500,3 +500,47 @@ def write_to_disk(data: pd.DataFrame, path: Path):
...
+            households = pop.groupby("household_id").apply(lambda df_g: list(df_g.index))
+        # Drop GQ and households with single member
+        households = households[households.apply(len) > 1]
+        households = households.loc[
+            households.index.difference(data_values.GQ_HOUSING_TYPE_MAP.keys())
+        ]
+        simulants_and_household_members = pop["household_id"].map(households).dropna()
```
stevebachmeier, line 527: Do you need the .dropna()? Wouldn't the closest to an NaN be an empty list from households?

tests/test_synthetic_pii.py
```diff
@@ -731,3 +732,72 @@ def test_id_uniqueness(artifact, hdf_key, num_need_ids, random_seed):
...
     else:
         assert len(np.unique(ids)) == num_need_ids
+
+
+def test_copy_household_member():
+    # Dummy dataset
+    # Note this function requires columns "household_id" and will copy from "age", "date_of_birth", and
+    # "ssn" if that column is present
+    hh_ids = [0, 0, 21, 21, 21, 21, 12, 12, 13, 13, 14, 14, 14, 14, 14, 25, 25, 25, 26, 27]
+    ages = [
```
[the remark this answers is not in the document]
stevebachmeier: How about `ages = list(range(10,30))`

tests/test_synthetic_pii.py
```diff
@@ -731,3 +732,72 @@ def test_id_uniqueness(artifact, hdf_key, num_need_ids, random_seed):
...
+    expected_no_copy_idx = pop.index[
+        pop["household_id"].isin(
+            list(data_values.GQ_HOUSING_TYPE_MAP.keys()) + [12, 13, 25, 26, 27]
+        )
+    ]
+    assert no_copy_idx.equals(expected_no_copy_idx)
+    # SSNs are unique so we cannot copy the original value like we can with age and date of birth so
+    # we can check that there are no matches
+    have_ssn_copy = copy.index.difference(no_copy_idx)
+    assert (copy.loc[have_ssn_copy, "ssn"] != copy.loc[have_ssn_copy, "copy_ssn"]).all()
```
stevebachmeier, line 803: Do you think it's worth adding a test to ensure you are uniformly choosing from available values?

.gitignore
albrja: If I use the the vscode debugger it adds this folder and I am lazy and didnt want to remove it every time.

src/vivarium_census_prl_synth_pop/components/observers.py
```diff
@@ -752,6 +782,7 @@ def get_observation(self, event: Event) -> pd.DataFrame:
             pop[pop["has_ssn"]].groupby("address_id").apply(lambda df_g: list(df_g.index))
         )
         household_members_w_ssn = simulants_wo_ssn.map(household_members_w_ssn).dropna()
+        # todo: Fix me MIC-4092
```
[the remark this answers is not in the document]
albrja: I talked to Abie and am not going to make a change to this but essentially we initialize a np rng generator here and use vivarium random seed so the np generator is the same on each time step because the seed doesn't change. For comparison in the utils function I wrote I have the seed I create from _key. He also confirmed to leave both approaches and hopes RT will determine at some point which approach is better but for now both implementations are acceptable. I will remove the TODO.

src/vivarium_census_prl_synth_pop/components/observers.py
albrja: I forgot to remove it.

src/vivarium_census_prl_synth_pop/results_processing/formatter.py
albrja: See other comment.

src/vivarium_census_prl_synth_pop/results_processing/formatter.py
albrja: See other comment.

src/vivarium_census_prl_synth_pop/results_processing/formatter.py
albrja: See other comment.

src/vivarium_census_prl_synth_pop/results_processing/formatter.py
albrja: ".0". Because of nans, the simulant ids in copy_ssn are 1234.0 instead of 1234.

src/vivarium_census_prl_synth_pop/tools/cli.py
albrja: Not sure why this is showing up. This is from a bugfix that I rebased on

src/vivarium_census_prl_synth_pop/tools/make_results.py
albrja: In the main loop we are generating and then applying all our maps. We need the full ssn map to be created before we can map them to "copy_ssn". It is very similar logic to the ssn collision right above, we need to post-process the ssn map which we just can do inside the main loop.

src/vivarium_census_prl_synth_pop/utilities.py
albrja: This comment is basically stating that since this our population state table, age and date of birth should not have nans as we do not have anything in the simulation that would create nans in those columns. It is ultimately a reminder that if for some reason we change that in the future we would need to subset our groupby object like we do in 517.

src/vivarium_census_prl_synth_pop/utilities.py
albrja: I believe you are right. We also drop lists that are less than length 2 so I think this is unnecessary but the idea was to prevent tryign to map nans.

tests/test_synthetic_pii.py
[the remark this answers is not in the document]
albrja: Yes

tests/test_synthetic_pii.py
```diff
@@ -731,3 +732,72 @@ def test_id_uniqueness(artifact, hdf_key, num_need_ids, random_seed):
...
+        # Check expected missing rows based on household for age and date of birth
+        no_copy_idx = copy.index[copy[copy_cols[col]].isna()]
+        # Get household ids that are in GQ or only have a single occupant
+        expected_no_copy_idx = pop.index[
+            pop["household_id"].isin(list(data_values.GQ_HOUSING_TYPE_MAP.keys()) + [26, 27])
+        ]
+        assert no_copy_idx.equals(expected_no_copy_idx)
+
+    # Check expected missing rows for ssns. Note: These will be the same household ids along
+    # with household ids 12, 13 and 25 because those households only have 1 member with a SSN
```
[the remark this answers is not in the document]
albrja: Yes the comment is unclear. Its the the "same households" in the previous check + 12, 13, 25. Will update comment.

tests/test_synthetic_pii.py
albrja: I think its a good idea but I'm not sure how useful or accurate we could make it since we have so many 2-3 person households.

.gitignore
stevebachmeier: I mean, I don't blame you but I haven't needed to do it. weird.

stevebachmeier (approved):

PR #293, opened 2023-06-07
Bugfix/empty series dtype change

albrja:
### Fixes bug that where applying a lambda function on an empty series returns the dataframe instead of a series.
- *Category*: Bugfix
- *JIRA issue*: [MIC-4132](https://jira.ihme.washington.edu/browse/MIC-4132)
- *Research reference*: 

### Changes and notes
-handles empty dataframe in utils function

### Verification and Testing
Successfully ran simulation.

### DIFF
--- a/src/vivarium_census_prl_synth_pop/utilities.py
+++ b/src/vivarium_census_prl_synth_pop/utilities.py
@@ -521,32 +521,32 @@ def copy_from_household_member(
             households = pop.groupby("household_id").apply(lambda df_g: list(df_g.index))
         # Drop GQ and households with single member
         households = households[households.apply(len) > 1]
-        if households.empty:
+        if not households.empty:
+            households = households.loc[
+                households.index.difference(data_values.GQ_HOUSING_TYPE_MAP.keys())
+            ]
+            simulants_and_household_members = pop["household_id"].map(households).dropna()
+            # Note the following call results in a dataframe if 'simulants_and_household_members' is empty
+            # and is handled above
+            simulant_ids_to_copy = simulants_and_household_members.reset_index().apply(
+                lambda row: [
+                    household
+                    for household in row.loc["household_id"]
+                    if household != row.loc["index"]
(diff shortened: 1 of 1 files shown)


src/vivarium_census_prl_synth_pop/utilities.py
```diff
@@ -524,23 +524,29 @@ def copy_from_household_member(
...
+            copy_ids = simulant_ids_to_copy.map(np.random.default_rng(seed).choice)
+            if col == "has_ssn":
+                pop.loc[copy_ids.index, copy_cols[col]] = copy_ids
+            else:
+                pop.loc[copy_ids.index, copy_cols[col]] = copy_ids.map(
+                    pop.loc[copy_ids.index, col]
+                )
+        else:
+            # Save as object type - current pandas defaults to dtype float with future warning
+            pop[copy_cols[col]] = pd.Series(np.nan, index=pop.index, dtype=object)
```
mattkappel, line 550: So this else block is the new. I'm wondering if there's a good way to avoid all these indents. Not sure it's an improvement (i.e., take it or leave it imho), but what about

```
if households.empty:
    pop[copy_cols[col]] = pd.Series(np.nan, index=pop.index, dtype=object)
    continue
# lines 528-547 here
```

mattkappel (approved):

src/vivarium_census_prl_synth_pop/utilities.py
albrja: I had tried something very similar to this with continue to break out of the loop for that column and I couldn't get it to work but I will keep trying. It seems like it should and I agree it is a lot cleaner.

albrja: @mattkappel  if I use continue I still hit line 541 which is the original cause of the bugfix. I thought using continue would send me out of the for loop for the iteration and into it for the next iteration.

PR #303, opened 2023-07-18
change 'income' to 'wages' in w2/1099 observer

stevebachmeier:
### Description

- *Category*:  misc
- *JIRA issue*: [MIC-4199](https://jira.ihme.washington.edu/browse/MIC-4199)
- *Research reference*:  https://vivarium-research.readthedocs.io/en/latest/models/concept_models/vivarium_census_synthdata/concept_model.html#w2-and-1099-forms

### Changes and notes
 
This just changes the "income" column to "wages" in the w2/1099 observer.

NOTE: It does NOT change the column name in the state table. This seemed like
the right thing to do but that does need to be kept in mind when, say, 
writing new tests that work on the state table.

### Verification and Testing

pytests pass and I ran a small sim and made results to ensure the column is there.


rmudambi (commented):
Any reason not to change the income pipeline/column name? Seems like that would reduce confusion down the line.

stevebachmeier:
> Any reason not to change the income pipeline/column name? Seems like that would reduce confusion down the line.

I thought about it and I'm happy to, but I saw that it was also referred to as "income" in WIC. That said, WIC also doesn't end up with an income or wages column. 

But yeah, maybe that's the key and it should just wholesale be referred to as "wages"

rmudambi (approved):

PR #304, opened 2023-07-20
Revert most income-wage changes

stevebachmeier:
### Description

- *Category*:  hotfix
- *JIRA issue*: na
- *Research reference*:  na

### Changes and notes
 
Income is different than wages (wages include only money from employers
while income includes many other streams, e.g. capital gains, child
support, etc.

We decided that we really should only change the w2 "income" column
to "wages" and keep basically everything else as income, i.e. the
pipeline, propensity columns, docs, etc.

This reverts commit 8d3c2d166e5c9ad16b82dfc2d9d417bee3305f19 and 
then applies the minor changes required.

NOTE: The first commit is just the revert. I recommend reviewing just the second
commit to see how we deviate from the original prior-to-yesterday code.

### Verification and Testing

pytests pass and ran make_results to be sure "wages" col exists in w2.


rmudambi (approved):

rmudambi:
Thanks for handling this. Sorry for the bad instructions!

PR #315, opened 2023-08-23
Mic 4245/add ssa state pp

albrja:
### Adds SSA results when running make_state_results in post-processing
- *Category*: Post-processing
- *JIRA issue*: [MIC-4245](https://jira.ihme.washington.edu/browse/MIC-4245)
- *Research reference*: 

### Changes and notes
-Now writes all SSA data out when running make_state_results

### Verification and Testing
Successfully ran make_state_results with SSA in final outputs.

### DIFF
--- a/notebooks/2022_04_05_PRL_explore_acs_data.ipynb
+++ b/notebooks/2022_04_05_PRL_explore_acs_data.ipynb
@@ -148,7 +148,7 @@
     "# 31 .Parent-in-law\n",
     "# 32 .Son-in-law or daughter-in-law\n",
     "# 33 .Other relative\n",
-    "# 34 .Roommate or housemate or housemate\n",
+    "# 34 .Roommate or housemate\n",
     "# 35 .Foster child\n",
     "# 36 .Other nonrelative\n",
     "# 37 .Institutionalized group quarters population\n",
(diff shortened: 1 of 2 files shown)


src/vivarium_census_prl_synth_pop/tools/make_results.py
```diff
@@ -544,6 +541,9 @@ def subset_results_by_state(processed_results_dir: str, state: str) -> None:
             output_file_path = state_observer_dir / usa_obs_file.name
             if output_file_path.exists():
                 continue
-            state_data = read_datafile(usa_obs_file, reset_index=False, state=state)
+            if observer == metadata.DatasetNames.SSA:
```
albrja, line 544: @rmudambi I went back and forth on which is better - this implementation or setting state=None or the state provided in the cli command and then having one call to read_datafile. I think the method I implemented is easier to understand but not as clean.

notebooks/2022_04_05_PRL_explore_acs_data.ipynb
```diff
@@ -148,7 +148,7 @@
     "# 31 .Parent-in-law\n",
     "# 32 .Son-in-law or daughter-in-law\n",
     "# 33 .Other relative\n",
-    "# 34 .Roommate or housemate\n",
+    "# 34 .Roommate or housemate or housemate\n",
```
rmudambi, line 151: This seems wrong? Also unexpected.

src/vivarium_census_prl_synth_pop/tools/make_results.py
rmudambi: I think I prefer setting `state = None`, since you remove the duplication of calls to `read_datafile`, but you need to be careful that you do this by defining a new variable, since we don't want to change the existing value of `state` since that will be used by other datasets in subsequent iterations of the loop.

rmudambi (approved):

PR #324, opened 2023-09-27
Population component refactor

albrja:
## Refactor Population Component

###\Makes Population a sub-class of Component
- *Category*: Refactor
- *JIRA issue*: [MIC-4414](https://jira.ihme.washington.edu/browse/MIC-4414)
- *Research reference*: 

### Changes and notes
-updates pins for vivarium/vph
-updates pins for pandas and numpy
-refactors population component to be a subclass of Component

### Verification and Testing
-Ran minimum simulation with population and household components. All tests pass

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -15,7 +15,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        python-version: ["3.9", "3.10"]
+        python-version: ["3.9", "3.10", "3.11"]
     defaults:
       run:
         shell: bash -le {0}
--- a/setup.py
+++ b/setup.py
@@ -15,14 +15,14 @@ if __name__ == "__main__":
         long_description = f.read()
 
     install_requirements = [
-        "vivarium>=1.1.0",
-        "vivarium_public_health==0.10.22",
+        "vivarium>=2.0.0",
+        "vivarium_public_health>=2.0.0",
         "click",
         "gbd_mapping>=3.0.6, <4.0.0",
         "jinja2",
         "loguru",
         "numpy",
-        "pandas>=1.0.0, <2.0.0",
+        "pandas",
         "scipy",
         "tables",
         "pyyaml",
--- a/setup.py
+++ b/setup.py
@@ -22,7 +22,7 @@ if __name__ == "__main__":
         "jinja2",
         "loguru",
         "numpy",
-        "pandas",
+        "pandas>=1.0.0",
         "scipy",
         "tables",
         "pyyaml",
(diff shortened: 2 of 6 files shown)


tests/test_synthetic_pii.py
```diff
@@ -167,6 +167,7 @@ def test_first_and_middle_names(mocker, given_names, fake_obs_data):
     assert (first_names["first_name"] != middle_names["middle_name"]).any()
 
 
+@pytest.mark.slow
```
albrja, line 170: The culprit.

tests/components/observers/test_household_survey_observers.py
```diff
@@ -96,6 +96,8 @@ def test_get_observation(
     expected["middle_initial"] = ["J", "M"]
     expected[["guardian_1_address_id", "guardian_2_address_id"]] = np.nan
     expected["survey_date"] = [pd.to_datetime(sim_start_date)] * 2
+    # FIXME: Having dtype with with datime64[s] and [ns] causes pd.testing.assert_frame_equal to fail
+    expected["survey_date"] = expected["survey_date"].astype("datetime64[s]")
```
patricktnast, line 100: this is a nit, but is there a reason that setting the unit from to_datetime() wouldn't work?

patricktnast (approved):

.github/workflows/build.yml
```diff
@@ -15,7 +15,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        python-version: ["3.9", "3.10"]
```
stevebachmeier, line 18: Don't we typically include 3.8 as well?

tests/components/observers/test_household_survey_observers.py
albrja: The string we are converting to datetime sim_start_date = "2021-01-01" doesn't have the seconds units and I tried a few different things and couldn't get it to work.

.github/workflows/build.yml
albrja: We just need the latest 3 versions.

rmudambi: We usually support 3.8 as well, but there may have been some build issues with 3.8? I can't remember the specifics.

setup.py
```diff
@@ -15,14 +15,14 @@
...
-        "vivarium_public_health==0.10.22",
+        "vivarium>=2.0.0",
+        "vivarium_public_health>=2.0.0",
         "click",
         "gbd_mapping>=3.0.6, <4.0.0",
         "jinja2",
         "loguru",
         "numpy",
-        "pandas>=1.0.0, <2.0.0",
+        "pandas",
```
rmudambi, line 25: Unless you know the reason for the lower bound pin no longer applies, I'd suggest we leave it.

rmudambi (approved):

PR #326, opened 2023-09-29
Moving people component refactor

albrja:
### Refactors components where people move to use vivarium Component sublcasses.
- *Category*: Refactor
- *JIRA issue*: [MIC-4414](https://jira.ihme.washington.edu/browse/MIC-4414)
- *Research reference*: 

### Changes and notes
-refactors person and household immigration/emigration to use vivarium component

### Verification and Testing
Successfully ran simulation

### DIFF
--- a/src/vivarium_census_prl_synth_pop/components/immigration.py
+++ b/src/vivarium_census_prl_synth_pop/components/immigration.py
@@ -24,10 +24,6 @@ class Immigration(Component):
     ##############
 
     @property
-    def columns_required(self) -> List[str]:
-        return []
-
-    @property
     def on_time_step_priority(self) -> int:
         return metadata.PRIORITY_MAP["immigration.on_time_step"]
 
--- a/src/vivarium_census_prl_synth_pop/components/immigration.py
+++ b/src/vivarium_census_prl_synth_pop/components/immigration.py
@@ -1,7 +1,8 @@
 from math import ceil, floor
-from typing import Callable
+from typing import List
 
 import pandas as pd
+from vivarium import Component
 from vivarium.framework.engine import Builder
 from vivarium.framework.event import Event
 from vivarium.framework.utilities import from_yearly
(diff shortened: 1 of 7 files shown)


src/vivarium_census_prl_synth_pop/components/immigration.py
```diff
@@ -13,21 +14,22 @@
...
-
     ##############
     # Properties #
     ##############
 
     @property
-    def name(self):
-        return "immigration"
+    def columns_required(self) -> List[str]:
+        return []
```
rmudambi, line 28: Does this component needs all columns in its population view? I don't see a population view at all in the old code.

src/vivarium_census_prl_synth_pop/components/person_migration.py
```diff
@@ -24,16 +26,27 @@ class PersonMigration:
...
+        return [
+            "household_id",
+            "relationship_to_reference_person",
+            "previous_timestep_address_id",
+        ]
+
+    @property
+    def initialization_requires(self) -> Dict[str, List[str]]:
+        return {
+            "requires_columns": ["previous_timestep_address_id"],
```
rmudambi, line 44: In the previous version it looks like this component creates the column `previous_timestep_address_id` but here you're saying it requires it in initialization.

rmudambi (changes requested):

src/vivarium_census_prl_synth_pop/components/person_migration.py
albrja: Yes this was a mistake.

src/vivarium_census_prl_synth_pop/components/immigration.py
albrja: I was unsure of this. There is no population view but doesn't the component still need to use columns?

rmudambi (approved):

PR #331, opened 2023-10-26
Mic 4667/metadata file

albrja:
### Adds feature to write out metadata file in post-processing to port over values to pseudopeople
- *Category*: Post-processing
- *JIRA issue*: [MIC-4667](https://jira.ihme.washington.edu/browse/MIC-4667)
- *Research reference*: 

### Changes and notes
-adds metadata file to post-processing

### Verification and Testing
Ran post-processing and metadata file is written

### DIFF
--- a/src/vivarium_census_prl_synth_pop/tools/cli.py
+++ b/src/vivarium_census_prl_synth_pop/tools/cli.py
@@ -66,12 +66,9 @@ def make_artifacts(
 
 @click.command()
 @click.argument("output_dir", type=click.Path(exists=True))
-@click.option(
+@click.argument(
     "-V",
     "--label-version",
-    type=str,
-    default=None,
-    show_default=True,
     help="Provide a version number for final results. "
     "Version should follow format of '#.#.#'.",
 )
--- a/src/vivarium_census_prl_synth_pop/tools/cli.py
+++ b/src/vivarium_census_prl_synth_pop/tools/cli.py
@@ -69,8 +69,7 @@ def make_artifacts(
 @click.argument(
     "-V",
     "--label-version",
-    help="Provide a version number for final results. "
-    "Version should follow format of '#.#.#'.",
+    # Version for final results and metadata file. This should be of format '#.#.#'
 )
 @click.option("-v", "verbose", count=True, help="Configure logging verbosity.")
 @click.option(
--- a/src/vivarium_census_prl_synth_pop/tools/cli.py
+++ b/src/vivarium_census_prl_synth_pop/tools/cli.py
@@ -174,7 +174,24 @@ def make_results(
     peak_memory: int,
     max_runtime: str,
 ) -> None:
-    """Create final results datasets from the raw results output by observers"""
+    """Create final results datasets from the raw results output by observers
+
+    :param output_dir: Directory where the raw results are stored
+    :param label_version: Version for final results and metadata file. This should be of format '#.#.#'
+    :param verbose: Configure logging verbosity.
+    :param with_debugger: Drop into python debugger if an error occurs.
+    :param mark_best: Marks this version of results with a 'best' symlink.
+    :param test_run: Skips updating the 'latest' symlink with this version of results.
+    :param extension: File type to write results out as. Supported file types are hdf and parquet.
+    :param public_sample: Generates results for the small-scale public sample data.
+    :param seed: Provide seed in order to run only on the files corresponding to that seed. If no seed it provided, will run on all files.
+    :param artifact_path: Path to artifact used in simulation.
+    :param jobmon: Launches a jobmon workflow to generate results.
+    :param queue: NOTE: only required if --jobmon. The cluster queue to assign jobmon tasks to. long.q allows for much longer runtimes although there may be reasons to send jobs to that queue even if they don't have runtime constraints (eg node availability).
+    :param peak_memory: NOTE: only required if --jobmon. The estimated maximum memory usage in GB of an individual task. The tasks will be run with this as a limit.
+    :param max_runtime: NOTE: only required if --jobmon. The estimated maximum runtime ('hh:mm:ss') of the jobmon tasks. Once this time limit is hit, the cluster will terminate jobs regardless of queue. The maximum supported runtime is 3 days. Keep in mind that runtimes by node vary wildly.
+
+    """
     configure_logging_to_terminal(verbose)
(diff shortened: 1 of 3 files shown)


src/vivarium_census_prl_synth_pop/tools/cli.py
```diff
@@ -66,14 +66,9 @@ def make_artifacts(
...
-    "-V",
-    "--label-version",
-    type=str,
-    default=None,
-    show_default=True,
-    help="Provide a version number for final results. "
-    "Version should follow format of '#.#.#'.",
+@click.argument(
+    "label_version",
+    # Version for final results and metadata file. This should be of format '#.#.#'
```
mattkappel, line 71: This isn't enforced. Should it?

stevebachmeier: Yes it is https://github.com/ihmeuw/vivarium_census_prl_synth_pop/blob/main/src/vivarium_census_prl_synth_pop/tools/cli.py#L184

src/vivarium_census_prl_synth_pop/tools/cli.py
```diff
@@ -66,14 +66,9 @@ def make_artifacts(
...
 @click.command()
 @click.argument("output_dir", type=click.Path(exists=True))
-@click.option(
-    "-V",
-    "--label-version",
-    type=str,
-    default=None,
-    show_default=True,
-    help="Provide a version number for final results. "
-    "Version should follow format of '#.#.#'.",
```
rmudambi, line 76: Why did you remove this help text?

src/vivarium_census_prl_synth_pop/utilities.py
```diff
@@ -532,3 +532,10 @@ def copy_from_household_member(
             )
 
     return pop
+
+
+def write_metadata_file(final_output_dir: Path, label_version: str) -> None:
```
rmudambi, line 537: Why is this a parquet file? Can't it just be a yaml or json file?

rmudambi (changes requested):

src/vivarium_census_prl_synth_pop/utilities.py
```diff
@@ -532,3 +532,10 @@ def copy_from_household_member(
...
+    data_version = pd.Series(label_version)
+    metadata = pd.DataFrame({"data_version": data_version})
+    outpath = final_output_dir / "metadata.parquet"
+    metadata.to_parquet(outpath)
```
stevebachmeier, line 541: Do we want the metadata file to be parquet or somethign else like json?

src/vivarium_census_prl_synth_pop/utilities.py
```diff
@@ -532,3 +532,10 @@ def copy_from_household_member(
             )
 
     return pop
+
+
+def write_metadata_file(final_output_dir: Path, label_version: str) -> None:
+    data_version = pd.Series(label_version)
```
stevebachmeier, line 538: NOt sure what the ultimate goal of this metadata file is, but might be worth adding all the things (like all args used as input)

src/vivarium_census_prl_synth_pop/tools/cli.py
stevebachmeier: I don't think arguments can have help text. But it could certainly go in the function's docstring

albrja: Arguments can't. I can add a docstring and add it there.

src/vivarium_census_prl_synth_pop/utilities.py
albrja: I was trying to align with our other outputs from post-processing that are either hdf or parquet. We don't know all of what will be going into this file and what the format should be so I was just trying to stay consistent.

src/vivarium_census_prl_synth_pop/utilities.py
albrja: Same response as above - It can be whatever we want it to be. I was trying to align with our other outputs from post-processing that are either hdf or parquet. We don't know all of what will be going into this file and what the format should be so I was just trying to stay consistent.

src/vivarium_census_prl_synth_pop/tools/cli.py
```diff
@@ -66,14 +66,9 @@ def make_artifacts(
...
+@click.argument(
```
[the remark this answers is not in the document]
albrja: I originally had the same thought. My opinion is we have to (at any point) determine a version number to put in the metadata file. My reason for making it a required arg is since we need to it to the metadata file, we should be passing something thats in a consistent format. In pseudopeople we would then have to compare data versions for 2.0.0 or for SOME_RANDOM_DATE..

src/vivarium_census_prl_synth_pop/utilities.py
albrja: We haven't fleshed out everything that will need to be in this file. THe main purpose at some point is going to be to calculate different proportions of the datasets to use for noising in pseudopeople and not have to calculate these values in pseudopeople and do it in post-processing. I'm not sure how helpful run params would be but I'll think about it because adding more stuff now wouldn't hurt.

rmudambi: This is just to create the metadata file and validate the data version using it for now. We will almost certainly add more things to it.

NathanielBlairStahn: I don't remember everything we've talked about putting in the metadata file, but it seems like it would be good to think through it and come up with a list of ideas. I think a lot of the use cases came up in the context of providing useful user warnings without having to read the whole dataset. Here are the ones I remember:

- Available years for each dataset
- Available states for each dataset
- What locale the data is for (Anytown, RI, or USA)
- Version number of the data
- Fraction of eligible rows for various noise types (nicknames, copy from household member, guardian-based duplication) -- this may be less relevant if @zmbc [implements concatenating the data before adding noise](https://ihme.slack.com/archives/C02KUQ9LX32/p1698265517189139) rather than noising each shard separately, though we'd still need it for noising in parallel, e.g., using Modin
- I think there was some other issue with nicknames that we talked about potentially putting in a metadata file, but it would require storing a lot of data. I forget exactly what it was -- does anyone remember?

Can anyone think of other things that should go in the metadata file?

rmudambi (approved):

PR #333, opened 2023-10-28
Fuzzy checking: Bayesian version

zmbc:
### Description
- *Category*: test
- *Research reference*: none

### Changes and notes

**Terminology**:
- Here, I am considering our existing integration tests to be part of "V&V" (which I write "v_and_v" in file paths). They are the automated part, while our traditional V&V process is the manual part. This PR adds automated V&V/integration testing **of proportions** (that are not 0 or 1).
- "Fuzzy checking" is the term I am using for checking anything that is subject to sampling error/stochastic variation. For the time being, this is only proportions, but could be expanded to other types of quantities in the future.

Statistically, this uses Bayesian hypothesis testing. Each check compares two models/hypotheses of the underlying rate, one that represents "if there is a bug" and one that represents "if there is not a bug." We use a simple Jeffreys beta prior for "if there is a bug." For "if there is not a bug" we fit a beta distribution to a 95% confidence interval provided by the user. In both cases, this is then fed through a binomial (creating a beta-binomial) for the actual count data. We fail the tests if any check is "decisive" (Bayes factor > 100) in favor of "there is a bug."

Technically, this uses a pytest fixture that is kept active for the entire test session and allows testing these hypotheses while maintaining a log of information about them. This diagnostic information is output at the end of a test run for optional human inspection. I have gitignored the diagnostic output -- is this the right call?

### Verification and Testing

I have verified that these tests pass. I have also verified that they do _not_ pass, even with relatively small population (20k), when I intentionally introduce bugs into the simulation. The two that I tried were:

* Multiplying all domestic migration rates by 1.5
* Always rounding the immigration rate upward instead of using stochastic rounding


zmbc:
Note: I'm currently hoping to get review from RT on this math, both here and in https://github.com/ihmeuw/vivarium_research/pull/1382. Only after that do I think it makes sense for engineering folks to review this for code quality.

integration_tests/conftest.py
```diff
@@ -74,3 +80,189 @@ def pipeline_columns(sim, populations) -> List[str]:
...
+            return (
+                scipy.special.logit(actual_quantile) - scipy.special.logit(intended_quantile)
+            ) ** 2
+        else:
+            # In this case, we were so far off that the actual quantile can't even be
+            # precisely calculated.
+            # We return an arbitrarily large penalty to ensure this is never selected as the minimum.
+            return 1_000_000.0**2
+
+    def write_output(self) -> None:
```
albrja, line 254: This is also just writing an output file to look at for further investigation to go along with the normal test suites having to pass correct?

integration_tests/conftest.py
```diff
@@ -74,3 +80,189 @@ def pipeline_columns(sim, populations) -> List[str]:
...
+            bug_marginal_likelihood = bug_distribution.pmf(numerator)
+            no_bug_marginal_likelihood = no_bug_distribution.pmf(numerator)
+
+        try:
+            return bug_marginal_likelihood / no_bug_marginal_likelihood
+        except (ZeroDivisionError, FloatingPointError):
+            return 1_000_000.0
+
+    @cache
+    def _fit_beta_distribution_to_ui(self, lb: float, ub: float) -> Tuple[float, float]:
```
albrja, line 203: How (or why) is this being cached? Doesn't this method have to run if you hit the branch in fuzzy_assert_proportions?

integration_tests/conftest.py
zmbc: Yes, this is an output file a human could inspect, either just for visibility into how the fuzzy checking is working overall, or to see more details about the cause of warnings. I've proposed here that we output this file each time we run the integration tests, but do _not_ track it in source control, to avoid lots of diff/merge conflicts on what is essentially a build artifact.

integration_tests/conftest.py
zmbc: This is being cached in memory, so if it is called multiple times in the same run with the same lb and ub parameters, it doesn't re-do the calculation, just uses the value it got last time.

In cases where we check something every timestep, for instance, this would prevent us from having to keep recalculating the beta for the same target UI over and over.

I didn't look at how much this cache is actually being used but it seems very low cost.

integration_tests/conftest.py
```diff
@@ -74,3 +80,189 @@ def pipeline_columns(sim, populations) -> List[str]:
...
+    """
+
+    def __init__(self) -> None:
+        self.comparisons_made = []
+
+    def fuzzy_assert_proportion(
+        self,
+        name: str,
+        boolean_values: pd.Series,
+        target_value: Optional[float] = None,
```
zmbc, line 105: Could use a tuple parameter here to avoid the complexity of the checks/optionals.

integration_tests/conftest.py
```diff
@@ -74,3 +80,189 @@ def pipeline_columns(sim, populations) -> List[str]:
...
+        target_value_lb: Optional[float] = None,
```
zmbc, line 106: Don't abbreviate `lb`, `ub` etc.

integration_tests/conftest.py
```diff
@@ -74,3 +80,189 @@ def pipeline_columns(sim, populations) -> List[str]:
...
+    https://vivarium-research.readthedocs.io/en/latest/model_design/vivarium_features/automated_v_and_v/index.html#fuzzy-checking
+    """
+
+    def __init__(self) -> None:
+        self.comparisons_made = []
+
+    def fuzzy_assert_proportion(
+        self,
+        name: str,
+        boolean_values: pd.Series,
```
zmbc, line 104: Should we just pass in the aggregations we need? Proportion, length?

integration_tests/test_domestic_migration.py
```diff
@@ -1,23 +1,91 @@
...
+    for time_steps, (before, after) in enumerate(simulants_on_adjacent_timesteps):
         individual_movers = before["household_id"] != after["household_id"]
-        assert individual_movers.any()
+        fuzzy_checker.fuzzy_assert_proportion(
+            "Individual migration rate",
+            individual_movers,
+            # We say it could vary by up to 1% per timestep because of demographic
+            # change over the sim
+            target_value_lb=target_migration_rates["individual"]["total"]
+            * pow(0.99, time_steps),
```
zmbc, line 58: This "magic value" (0.99) should have units and be in the YAML.

integration_tests/test_domestic_migration.py
```diff
@@ -1,23 +1,91 @@
...
         check_po_box_collisions(after, before, individual_movers)
+        all_time_individual_movers.append(individual_movers)
+
+    all_time_individual_movers = pd.concat(all_time_individual_movers, ignore_index=True)
+    fuzzy_checker.fuzzy_assert_proportion(
+        "Individual migration rate",
+        all_time_individual_movers,
+        # We say it could vary by up to 10% because of demographic
+        # change over the sim
+        target_value_lb=target_migration_rates["individual"]["total"] * 0.9,
```
zmbc, line 76: For simplicity, let's aggregate this from the input drift per unit time, assuming each timestep has equal population size (roughly true). Create a function to do this.

zmbc (commented):
One future concern will be how we deal with this in CI, once integration tests are run in CI. Rajan's idea is to have a nightly build run the integration tests, and that nightly build should do the following:
* First, run the integration tests with a seed that is chosen randomly (different every day, and different than the engineers run manually)
* If that fails, also run the tests with the "standard" seed that is used by the engineers routinely; this will allow us to separate random failures due to seed from bugs that just weren't caught before merging

Comments transcribed from Rajan (errors are my own):

integration_tests/conftest.py
```diff
@@ -265,4 +342,68 @@ def fuzzy_checker() -> FuzzyChecker:
...
+# per unit time.
+
+
+def multiplicative_drift_to_bound_at_timestep(
+    value: float, drift_per_timestep: float, num_timesteps: int
+) -> float:
+    return value * pow(drift_per_timestep, num_timesteps)
+
+
+def multiplicative_drift_to_bounds_at_timestep(
```
rmudambi, line 362: Nit: shouldn't this be `drifts` not `bounds`? i.e.
```suggestion
def multiplicative_drifts_to_bound_at_timestep(
```

integration_tests/conftest.py
```diff
@@ -265,4 +342,68 @@ def fuzzy_checker() -> FuzzyChecker:
...
+    # Assumption: timesteps have equal sample sizes. This is reasonably accurate.
+    return np.mean(
+        [
+            multiplicative_drift_to_bound_at_timestep(value, drift_per_timestep, x)
+            for x in range(num_timesteps)
+        ]
+    )
+
+
+def multiplicative_drift_to_bounds_through_timestep(
```
rmudambi, line 390: Same here

integration_tests/test_domestic_migration.py
```diff
@@ -1,23 +1,111 @@
...
+    from_yearly_multiplicative_drift,
+    multiplicative_drift_to_bounds_at_timestep,
+    multiplicative_drift_to_bounds_through_timestep,
+)
 
-def test_individuals_move(simulants_on_adjacent_timesteps):
-    for before, after in simulants_on_adjacent_timesteps:
+
+@pytest.fixture(scope="module")
+def target_migration_rates(sim):
```
zmbc, line 21: This is where we load the YAML targets.

integration_tests/test_domestic_migration.py
```diff
@@ -1,23 +1,111 @@
...
+
+def test_individuals_move(
+    simulants_on_adjacent_timesteps, fuzzy_checker: FuzzyChecker, target_migration_rates
+):
+    all_time_individual_movers = []
+
+    for time_steps, (before, after) in enumerate(simulants_on_adjacent_timesteps):
         individual_movers = before["household_id"] != after["household_id"]
-        assert individual_movers.any()
+        fuzzy_checker.fuzzy_assert_proportion(
```
zmbc, line 68: Here's an example of a converted assert -- instead of .any(), we check the proportion using the target. Most of the diff in the test files is these conversions.

integration_tests/test_domestic_migration.py
```diff
@@ -40,19 +127,43 @@ def test_individuals_move_into_new_households(simulants_on_adjacent_timesteps):
...
-        assert new_household_movers.any()
+        joining_recently_created_households = movers_into_new_households & (
+            ~new_household_movers
+        )
+        fuzzy_checker.fuzzy_assert_proportion(
+            "Domestic migration joining recently created households",
+            observed_numerator=joining_recently_created_households.sum(),
+            observed_denominator=len(joining_recently_created_households),
+            # This case -- people who joined a just-created household -- should be exceedingly rare.
+            target_proportion=(
```
zmbc, line 138: Note: This is an unusual one where I think it makes sense for the target to be set on the engineering side. It can be very conservative.

integration_tests/test_domestic_migration.py
```diff
@@ -63,14 +174,74 @@ def test_individuals_move_into_new_households(simulants_on_adjacent_timesteps):
...
+        all_time_gq_movers.append(gq_movers)
+        into_gq_from_gq = before[gq_movers]["household_details.housing_type"] != "Household"
+        fuzzy_checker.fuzzy_assert_proportion(
+            "Domestic migration into GQ from GQ",
+            observed_numerator=into_gq_from_gq.sum(),
+            observed_denominator=len(into_gq_from_gq),
+            # Since this is sampled randomly from the population only based on demographics, it should
+            # be around 3%, with substantial leeway for demographics correlations
+            target_proportion=(
+                data_values.PROP_POPULATION_IN_GQ * 0.5,
```
zmbc, line 239: Note: This is an unusual one where I think setting the target UI on the engineering side makes sense. It can be very conservative. It's kind of a strange assertion (see line 73 on the deleted side).

integration_tests/test_domestic_migration.py
```diff
@@ -158,20 +182,26 @@ def test_individuals_move_into_new_households(
...
         "Domestic migration joining recently created households",
+        observed_numerator=all_time_joining_recently_created_households.sum(),
+        observed_denominator=len(all_time_joining_recently_created_households),
         # This case -- people who joined a just-created household -- should be exceedingly rare.
-        all_time_joining_recently_created_households,
-        target_value_lb=(1 / 1_000_000_000),
-        target_value_ub=0.001,
-        name_addl="All time steps",
+        target_proportion=(
+            (1 / 1_000_000_000),
```
rmudambi, line 189: These values are used twice (at least?). Store them as constants.

rmudambi (approved):

integration_tests/conftest.py
zmbc: There are two drift values input, and two bound values output. So maybe it's both -- `multiplicative_drifts_to_bounds_at_timestep`.

integration_tests/test_domestic_migration.py
zmbc: Fair enough. This one felt a little weird to me because it's not something I can imagine the RT thinking to validate ahead of time -- it's such a corner case. If I'm reading your comment correctly, that's okay -- but at the point where engineering decided to check this, they would have asked RT to specify the value?

If I'm understanding correctly, that makes sense to me; I can throw this in a YAML on the research side.

integration_tests/test_domestic_migration.py
zmbc: I think this is in the same category as the one above.

integration_tests/test_domestic_migration.py
rmudambi: Yes, that's exactly what I meant!

rmudambi (approved):

integration_tests/conftest.py
```diff
@@ -74,3 +81,335 @@ def pipeline_columns(sim, populations) -> List[str]:
...
+        :param target_proportion:
+            What the proportion of events / opportunities *should* be if there is no bug/issue
+            in the simulation, as the number of opportunities goes to infinity.
+            If this parameter is a tuple of two floats, they are interpreted as the 2.5th percentile
+            and the 97.5th percentile of the uncertainty interval about this value.
+            If this parameter is a single float, it is interpreted as an exact value (no uncertainty).
+            Setting this target distribution is a research task; there is much more guidance on
+            doing so at https://vivarium-research.readthedocs.io/en/latest/model_design/vivarium_features/automated_v_and_v/index.html#interpreting-the-hypotheses
+        :param fail_bayes_factor_cutoff:
+            The Bayes factor above which a hypothesis test is considered to favor a bug/issue so strongly
```
patricktnast, line 146: I think it's mostly a matter of taste in this application, but any chance I could convince you to express the Bayes factor on a decibel scale, a la Jaynes and Good?

zmbc: Interesting, TIL! I am not opposed, but would prefer to postpone that to a different PR. There are a number of directions we could go here, and the one that feels most useful to me would be to specify desired sensitivity or specificity, given the prior distributions.

The intention here is that these cutoffs wouldn't be changed often. In fact, I don't think they've had to be tweaked at all here or in pseudopeople.

patricktnast: yeah, i'm thinking more about the output message of a particular test than how to specify the cutoffs--regardless of how you specify the cutoff, would you report the test outcome in terms of odds like the Bayes Factor against a cutoff value? If so, reporting that in terms of decibel / log odds is nice/more interpretable for the human reading it (or at least for me 😄).

w.r.t. choosing the cutoffs, have you thought about going a step further in the default case and choose boundaries that given the prior optimize some discrimination metric like F score or log loss or something? Of course, in general I think you're right that it makes sense to let the test designer prioritize sensitivity / specificity/ some confusion matrix metric.

zmbc: Interesting ideas. I agree the output message could be improved; right now the Bayes factor gets printed in scientific notation a lot since it can get so large.

I found [this](https://www.johndcook.com/blog/2008/03/11/how-loud-is-the-evidence/); is this the same decibel scale you are thinking of? So with this cutoff, the test would fail when the evidence was more than 20 dB. Seems like a more intuitive way to think about it.

Really interesting idea about the discrimination metric. I am thinking about this in terms of decision theory, and I think in most cases the cost of a miss will be quite asymmetric. It might be interesting for _that_ to be the configurable parameter: what is the cost of not catching an issue vs raising a false alarm? But, the more fancy we get, the more we are relying on the prior distributions being somewhat accurate.

patricktnast: Yes, that's right--Jaynes calls it "evidence" but it has [several names](https://en.wikipedia.org/wiki/Hartley_(unit)). The particular points in favor, one of which the post you found covers:

- It's symmetric around zero (50%), so e.g. 10 and -10 dB represent the same odds for and against
- It is additive, so it is easy to keep a running total of dB with new pieces of evidence, which proponents identify with the procedure of incremental Bayesian updates
- Folks who are familiar with working with a decibel scale may know the simple power ratio heuristic 3 dB ~ 2x power and 10 dB = 10x power
-  It is speculated to be calibrated to human perception in the vein of the [Weber Fetchner law](https://en.wikipedia.org/wiki/Weber%E2%80%93Fechner_law), so that 1 dB is about as much "evidential strength" that a person can meaningfully distinguish (~25% increase)

> I am thinking about this in terms of decision theory, and I think in most cases the cost of a miss will be quite asymmetric. It might be interesting for that to be the configurable parameter: what is the cost of not catching an issue vs raising a false alarm?

Yeah I'm not super familiar with evaluation metrics--there seem to be many, especially for binary classifiers--but just looking at wikipedia they show a [parameterized Beta f-score](https://en.wikipedia.org/wiki/F-score) that weights the harmonic mean toward precision or recall. I wonder if it would be sensible to do something similar with sensitivity / specificity. At the end of the day you're trying to trade off type I and type II errors i guess, but I'm not sure how it all pencils out. But it would be nice to say "I want to prioritize sensitivty / specificity X times more than the other" without having to know the right cutoffs

> But, the more fancy we get, the more we are relying on the prior distributions being somewhat accurate.
Yeah, this is definitely an argument not to go too crazy with the information theory stuff

zmbc: I believe any decision metric will correspond directly to some loss function. In the case of a binary classification, the loss function is just a 2x2 matrix of how good/bad it is to have any combination of predicted and true class. My guess is that the loss function itself will be more intuitive to specify than the corresponding metric or even the desired sensitivity/specificity.

patricktnast (approved):

PR #334, opened 2023-11-07
Mic 4669/write shard metadata

albrja:
### Writes shard metadata..
- *Category*: Post-processing
- *JIRA issue*: [MIC-4669](https://jira.ihme.washington.edu/browse/MIC-4669)
- *Research reference*: 

### Changes and notes
-outputs a CSV metadata file for each shard which will be later processed to a final metadata yml file.

### Verification and Testing
Successfully ran post-processing and saw shard metadata files as desired.

### DIFF
--- a/src/vivarium_census_prl_synth_pop/constants/metadata.py
+++ b/src/vivarium_census_prl_synth_pop/constants/metadata.py
@@ -295,11 +295,3 @@ COPY_HOUSEHOLD_MEMBER_COLS = {
     "date_of_birth": "copy_date_of_birth",
     "has_ssn": "copy_ssn",
 }
-
-METADATA_COLUMNS = [
-    "age",
-    "date_of_birth",
-    "ssn",
-    "first_name",
-    # TODO: add guardian based duplication columns
-]
--- a/src/vivarium_census_prl_synth_pop/constants/metadata.py
+++ b/src/vivarium_census_prl_synth_pop/constants/metadata.py
@@ -295,3 +295,11 @@ COPY_HOUSEHOLD_MEMBER_COLS = {
     "date_of_birth": "copy_date_of_birth",
     "has_ssn": "copy_ssn",
 }
+
+METADATA_COLUMNS = [
+    "age",
+    "date_of_birth",
+    "ssn",
+    "first_name",
+    # TODO: add guardian based duplication columns
+]
--- a/src/vivarium_census_prl_synth_pop/tools/make_results.py
+++ b/src/vivarium_census_prl_synth_pop/tools/make_results.py
@@ -571,6 +571,7 @@ def write_shard_metadata(
             elif f"copy_{column}" in COPY_HOUSEHOLD_MEMBER_COLS.values():
                 # Proportion based on missingness from available household members that can
                 # have a copy value
+                print(f"{column} in output. Calculating metadata proportion for copy_{column}...")
                 metadata_df[f"{column}_proportion_to_noise"] = (
                     df[f"copy_{column}"].notna().mean()
                 )
--- a/src/vivarium_census_prl_synth_pop/tools/make_results.py
+++ b/src/vivarium_census_prl_synth_pop/tools/make_results.py
@@ -571,9 +571,9 @@ def write_shard_metadata(
             elif f"copy_{column}" in COPY_HOUSEHOLD_MEMBER_COLS.values():
                 # Proportion based on missingness from available household members that can
                 # have a copy value
-                metadata_df[f"{column}_proportion_to_noise"] = df[
-                    f"copy_{column}"
-                ].notna().sum() / len(df)
+                metadata_df[f"{column}_proportion_to_noise"] = (
+                    df[f"copy_{column}"].notna().mean()
+                )
             elif column == "first_name":
                 # Proportion of first names that have an associated nickname to be noised
                 # Everyone should have a first name so there is no missingness
@@ -591,13 +591,8 @@ def write_shard_metadata(
     if observer == metadata.DatasetNames.SSA:
         shard_metadata = _get_metadata_values(obs_data, "USA")
     else:
-        location_groups = obs_data.groupby(
-            [
-                col
-                for col in obs_data.columns
-                if "state" in col and col not in ["employer_state", "state_id"]
-            ]
-        )
+        groupby_col = "state" if "state" in obs_data.columns else "mailing_address_state"
+        location_groups = obs_data.groupby(groupby_col)
         metadata_dfs = []
         for location, location_data in location_groups:
             state_df = obs_data.loc[location_data.index]
--- a/src/vivarium_census_prl_synth_pop/tools/make_results.py
+++ b/src/vivarium_census_prl_synth_pop/tools/make_results.py
@@ -589,6 +589,7 @@ def write_shard_metadata(
             else:
                 raise ValueError(f"Column '{column}' not supported for metadata recording.")
         metadata_df.set_index("state", inplace=True)
+        breakpoint()
         return metadata_df
 
     # Special case SSA dataset since that does not have a state column
@@ -601,11 +602,12 @@ def write_shard_metadata(
         for location, location_data in location_groups:
             state_df = obs_data.loc[location_data.index]
             state_metadata = _get_metadata_values(state_df, location)
+            breakpoint()
             metadata_dfs.append(state_metadata)
         shard_metadata = pd.concat(metadata_dfs)
 
     shard_metadata["dataset"] = observer
-    # Seed will be "" for sample data
+    # Seed will be "" for sample data which turns into 0
     shard_metadata["random_seed"] = seed
 
     # Write shard metadata
--- a/src/vivarium_census_prl_synth_pop/tools/make_results.py
+++ b/src/vivarium_census_prl_synth_pop/tools/make_results.py
@@ -572,14 +572,14 @@ def write_shard_metadata(
                 noise_type_names = [noise_type.name for noise_type in column.noise_types]
                 if "copy_from_household_member" in noise_type_names:
                     # Get number of rows that could potentially copy a household member
-                    metadata_df[f"{column}.copy_from_household_member"] = (
-                        df[COPY_HOUSEHOLD_MEMBER_COLS[column]].notna().sum()
+                    metadata_df[f"{column_name}.copy_from_household_member"] = (
+                        df[COPY_HOUSEHOLD_MEMBER_COLS[column_name]].notna().sum()
                     )
                 if "use_nickname" in noise_type_names:
                     # Get number of rows eligible to be noised to a nickname
                     nicknames = load_nicknames_data()
-                    metadata_df[f"{column}.use_nickname"] = (
-                        df[column].isin(nicknames.index).sum()
+                    metadata_df[f"{column_name}.use_nickname"] = (
+                        df[column_name].isin(nicknames.index).sum()
                     )
                 # TODO: Add guardian based duplication
         return metadata_df
--- a/src/vivarium_census_prl_synth_pop/tools/make_results.py
+++ b/src/vivarium_census_prl_synth_pop/tools/make_results.py
@@ -607,4 +607,4 @@ def write_shard_metadata(
 
     # Write shard metadata
     shard_metadata_path = obs_dir / f"shard_metadata_{seed}.csv"
-    shard_metadata.to_csv(shard_metadata_path, index=False)
+    shard_metadata.to_csv(shard_metadata_path)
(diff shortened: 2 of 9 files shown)


src/vivarium_census_prl_synth_pop/tools/make_results.py
```diff
@@ -548,3 +554,58 @@ def subset_results_by_state(processed_results_dir: str, state: str) -> None:
...
+            else:
+                raise ValueError(f"Column '{column}' not supported for metadata recording.")
+        return metadata_df
+
+    # Special case SSA dataset since that does not have a state column
+    if observer == metadata.DatasetNames.SSA:
+        shard_metadata = _get_metadata_values(obs_data, "USA")
+    else:
+        groupby_col = "state" if "state" in obs_data.columns else "mailing_address_state"
+        location_groups = obs_data.groupby(groupby_col)
```
rmudambi, line 597: Won't this exclude a state if it is not present in a given shard? You will need to make sure you account for this when you compute the overall values.

src/vivarium_census_prl_synth_pop/tools/make_results.py
```diff
@@ -548,3 +554,58 @@ def subset_results_by_state(processed_results_dir: str, state: str) -> None:
...
+        metadata_dfs = []
+        for location, location_data in location_groups:
+            state_df = obs_data.loc[location_data.index]
+            state_metadata = _get_metadata_values(state_df, location)
+            metadata_dfs.append(state_metadata)
+        shard_metadata = pd.concat(metadata_dfs)
+
+    shard_metadata["dataset"] = observer
+    # Seed will be "" for sample data which turns into 0
+    shard_metadata["random_seed"] = seed
```
rmudambi, line 607: Do we need this seed column?

src/vivarium_census_prl_synth_pop/tools/make_results.py
```diff
@@ -548,3 +554,58 @@ def subset_results_by_state(processed_results_dir: str, state: str) -> None:
...
+def write_shard_metadata(
+    observer: str, obs_data: pd.DataFrame, obs_dir: Path, seed: str
+) -> None:
+    # Writes metadata for each shard of data. This will be proportions available to noise
+    # for specific columns.
+
+    def _get_metadata_values(df: pd.DataFrame, location: str):
+        metadata_df = pd.DataFrame(index=[location])
+        metadata_df["number_of_rows"] = len(df)
+        for column in METADATA_COLUMNS:
```
rmudambi, line 568: Wouldn't it be better to import `pseudopeople` and use the `pseudopeople` metadata to determine which columns have the copy from within household and nicknames noise functions? That would mean that we'd keep this in sync with the code from pseudopeople.

src/vivarium_census_prl_synth_pop/tools/make_results.py
```diff
@@ -548,3 +554,58 @@ def subset_results_by_state(processed_results_dir: str, state: str) -> None:
...
+            if column not in df.columns:
+                continue
+            elif f"copy_{column}" in COPY_HOUSEHOLD_MEMBER_COLS.values():
+                # Proportion based on missingness from available household members that can
+                # have a copy value
+                print(
```
rmudambi, line 574: You should use the logger rather than printing here

src/vivarium_census_prl_synth_pop/tools/make_results.py
```diff
@@ -548,3 +554,58 @@ def subset_results_by_state(processed_results_dir: str, state: str) -> None:
...
+                    df[f"copy_{column}"].notna().sum()
+                )
+            elif column == "first_name":
+                # Proportion of first names that have an associated nickname to be noised
+                # Everyone should have a first name so there is no missingness
+                nicknames = load_nicknames_data()
+                metadata_df[f"{column}_nicknames_number_of_rows"] = (
+                    df[column].isin(nicknames.index).sum()
+                )
+            # TODO: Add guaridan based duplication
```
rmudambi, line 587: Nit: typo "guaridan"

src/vivarium_census_prl_synth_pop/constants/metadata.py
```diff
@@ -295,3 +295,11 @@ class DatasetNames:
     "date_of_birth": "copy_date_of_birth",
     "has_ssn": "copy_ssn",
 }
+
+METADATA_COLUMNS = [
```
rmudambi, line 299: If you import from `pseudopeople` you won't need this.

src/vivarium_census_prl_synth_pop/tools/make_results.py
albrja: Correct. When I compute overall values wouldn't that just mean that the shard missing a state would be 0 for that state and not affect the overall values?

src/vivarium_census_prl_synth_pop/tools/make_results.py
albrja: As of now I do not see use but this was something that is obviously easy to add or take away and figured I would add it while we were setting this up.

src/vivarium_census_prl_synth_pop/tools/make_results.py
albrja: Yes - so move this to pseudopeople metadata then?

src/vivarium_census_prl_synth_pop/utilities.py
```diff
@@ -540,3 +540,11 @@ def write_metadata_file(final_output_dir: Path, label_version: str) -> None:
     outpath = final_output_dir / "metadata.yaml"
     with open(outpath, "w") as file:
         yaml.dump(metadata, file)
+
+
+def load_nicknames_data():
+    # Load and format nicknames dataset
+    nicknames = pd.read_csv(paths.NICKNAMES_DATA_PATH)
+    nicknames = nicknames.apply(lambda x: x.astype(str).str.title()).set_index("name")
```
stevebachmeier, line 548: This can be simplified to just `nicknames = nicknames.apply(lambda x: x.str.title()).set_index("name")`

Anything that didn't work will already be cast as Nonetype so no need for replacement.

src/vivarium_census_prl_synth_pop/tools/make_results.py
stevebachmeier: I _think_ what Rajan is referring to is the pseudopeople.metadata::COPY_HOUSEHOLD_MEMBER_COLS? 

But it doesn't really match what you're doing here either so idk.
https://github.com/ihmeuw/pseudopeople/blob/main/src/pseudopeople/constants/metadata.py#L86-L96

src/vivarium_census_prl_synth_pop/tools/make_results.py
```diff
@@ -548,3 +554,55 @@ def subset_results_by_state(processed_results_dir: str, state: str) -> None:
...
+    # for specific columns.
+
+    def _get_metadata_values(df: pd.DataFrame, location: str):
+        metadata_df = pd.DataFrame(index=[location])
+        metadata_df["number_of_rows"] = len(df)
+        for column in METADATA_COLUMNS:
+            if column not in df.columns:
+                continue
+            elif f"copy_{column}" in COPY_HOUSEHOLD_MEMBER_COLS.values():
+                # Proportion based on missingness from available household members that can
```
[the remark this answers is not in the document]
albrja: Yes need to update comments.

src/vivarium_census_prl_synth_pop/utilities.py
albrja: But we want them to be NaNs instead of Nonetype. I'm mainly just trying to be consistent with handling missing values as NaNs to avoid any confusion but I don't think it really matters.

src/vivarium_census_prl_synth_pop/tools/make_results.py
```diff
@@ -548,3 +552,69 @@ def subset_results_by_state(processed_results_dir: str, state: str) -> None:
...
+                # TODO: Add guardian based duplication
+                continue
+        return metadata_df
+
+    # Get year column to group by
+    date_columns = ["year", "tax_year", "event_date", "survey_date"]
+    year_col = [col for col in obs_data.columns if col in date_columns]
+    state_col = "state" if "state" in obs_data.columns else "mailing_address_state"
+    # Special case SSA dataset since that does not have a state column
+    if observer == metadata.DatasetNames.SSA:
```
albrja, line 599: There has to be a cleaner way to do some of the logic since its very similar in both cases.

src/vivarium_census_prl_synth_pop/tools/make_results.py
```diff
@@ -548,3 +552,67 @@ def subset_results_by_state(processed_results_dir: str, state: str) -> None:
...
+            for col in COLUMNS
+            for noise_type in col.noise_types
+            if "copy_from_household_member" == noise_type.name
+        ]
+        nicknames_columns = [
+            col.name
+            for col in COLUMNS
+            for noise_type in col.noise_types
+            if "use_nickname" == noise_type.name
+        ]
```
rmudambi, line 578: You can write a function to do this:

```python
def get_noise_type_columns(noise_type: str) -> List[str]:
    return [
        col.name
        for col in COLUMNS
        if noise_type in col.noise_types
    ]
```

src/vivarium_census_prl_synth_pop/tools/make_results.py
```diff
@@ -548,3 +552,67 @@ def subset_results_by_state(processed_results_dir: str, state: str) -> None:
...
+                metadata_df[f"{column}.copy_from_household_member"] = (
+                    df[COPY_HOUSEHOLD_MEMBER_COLS[column]].notna().sum()
+                )
+            elif column in nicknames_columns:
+                # Get number of rows eligible to be noised to a nickname
+                nicknames = load_nicknames_data()
+                metadata_df[f"{column}.use_nickname"] = df[column].isin(nicknames.index).sum()
+            else:
+                # TODO: Add guardian based duplication
+                continue
```
rmudambi, line 591: Instead of doing the above, you can actually be even cleaner by doing this:
```python
for column_name in df.columns:
    column = COLUMNS.get_column(column_name)
    if "copy_from_household_member" in column.noise_types:
        ...
    if "use_nickname" in column.noise_types:
        ...
```

src/vivarium_census_prl_synth_pop/tools/make_results.py
```diff
@@ -548,3 +552,67 @@ def subset_results_by_state(processed_results_dir: str, state: str) -> None:
...
+    if observer == metadata.DatasetNames.SSA:
+        metadata_dfs = []
+        for year, year_data in obs_data.groupby(year_col):
+            year_df = obs_data.loc[year_data.index]
+            year_metadata = _get_metadata_values(year_df, "USA", year)
+            metadata_dfs.append(year_metadata)
+    else:
+        groupby_cols = year_col + [state_col]
+        metadata_dfs = []
+        for (year, location), location_data in obs_data.groupby(groupby_cols):
```
rmudambi, line 608: Nit: this shouldn't be called `location_data`, since it's grouped by both location and year.

src/vivarium_census_prl_synth_pop/tools/make_results.py
```diff
@@ -548,3 +552,67 @@ def subset_results_by_state(processed_results_dir: str, state: str) -> None:
...
+            state_df = obs_data.loc[location_data.index]
```
rmudambi, line 609: How is `state_df` different from `location_data`?

src/vivarium_census_prl_synth_pop/tools/make_results.py
```diff
@@ -548,3 +552,67 @@ def subset_results_by_state(processed_results_dir: str, state: str) -> None:
...
+            state_metadata = _get_metadata_values(state_df, location, year)
+            metadata_dfs.append(state_metadata)
```
rmudambi, line 611: This could maybe be simplified:

```python
if observer == metadata.DatasetNames.SSA:
    obs_data[state_col] = "USA"

metadata_dfs = []
for (year, location), group_data in obs_data.groupby(groupby_cols):
    group_data = obs_data.loc[group_data.index]    # I think this is not actually doing anything? See question above.
    group_metadata = _get_metadata_values(group_data , location, year)
    metadata_dfs.append(group_metadata)
```

src/vivarium_census_prl_synth_pop/tools/make_results.py
rmudambi: Yes, the counts will be 0

rmudambi (commented):
Seems like this should all work, but I have some suggestions that will reduce the amount of code required and should make it easier to read.

src/vivarium_census_prl_synth_pop/tools/make_results.py
albrja: The logic isn't that simple because column.noise_types is a tuple of noise types and each noise type has different attributes that's why I'm getting the different list of columns upfront.

src/vivarium_census_prl_synth_pop/tools/make_results.py
albrja: They are the same thing.

src/vivarium_census_prl_synth_pop/utilities.py
stevebachmeier: I meant NaN.

rmudambi (approved):

PR #335, opened 2023-11-09
Mic 4670/process shard metadata

albrja:
### Processes shard metadata files in to one metadata proportions file to pass to pseudopeople for noise level validation.
- *Category*: Post-processing
- *JIRA issue*: [MIC-4670](https://jira.ihme.washington.edu/browse/MIC-4670)
- *Research reference*: 

### Changes and notes
-adds feature to the end of post-processing which aggregate and calculates metadata proportions for all dataset, location, and year combinations.

### Verification and Testing
Successfully ran make results and metadata file was written with correct outputs.
metadata

### DIFF
--- a/src/vivarium_census_prl_synth_pop/utilities.py
+++ b/src/vivarium_census_prl_synth_pop/utilities.py
@@ -550,7 +550,7 @@ def load_nicknames_data():
     return nicknames
 
 
-def extract_metadata_proportions(final_output_dir: Path) -> None:
+def iextract_metadata_proportions(final_output_dir: Path) -> None:
     # Get directories for each dataset which will contain all shard metadata files
     dataset_metadata_dfs = []
     directories = [
@@ -558,7 +558,7 @@ def extract_metadata_proportions(final_output_dir: Path) -> None:
     ]
     for directory in directories:
         dataset_directory = final_output_dir / directory
-        shard_metadata_files = sorted(list(chain(*[dataset_directory.glob(f"*.csv")])))
+        shard_metadata_files = sorted(list(chain(*[dataset_directory.glob("*.csv")])))
         metadata_dfs = []
         for shard_metadata_file in shard_metadata_files:
             metadata_df = pd.read_csv(shard_metadata_file)
--- a/src/vivarium_census_prl_synth_pop/utilities.py
+++ b/src/vivarium_census_prl_synth_pop/utilities.py
@@ -8,6 +8,7 @@ import numpy as np
 import pandas as pd
 import yaml
 from loguru import logger
+from pseudopeople.schema_entities import DATASETS
 from scipy import stats
 from vivarium.framework.engine import Builder
 from vivarium.framework.lookup import LookupTable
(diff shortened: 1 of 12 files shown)


albrja:
One thing I think needs to be in here that is not is aggregate for all locations. This seems a bit tricky because with the sample data we would already have just a US location. Not sure what your thoughts on this are @rmudambi

rmudambi:
> One thing I think needs to be in here that is not is aggregate for all locations. This seems a bit tricky because with the sample data we would already have just a US location. Should the full data (at least for each year) be calculated in here as well? Not sure what your thoughts on this are @rmudambi

Since it is possible for a user to specify "all years" we do need to have that aggregated value as well.

src/vivarium_census_prl_synth_pop/utilities.py
```diff
@@ -548,3 +548,52 @@ def load_nicknames_data():
     nicknames = nicknames.apply(lambda x: x.astype(str).str.title()).set_index("name")
     nicknames = nicknames.replace("Nan", np.nan)
     return nicknames
+
+
+def extract_metadata_proportions(final_output_dir: Path) -> None:
```
rmudambi, line 553: Nit: something like `record_metadata_proportions` would be a better method name.

src/vivarium_census_prl_synth_pop/utilities.py
```diff
@@ -548,3 +548,52 @@ def load_nicknames_data():
...
+    # Get directories for each dataset which will contain all shard metadata files
+    dataset_metadata_dfs = []
+    directories = [
```
rmudambi, line 556: Can we explicitly iterate over dataset names here? That seems clearer and would continue to work if we added any more directories in addition to "logs".

src/vivarium_census_prl_synth_pop/utilities.py
```diff
@@ -548,3 +548,52 @@ def load_nicknames_data():
...
+        x.name for x in final_output_dir.iterdir() if x.is_dir() and x.name != "logs"
+    ]
+    for directory in directories:
+        dataset_directory = final_output_dir / directory
+        shard_metadata_files = sorted(list(chain(*[dataset_directory.glob("*.csv")])))
+        metadata_dfs = []
+        for shard_metadata_file in shard_metadata_files:
+            metadata_df = pd.read_csv(shard_metadata_file)
+            metadata_dfs.append(metadata_df)
```
rmudambi, line 565: Nit:
```python
metadata_dfs = [
    pd.read_csv(shard_metadata_file) for shard_metadata_file in shard_metadata_files
]
```

src/vivarium_census_prl_synth_pop/utilities.py
```diff
@@ -548,3 +548,52 @@ def load_nicknames_data():
...
+        dataset_metadata = pd.concat(metadata_dfs)
+        # Aggregate metadata
+        groupby_cols = ["dataset", "state", "year"]
+        aggregate_cols = [col for col in dataset_metadata.columns if col not in groupby_cols]
+        aggregated_metadata = (
```
rmudambi, line 570: I think to answer your question about recording aggregated metadata, you will need to do a groupby three times, but the operation is the same as what you have below. Group by `["dataset", "state", "year"]` as you have here, but also do the same when grouping by `["dataset", "state"]` and `["dataset", "year"]`. In the aggregated case, set the excluded column equal to "all" or something similar. 

If you are concerned about the inherent duplication we'd get in the sample data since all rows will have the same state, you could the groupby to when we grouped by all the groupby columns, and not do "all" if they are identical, but I think it's actually correct to record the metadata for "US" and "all" separately even though the values will be identical, since the user could get there by having filtered to the "state" or by asking for "all states".

rmudambi (changes requested):

src/vivarium_census_prl_synth_pop/utilities.py
```diff
@@ -548,3 +548,52 @@ def load_nicknames_data():
...
+            dataset_metadata.groupby(by=groupby_cols)[aggregate_cols].sum().reset_index()
+        )
+        # Calculate proportions for each dataset's location, year combination.
+        # Reshape metadata to have dataset, year, state, column, noise_type
+        # and proportion columns
+        melted_metadata = pd.melt(
+            aggregated_metadata,
+            id_vars=["dataset", "state", "year", "number_of_rows"],
+        )
+        # "Variable" column is created from melting and is all the different column and noise types
```
stevebachmeier, line 580: Thanks for the helpful comments

stevebachmeier (commented):
This seems complicated enough to perhaps warrant some tests?

src/vivarium_census_prl_synth_pop/utilities.py
albrja: I think that should work.

src/vivarium_census_prl_synth_pop/utilities.py
```diff
@@ -550,26 +551,24 @@ def load_nicknames_data():
...
+            pd.read_csv(shard_metadata_file) for shard_metadata_file in shard_metadata_files
+        ]
         dataset_metadata = pd.concat(metadata_dfs)
         # Aggregate metadata
         groupby_cols = ["dataset", "state", "year"]
         aggregate_cols = [col for col in dataset_metadata.columns if col not in groupby_cols]
         aggregated_metadata = (
             dataset_metadata.groupby(by=groupby_cols)[aggregate_cols].sum().reset_index()
         )
+        breakpoint()
```
rmudambi, line 571: Breakpoint?

rmudambi:
Sorry, I accidentally approved this thinking it was the other PR

PR #340, opened 2024-01-10
Bugfix/guardian dup counts

albrja:
### Fixes calculation for guardian duplication metadata
- *Category*: Bugfix
- *JIRA issue*: [MIC-4733](https://jira.ihme.washington.edu/browse/MIC-4733)
- *Research reference*:

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -46,6 +46,13 @@ jobs:
           else
             echo "upstream_vivarium_inputs_exist=false" >> $GITHUB_ENV
           fi
+      - name: check for upstream pseudopeople
+        run: |
+          if git ls-remote --exit-code --heads https://github.com/ihmeuw/pseudopeople.git ${GITHUB_HEAD_REF} == "0"; then
+            echo "upstream_pseudopeople_exist=true" >> $GITHUB_ENV
+          else
+            echo "upstream_pseudopeople_exist=false" >> $GITHUB_ENV
+          fi
       - name: print environment values
         run: |
           cat $GITHUB_ENV
@@ -74,6 +81,14 @@ jobs:
           echo "Cloning upstream vivarium_inputs branch: ${GITHUB_HEAD_REF}"
           git clone --branch=${GITHUB_HEAD_REF} https://github.com/ihmeuw/vivarium_inputs.git
           pushd vivarium_inputs
+          pip install .
+          popd
+      - name: Retrieve upstream pseudopeople
+        if: env.upstream_pseudopeople_exist == 'true'
+        run: |
+          echo "Cloning upstream pseudopeople branch: ${GITHUB_HEAD_REF}"
+          git clone --branch=${GITHUB_HEAD_REF} https://github.com/ihmeuw/pseudopeople.git
+          pushd pseudopeople
           pip install .
           popd
       - name: Install dependencies
--- a/src/vivarium_census_prl_synth_pop/utilities.py
+++ b/src/vivarium_census_prl_synth_pop/utilities.py
@@ -572,38 +572,53 @@ def record_metadata_proportions(final_output_dir: Path) -> None:
             dataset_metadata.groupby(groupby_cols)[aggregate_count_cols].sum().reset_index()
         )
 
+        # Denominator columns
+        denominator_columns_mapper = {
+            "row_noise.row_probability_in_households_under_18": "group_rows.row_probability_in_households_under_18",
+            "row_noise.row_probability_in_college_group_quarters_under_24": "group_rows.row_probability_in_college_group_quarters_under_24",
+        }
+        denominator_columns = list(denominator_columns_mapper.values()) + [
+            data_keys.METADATA_COLUMNS.NUMBER_OF_ROWS
+        ]
+        numerator_columns = [
+            col
+            for col in dataset_metadata.columns
+            if col not in groupby_cols + denominator_columns
+        ]
+        for column in numerator_columns:
+            denominator_column = denominator_columns_mapper.get(
+                column, data_keys.METADATA_COLUMNS.NUMBER_OF_ROWS
+            )
+            # Calculate noise proportions - prevent divide by zero error caused by very
+            dataset_metadata[f"{column}.proportion"] = (
+                dataset_metadata[column] / dataset_metadata[denominator_column]
+            )
+            # This replaces nan caused by divide by zero with 0.0 in cases where there is really small
+            # population.
+            dataset_metadata.loc[
+                dataset_metadata[f"{column}.proportion"].isna(), f"{column}.proportion"
+            ] = 0.0
+
+        # Drop columns that are no longer needed
+        dataset_metadata = dataset_metadata.drop(
+            columns=numerator_columns + denominator_columns
+        )
+
         # Calculate proportions for each dataset's location, year combination.
         # Reshape metadata to have dataset, year, state, column, noise_type
         # and proportion columns
-        melted_metadata = pd.melt(
-            dataset_metadata,
-            id_vars=["dataset", "state", "year", "number_of_rows"],
-        )
+        melted_metadata = pd.melt(dataset_metadata, id_vars=groupby_cols)
 
         # "Variable" column is created from melting and is all the different column and noise types
         # we have in the metadata and value is that original columns value
-        # Example age.copy_from_household_member is in variable.unique() and its "value"
+        # Example age.copy_from_household_member.proportion is in variable.unique() and its "value"
         # is the number of non-null rows we recorded in the original shard metadata
         # Column is either a column name or row noise and noise type is the column
         # or row noise type.
         melted_metadata["column"] = melted_metadata["variable"].str.split(".").str[0]
         melted_metadata["noise_type"] = melted_metadata["variable"].str.split(".").str[1]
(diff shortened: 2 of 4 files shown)


.github/workflows/build.yml
```diff
@@ -46,6 +46,13 @@ jobs:
           else
             echo "upstream_vivarium_inputs_exist=false" >> $GITHUB_ENV
           fi
+      - name: check for upstream pseudopeople
```
albrja, line 49: The changes in this file will allow us to import from pseudopeople

src/vivarium_census_prl_synth_pop/utilities.py
```diff
@@ -670,13 +681,28 @@ def get_guardian_duplication_row_counts(
...
+    )
 
     return metadata_df
 
 
 def _get_metadata_counts(
     observer: str, df: pd.DataFrame, location: str, year: int
 ) -> pd.DataFrame:
+    # FIXME: import this from pseudopeople after release
+    COPY_HOUSEHOLD_MEMBER_COLS = {
```
albrja, line 695: We currently can't important this from pseudopeople because pseudopeople will be on a release branch and not main.

stevebachmeier: Are you saying that this object isn't yet released in pseudopeople and so you need to wait for v1.0?

src/vivarium_census_prl_synth_pop/utilities.py
```diff
@@ -670,13 +681,28 @@ def get_guardian_duplication_row_counts(
...
+        "age": "copy_age",
+        "date_of_birth": "copy_date_of_birth",
+        "ssn": "copy_ssn",
```
stevebachmeier, line 698: Should any of these strings be containerized? Super annoying should we ever need to change col names - not to mention reducing chances of typos.

src/vivarium_census_prl_synth_pop/utilities.py
```diff
@@ -707,3 +733,31 @@ def _get_metadata_counts(
                     df[column_name].isin(nicknames.index).sum()
                 )
     return metadata_df
+
+
+def update_guardian_duplication_row_counts(melted_df: pd.DataFrame) -> pd.DataFrame:
+    # Update guardian duplication row counts to make number of rows match anticipated
```
stevebachmeier, line 739: nit: we typically wrap docstrings with """""" and then in-line comments with #. Maybe it's just me, but it makes it easier for me to distinguish between the function's comment and specific step comments.

src/vivarium_census_prl_synth_pop/utilities.py
```diff
@@ -707,3 +733,31 @@ def _get_metadata_counts(
...
+    )
+    # Remove these rows from so we do not output them later
+    melted_df = melted_df.loc[
+        melted_df.index.difference(guardian_duplication_group_row_counts.index)
+    ]
+    # Merge two dataframes
+    melted_df = melted_df.merge(
+        guardian_duplication_group_row_counts,
+        how="left",
+        on=["dataset", "state", "year", "noise_type", "number_of_rows"],
```
stevebachmeier, line 756: Same as above - I'm not sure if any of these strings are containerized somewhere but if not they probably should be

stevebachmeier (approved):

src/vivarium_census_prl_synth_pop/utilities.py
```diff
@@ -587,6 +588,13 @@ def record_metadata_proportions(final_output_dir: Path) -> None:
         # or row noise type.
         melted_metadata["column"] = melted_metadata["variable"].str.split(".").str[0]
         melted_metadata["noise_type"] = melted_metadata["variable"].str.split(".").str[1]
+        # Swap in correct number of rows for guardian duplication to calculate proportions
+        if dataset in [
+            metadata.DatasetNames.CENSUS,
+            metadata.DatasetNames.ACS,
+            metadata.DatasetNames.CPS,
+        ]:
+            melted_metadata = update_guardian_duplication_row_counts(melted_metadata)
```
stevebachmeier, line 597: I don't understand the motivation for this change. What exactly is this fixing?

stevebachmeier (changes requested):
No explicit changes aside from nits - I just don't understand what this is fixing

src/vivarium_census_prl_synth_pop/utilities.py
albrja: Yea. I could import it using editable install but the github builds will still break.

src/vivarium_census_prl_synth_pop/utilities.py
albrja: These are stored in the right places in pseudopeople. We have circular dependency issues in pseudopeople which prevents an ideal solution but the list are stored in the proper metadata locations.

src/vivarium_census_prl_synth_pop/utilities.py
albrja: Basically guardian duplication configuration values are conditional probabilities meaning its the probability you are in a one of the two groups and live at a different address from one of your guardians given you are in one of the groups. The error Im fixing is we get "number of rows" which is the length of our dataframe but for guardian duplication we need to get the number of rows for each group. So in one of the helper functions we capture that value and then in this function we swap that out with number of rows.

.github/workflows/build.yml
albrja: Yes

stevebachmeier (approved):
I still think there's a simpler way to update those denominators but can't quite sort it out yet. Nice work!

patricktnast (approved):

src/vivarium_census_prl_synth_pop/utilities.py
```diff
@@ -572,36 +572,57 @@ def record_metadata_proportions(final_output_dir: Path) -> None:
...
+            "row_noise.row_probability_in_households_under_18": "group_rows.row_probability_in_households_under_18",
+            "row_noise.row_probability_in_college_group_quarters_under_24": "group_rows.row_probability_in_college_group_quarters_under_24",
+        }
+
+        denominator_columns = (
+            list(denominator_columns_mapper.values())
+            + [data_keys.METADATA_COLUMNS.NUMBER_OF_ROWS]
+            if set(denominator_columns_mapper.values()).issubset(dataset_metadata.columns)
+            else [data_keys.METADATA_COLUMNS.NUMBER_OF_ROWS]
+        )
```
rmudambi, line 586: Are you doing this for any reason other than that you are getting an error when you try to drop columns that don't exist? If so, I'd suggest you do this ...
```suggestion
        denominator_columns = list(denominator_columns_mapper.values()) + [data_keys.METADATA_COLUMNS.NUMBER_OF_ROWS]
```

src/vivarium_census_prl_synth_pop/utilities.py
```diff
@@ -572,36 +572,57 @@ def record_metadata_proportions(final_output_dir: Path) -> None:
...
+            )
+            # This replaces nan caused by divide by zero with 0.0 in cases where there is really small
+            # population.
+            dataset_metadata.loc[
+                dataset_metadata[f"{column}.proportion"].isna(), f"{column}.proportion"
+            ] = 0.0
+
+        # Drop columns that are no longer needed
+        dataset_metadata = dataset_metadata.drop(
+            columns=numerator_columns + denominator_columns
```
rmudambi, line 608: ... and add `errors=ignore` to this `drop` call.

src/vivarium_census_prl_synth_pop/utilities.py
```diff
@@ -572,36 +572,57 @@ def record_metadata_proportions(final_output_dir: Path) -> None:
...
+            )
+            # Calculate noise proportions - prevent divide by zero error caused by very
+            dataset_metadata[f"{column}.proportion"] = (
+                dataset_metadata[column] / dataset_metadata[denominator_column]
+            )
+            # This replaces nan caused by divide by zero with 0.0 in cases where there is really small
+            # population.
+            dataset_metadata.loc[
+                dataset_metadata[f"{column}.proportion"].isna(), f"{column}.proportion"
+            ] = 0.0
```
rmudambi, line 604: This can be simplified to 

```suggestion
            dataset_metadata[f"{column}.proportion"] = (
                dataset_metadata[column] / dataset_metadata[denominator_column]
            ).fillna(0.0)
```

rmudambi (approved):

src/vivarium_census_prl_synth_pop/utilities.py
```diff
@@ -572,36 +572,57 @@ def record_metadata_proportions(final_output_dir: Path) -> None:
...
+        numerator_columns = [
+            col
+            for col in dataset_metadata.columns
+            if col not in groupby_cols + denominator_columns
+        ]
+        for column in numerator_columns:
+            denominator_column = denominator_columns_mapper.get(
+                column, data_keys.METADATA_COLUMNS.NUMBER_OF_ROWS
+            )
+            # Calculate noise proportions - prevent divide by zero error caused by very
```
stevebachmeier, line 596: finish the comment

src/vivarium_census_prl_synth_pop/utilities.py
stevebachmeier: lol I totally forgot about .fillna!

stevebachmeier (approved):

PR #346, opened 2024-02-29
Fit beta distributions more robustly

zmbc:
### Description
- *Category*: bugfix
- *JIRA issue*: [MIC-4828](https://jira.ihme.washington.edu/browse/MIC-4828)
- *Research reference*: none

### Changes and notes

Completely changes our approach to fitting beta distributions to make it less finicky. The thinking behind the new approach can be found in the docstring.

### Verification and Testing

Integration tests pass on the same CPU architecture and environment where they were failing before (like Jenkins), without any warnings related to convergence.

I have verified that this is reasonably robust to the input bounds with:

```python
def test_fit(lower_bound, upper_bound):
    import importlib
    import conftest
    importlib.reload(conftest)
    a, b = conftest.FuzzyChecker()._fit_beta_distribution_to_uncertainty_interval(lower_bound, upper_bound)
    dist = scipy.stats.beta(
        a=a,
        b=b,
    )
    with np.errstate(under="ignore"):
        lb_cdf = dist.cdf(lower_bound)
        ub_cdf = dist.cdf(upper_bound)
    assert np.isclose(lb_cdf, 0.025, atol=0.01), f"{lb_cdf} not close to {0.025}, {lower_bound} {upper_bound}"
    assert np.isclose(ub_cdf, 0.975, atol=0.01), f"{ub_cdf} not close to {0.975}, {lower_bound} {upper_bound}"

for start_val in [1e-14, 1e-10, 0.000001, 0.01, 0.1, 0.2, 0.3, 0.4, 0.5, 0.6, 0.8, 0.9, 0.99999, 1.0 - 1e-10]:
    for width in [1e-14, 1e-12, 1e-10, 0.0000001, 0.00001, 0.001, 0.01, 0.03, 0.05, 0.1, 0.2, 0.4, 0.6, 0.9]:
        if start_val + width >= 1:
            continue

        test_fit(start_val, start_val + width)
```

which runs with no errors or warnings. This code snippet can be converted into an actual test once FuzzyChecker is in a standalone package.

pseudopeople does not use this functionality, so is unaffected.

### DIFF
--- a/integration_tests/conftest.py
+++ b/integration_tests/conftest.py
@@ -332,7 +332,7 @@ class FuzzyChecker:
                 lb_cdf = dist.cdf(lower_bound)
                 ub_cdf = dist.cdf(upper_bound)
 
-                error = self._ui_squared_error(concentration, mean, lower_bound, upper_bound)
+                error = self._ui_squared_error(dist, lower_bound, upper_bound)
                 if best_error is None or error < best_error:
                     best_error = error
                     best_concentration = concentration
@@ -397,12 +397,8 @@ class FuzzyChecker:
         return tuple(result)
 
     def _ui_squared_error(
-        self, concentration: float, mean: float, lower_bound: float, upper_bound: float
+        self, dist: scipy.stats.rv_continuous, lower_bound: float, upper_bound: float
     ) -> float:
-        a = mean * concentration
-        b = (1 - mean) * concentration
-        dist = scipy.stats.beta(a=a, b=b)
-
         squared_error_lower = self._quantile_squared_error(dist, lower_bound, 0.025)
         squared_error_upper = self._quantile_squared_error(dist, upper_bound, 0.975)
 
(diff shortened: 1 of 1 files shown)


integration_tests/conftest.py
```diff
@@ -272,51 +272,145 @@ def _calculate_bayes_factor(
...
+        a = mean * concentration
+        b = (1 - mean) * concentration
+        dist = scipy.stats.beta(a=a, b=b)
+
+        squared_error_lower = self._quantile_squared_error(dist, lower_bound, 0.025)
+        squared_error_upper = self._quantile_squared_error(dist, upper_bound, 0.975)
+
+        try:
+            return squared_error_lower + squared_error_upper
+        except FloatingPointError:
```
albrja, line 411: Does this essentially raise this error before we would expect to hit it how we were with the current failures due to hardware?

zmbc: This catches errors due to error values that are too large to be stored in a float. Overall, this PR changes what we were doing completely!

rmudambi (approved):

PR #357, opened 2024-07-26
Feature/sbachmei/mic 5151 update post processing

stevebachmeier:
### Description

- *Category*:  other
- *JIRA issue*: https://jira.ihme.washington.edu/browse/MIC-5151
- *Research reference*:  na

### Changes and notes
 
MIC-5056 implemented the new results processing approach into the simulation.
In most models, this is good enough and we don't have a second `make_results`
step anymore. In this case, however, we do indeed need a second step that
takes the (previously-called) "raw_results" data (now just the data in the 
"results" folder), substantially modifies it, and saves out to a "final_results"
folder.

primary changes here include:

- getting rid of file support; we are all-in on parquet now.
- adding a bit of validation around the path passed into `make_results`
- (the primary goal) reading from a single dataset w/ all seeds as opposed
  to the old way of having seed-specific raw_results files

### Verification and Testing

pytest --runslow works. 

I also compared before/after like I did w/ the raw results
and things look fine except for the caveats I raised in that PR. Namely, there seems
to be some non-reproducibility in guardian_id column (which leads to other columns
not being reproducible, ee.g. copying from guardians, guardian names, etc) as well
as simulant_id in the wic datasets being off by a small number.


src/vivarium_census_prl_synth_pop/constants/paths.py
```diff
@@ -7,18 +7,19 @@
...
 RAW_DATA_ROOT = BASE_DIR / "data" / "raw_data"
 
-RAW_RESULTS_DIR_NAME = Path("raw_results")
-FINAL_RESULTS_DIR_NAME = Path("final_results")
+RAW_RESULTS_DIR_NAME = "results"
+FINAL_RESULTS_DIR_NAME = "final_results"
 
-BEST_DIR_NAME = Path("best")
-LATEST_DIR_NAME = Path("latest")
+PROCESSED_RESULTS_DIR_NAME_BASE = "pseudopeople_input_data_usa"
```
stevebachmeier, line 17: This is the only new object I added to paths. The rest of the changes are just my ocd doing its thing

src/vivarium_census_prl_synth_pop/results_processing/formatter.py
```diff
@@ -145,6 +145,11 @@ def format_columns(data: pd.DataFrame) -> pd.DataFrame:
     ) in COLUMN_FORMATTERS.items():
         if set(required_columns).issubset(set(data.columns)):
             data[output_column] = column_formatter(data[required_columns])
+    # Drop psimulate-specific columns that are not needed
+    drop_cols = ["input_draw", "scenario"]
```
stevebachmeier, line 149: I think it's actually fine to leave these columns (new when doing psimulate now), but I drop them for consistency w/ the old results. My understanding is that we don't (and never will?) vary by scenario or draw for this model

src/vivarium_census_prl_synth_pop/tools/cli.py
```diff
@@ -174,9 +165,10 @@ def make_results(
     peak_memory: int,
     max_runtime: str,
 ) -> None:
-    """Create final results datasets from the raw results output by observers
-
-    :param output_dir: Directory where the raw results are stored
+    """Create final results datasets from the raw results found in the OUTPUT_DIR
+    and store them in a time-stamped and versioned directory structure using LABEL_VERSION.
+    \f
```
stevebachmeier, line 170: `\f` breaks the click --help docstring. So when you do --help it won't show the `:param` stuff below it since those are duplicates of what's in the actualy click.option help string. (but the param stuff is needed for sphinx documentation so it'll show up there)

src/vivarium_census_prl_synth_pop/tools/cli.py
```diff
@@ -308,13 +289,37 @@ def jobmon_make_results_runner(
...
+    """Subset the processed results found in PROCESSED_RESULTS_DIR to a specific STATE.
+    \f
+    :param processed_results_dir: Directory where the already-processed results exist.
+    :param verbose: Configure logging verbosity.
+    :param with_debugger: Drop into python debugger if an error occurs.
+    :param state: The state to subset to. This should be the two letter postal abbreviation.
+    """
+    resolved_results_dir = Path(processed_results_dir).resolve()
+    # Ensure only a single expected results directory exists
+    resolved_results_dir = [
```
stevebachmeier, line 306: it was super confusing which path had to be passed in for this to work, so now it basically doesn't matter as long as there is one and only one "pseudopeople_input_data_usa" directory nested somewhere.

src/vivarium_census_prl_synth_pop/tools/make_results.py
```diff
@@ -396,66 +393,33 @@ def perform_post_processing(
...
-                    f"No file found with the seed {seed} for observer {observer}."
-                )
-            obs_file = obs_files[0]
-            logger.info(f"Loading data for {obs_file.name}")
-            obs_data = read_datafile(obs_file)
-            obs_data["random_seed"] = seed
-            obs_data = formatter.format_columns(obs_data)
-
-        observers_results[obs_dir.name] = obs_data
+        logger.info(f"Loading data for {observer}")
```
stevebachmeier, line 403: This bit is the primary goal of this PR. We no longer have seed-specific raw results but rather a single file (per observer) with all seeds in it. So between that, dropping support for multiple file types (parquet AND hdf), and a small refactor to `read_datafile`, this all becomes simpler

src/vivarium_census_prl_synth_pop/tools/make_results.py
```diff
@@ -396,66 +393,33 @@ def perform_post_processing(
...
+            write_to_disk(obs_data.copy(), obs_dir / f"{observer}{seed_ext}.parquet")
 
 
 def load_data(raw_results_dir: Path, seed: str) -> Dict[str, pd.DataFrame]:
-    # Loads in all observer outputs and stores them in a dict with observer name as keys.
+    """Loads in all observer outputs and stores them in a dict with observer name as keys."""
     observers_results = {}
     for observer in FINAL_OBSERVERS:
-        obs_dir = raw_results_dir / observer
-        if seed == "":
```
stevebachmeier, line 407: I'm able to delete this seed logic by instead adding `seed` as an arg to `read_datafile` with a defautl of "" and handle it in there

src/vivarium_census_prl_synth_pop/tools/make_results.py
```diff
@@ -396,66 +393,33 @@ def perform_post_processing(
...
-                    f" - {obs_files}."
-                )
-            elif len(obs_files) == 0:
-                raise FileNotFoundError(
-                    f"No file found with the seed {seed} for observer {observer}."
-                )
-            obs_file = obs_files[0]
-            logger.info(f"Loading data for {obs_file.name}")
-            obs_data = read_datafile(obs_file)
-            obs_data["random_seed"] = seed
```
stevebachmeier, line 435: random seed now comes for free w/ the new results processing

src/vivarium_census_prl_synth_pop/tools/make_results.py
```diff
@@ -396,66 +393,33 @@ def perform_post_processing(
...
-        raise ValueError(
-            f"Supported extensions are {metadata.SUPPORTED_EXTENSIONS}. "
-            f"{file.suffix[1:]} was provided."
-        )
+def read_datafile(
+    file: Path, reset_index: bool = True, state: Optional[str] = None, seed: str = ""
+) -> pd.DataFrame:
+    state_column = "mailing_address_state" if "tax" in file.name else "state"
+    state_filter = (state_column, "==", state) if state else None
+    seed_filter = ("random_seed", "==", int(seed)) if seed else None
```
stevebachmeier, line 416: We already had a state_filter but now I add a seed_filter

src/vivarium_census_prl_synth_pop/tools/make_results.py
```diff
@@ -510,56 +474,41 @@ def generate_maps(
...
-    version = usa_results_dir.name.split("_")[-1]
-    if version != "usa":
-        state_dir = (
-            processed_results_dir
-            / "states"
-            / f"pseudopeople_input_data_{state_name}_{version}"
-        )
-    else:
-        state_dir = processed_results_dir / "states" / f"pseudopeople_input_data_{state_name}"
+    version = processed_results_dir.name.split("_")[-1]
```
stevebachmeier, line 482: since we handled the `processed_results_dir` up front in cli.py, this all is simpler

src/vivarium_census_prl_synth_pop/tools/cli.py
albrja: So much better

albrja (approved):

src/vivarium_census_prl_synth_pop/utilities.py
```diff
@@ -451,24 +451,28 @@ def get_state_puma_options(builder: Builder) -> pd.DataFrame:
...
-    raw_results_files = list(
-        chain(*[raw_output_dir.rglob(f"*.{ext}") for ext in metadata.SUPPORTED_EXTENSIONS])
-    )
-
-    return sorted(list(set([x.name.split(".")[0].split("_")[-1] for x in raw_results_files])))
+    """Read the seeds from the either the finished_sim_metadata.csv if a
+    parallel simulation or the metadata.yaml if a single simulation
+    """
+    single_sim_metadata = raw_output_dir.parent / "metadata.yaml"
+    parallel_sim_metadata = raw_output_dir.parent / "finished_sim_metadata.csv"
```
rmudambi, line 458: Nit: should the file names be extracted as constants to `paths.py`?

.gitignore
```diff
@@ -134,6 +134,7 @@ dmypy.json
 
 # Model specification files
 src/vivarium_census_prl_synth_pop/model_specifications/*.yaml
+src/vivarium_census_prl_synth_pop/model_specifications/branches/*.yaml
```
rmudambi, line 137: Why are these being git-ignored?

rmudambi (approved):
Looks much cleaner!

.gitignore
stevebachmeier: I assume to prevent folks from continually and accidentally adding their test yamls? I saw that the model_spec was already being ignored so followed suit w/ the scenarios file.

